PR #537, opened 2022-05-04
Add StackReferenceArgs.Builder

### LINKED ISSUE #534: StackReferenceArgs missing `T` overload
## Hello!

## Issue details
Missed adding an overload to StackReferenceArgs.

StackReference can accept a plain string if the reference is passed as the first argument, but if the Args builder is used it only takes an output. Should be an easy fix.

https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReferenceArgs.java

pawelprazak:
# Description

Fixes #534

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,9 +1,10 @@
 ### Improvements
 
-
 - Fix #534: StackReferenceArgs can now be instantiated using StackReferenceArgs.builder() and StackReferenceArgs.Builder, 
   also StackReferenceArgs.getName() was deprecated in favour of StackReferenceArgs.name()
-
+[#537](https://github.com/pulumi/pulumi-java/pull/537)
+- Fix #430: now you can pass both plain Object and Output-wrapped objects to Output.format, as in: Output.format("Resource %s has ID %s", myResource, myResource.getId())
+[#539](https://github.com/pulumi/pulumi-java/pull/539)
 - Fix #476: add examples from PulumiUp
 [#542](https://github.com/pulumi/pulumi-java/pull/542)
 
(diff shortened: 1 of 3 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReferenceArgs.java
```diff
@@ -7,26 +7,79 @@
...
+     */
+    public static final class Builder {
+        @Nullable
+        private Output<String> name;
+
+        /**
+         * @param name the name of the stack to reference.
+         * @return the {@link Builder}
+         * @see StackReferenceArgs#name()
+         */
```
dixler, line 62: Nit: would like something describing the structure of the string `<organization>/<project>/<stack>`

dixler (approved):
Aside from the nitpick, it works on my machine.

CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,5 @@
 ### Improvements
 
 ### Bug Fixes
+
+- Fix #534: add StackReferenceArgs.builder() and StackReferenceArgs.Builder, deprecate StackReferenceArgs.getName()
```
t0yv0, line 5: This sounds like an Improvement not a Bug Fix.

t0yv0 (approved):

sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReferenceArgs.java
Frassle: It's actually <stack> or <org>/<stack> or <org>/<proj>/<stack> just to complicate things :)

CHANGELOG_PENDING.md
pawelprazak: thank you, moved

PR #539, opened 2022-05-04
Relax Output.format signature

### LINKED ISSUE #430: Relax signature of Output.format
## Hello!

## Issue details

Suggestion: accept Object instead of Output so that prompt values can be passed alongside Output values to Output.format. 

So that we can do:

```
String foo = "foo";
Output<String> bar = Output.format("%s: %s", foo, Output.of(foo));
// bar carries a value `foo: foo` now
```

pawelprazak:
# Description

Fixes #430

## Checklist


sdk/java/pulumi/src/test/java/com/pulumi/core/OutputsTest.java
```diff
@@ -32,7 +33,40 @@ public void testNullableSecretifyOutput() {
     public void testFormat() {
         var foo = Output.of("foo");
         var bar = Output.of("bar");
-        var result = Output.format("%s-%s", foo, bar);
-        assertThat(waitForValue(result)).isEqualTo("foo-bar");
+        var baz = "baz";
+        var result = Output.format("%s-%s-%s-%d", foo, bar, baz, null);
```
t0yv0, line 37: Good.

sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
```diff
@@ -221,20 +224,39 @@ private static <T> Output<List<T>> all(List<Output<T>> outputs) {
...
-     * @param formattableString the format string with same syntax as expected by {@link String#format(String, Object...)}
-     * @param arguments         {@link Output}s that values of will be applied to the format String
+     * @param formattableString The format string with same syntax as expected by {@link String#format(String, Object...)},
+     *                          the behaviour on a {@code null} argument depends on the format conversion.
+     * @param arguments         {@link Output}s or regular {@link Object}s that values of will be applied to the format String
      * @return an {@link Output} with the result of the formatting
      */
-    static Output<String> format(String formattableString, @SuppressWarnings("rawtypes") Output... arguments) {
-        var data = Lists.newArrayList(arguments).stream()
+    static Output<String> format(String formattableString, @Nullable Object... arguments) {
```
t0yv0, line 232: I wonder why does the implementation need to drop to OutputData layer?

You have `<T> Output<List<T>> all(List<Output<T>> outputs)`.

Why cannot that be used here? 

`List<Object> --> List<Output<Object>> --> Output<List<Object>> --> Output<String>`

t0yv0: https://github.com/pulumi/pulumi-java/issues/552 is the bug

pawelprazak: Yes, that one was doing weird things, didn't want to mix the two issues.

t0yv0: Yes, that makes sense. I think the implementation is functionally okay it does the same thing, just weird.

sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
```diff
@@ -221,20 +224,39 @@ private static <T> Output<List<T>> all(List<Output<T>> outputs) {
...
+                .thenApply(dataList ->
+                        OutputData.builder(new ArrayList<>(dataList.size()))
+                                .accumulate(dataList, (ts, t) -> {
+                                    ts.add(t);
+                                    return ts;
+                                })
+                                .build()
+                )
+                .thenApply(
+                        objs -> objs.apply(v -> v == null ? null : String.format(formattableString, v.toArray()))
```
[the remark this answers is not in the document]
pawelprazak: with the current implementation is should never be null

sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
pawelprazak: as for the dropping, `.map(OutputData::copyInputOutputData)` this defensive copy looks like the only reason

pawelprazak:
Did a merge since git rebase is broken by the case-sensitive path problem.

t0yv0:
Before we have a solution for that, you can workaround by using a Linux VM if you need to - the case-sensitive path is an OSX/Windows problem it seems.

PR #541, opened 2022-05-04
Remove SDK dependency on Mockito

### LINKED ISSUE #419: Remove SDK dependency on Mockito
## Hello!

## Issue details

Commit https://github.com/pulumi/pulumi-jvm/commit/a41a0e9951cf7d6d25e7b203f99263877d05d933 
has made SDK depend on a test-only dependency, if this is required then a separate library needs to be created just for tests.

User production code will depend on a test library if this issue is not addressed.

pawelprazak:
# Description

Fixes #419
Reverts most of 7033352e

## Checklist


t0yv0 (changes requested):
We need a test that demonstrates how the users are able to write mock tests in their own projects. It is my understanding that the users cannot reference code from `test/`, is that correct? It would seem the PR as it stands removes the ability of our users to do mock testing.

t0yv0:
We could consider publishing another artifact alongside the SDK that would enable writing mock tests, so that the SDK itself does not depend on the mock framework, though we need to weigh carefully if this bit of extra complexity is worth not having the dependency. Alternatively, we could find a way for users to write mock tests for Pulumi without resorting to a mock framework that could work as well. What we're mocking in Pulumi is resource providers that accept input property bags and compute output property bags, naively this can be implemented without a mock framework.

pawelprazak:
Thank you, I was not aware this is something a user would use. In that case, idiomatic solution would be a library, `pulumi-test`.

pawelprazak:
losing in favour of: https://github.com/pulumi/pulumi-java/pull/591

PR #545, opened 2022-05-04
Program gen fixes

Zaid-Ajaj:
# Description

This PR adapts programgen to latest changes:
 - `Config.of()` becomes `ctx.config()`
 - properties no longer have a `.get` prefix (i.e. `.getId()` -> `.id()`) 
 - added `.key()` and `.value()` functions to the `KeyedValue<T>` from sdk (codegen.internals) used in programgen
 - nested builders now use `Args` suffix instead of just the type name, see below

```java
// before
var nodeGroup = new NodeGroup("nodeGroup", NodeGroupArgs.builder()
    .clusterName(eksCluster.name())
    .scalingConfig(NodeGroupScalingConfig.builder()
        .desiredSize(2)
        .build())
    .build());

// After
var nodeGroup = new NodeGroup("nodeGroup", NodeGroupArgs.builder()
    .clusterName(eksCluster.name())
    .scalingConfig(NodeGroupScalingConfigArgs.builder() // NOTICE HERE "Args" suffix
        .desiredSize(2)
        .build())
    .build());
```

## Checklist

### DIFF
--- a/pkg/codegen/java/gen_program_expressions.go
+++ b/pkg/codegen/java/gen_program_expressions.go
@@ -605,13 +605,22 @@ func (g *generator) genObjectConsExpressionWithTypeName(
 		return
 	}
 
+	destTypeName := typeName(destType)
 	switch destType.(type) {
 	case *schema.ObjectType:
 		objectProperties := make(map[string]schema.Type)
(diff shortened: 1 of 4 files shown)


t0yv0 (approved):
Thank you, these are important fixes.

PR #558, opened 2022-05-05
[programgen] Generate custom resource options for resources

### LINKED ISSUE #546: [programgen] generate custom resource options from PCL
### What happened?

At the moment, programgen ignores custom resource options. it should include it when possible

### Steps to reproduce

Compile the following PCL program
```hcl
resource provider "pulumi:providers:aws" {
	region = "us-west-2"
}

resource bucket1 "aws:s3:Bucket" {
	options {
		provider = provider
		dependsOn = [provider]
		protect = true
	}
}
```

### Expected Behavior

Should generate custom resource options for resources

### Actual Behavior

Ignore custom resource options, instead it generates:
```java
    var provider = new Provider("provider", ProviderArgs.builder()
        .region("us-west-2")
        .build());

    var bucket1 = new Bucket("bucket1");
}
```

### Versions used

_No response_

### Additional context

_No response_

### Contributing

Zaid-Ajaj:
# Description

Fixes #546

## Checklist

### DIFF
--- a/pkg/codegen/testing/test/testdata/resource-options-pp/MyStack.java
+++ b/pkg/codegen/testing/test/testdata/resource-options-pp/MyStack.java
@@ -30,9 +30,7 @@ public class App {
             .provider(provider)
             .protect(true)
             .dependsOn(provider)
-            .ignoreChanges(            
-                bucket,
-                lifecycleRules[0])
+            .ignoreChanges("bucket", "lifecycleRules[0]")
             .build());
 
     }
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -388,7 +388,29 @@ func (g *generator) genCustomResourceOptions(w io.Writer, resource *pcl.Resource
 		}
 		if resource.Options.IgnoreChanges != nil {
 			g.genIndent(w)
-			g.Fgenf(w, ".ignoreChanges(%v)", resource.Options.IgnoreChanges)
+			g.Fgen(w, ".ignoreChanges(")
+			switch resource.Options.IgnoreChanges.(type) {
+			case *model.TupleConsExpression:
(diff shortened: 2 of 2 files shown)


pkg/codegen/testing/test/testdata/resource-options-pp/MyStack.java
```diff
@@ -15,7 +16,24 @@ public static void stack(Context ctx) {
...
+                .dependsOn(provider)
+                .build());
+
+        var bucketWithoutArgs = new Bucket("bucketWithoutArgs", BucketArgs.Empty, CustomResourceOptions.builder()
+            .provider(provider)
+            .protect(true)
+            .dependsOn(provider)
+            .ignoreChanges(            
+                bucket,
+                lifecycleRules[0])
```
t0yv0, line 35: I worry this is a dangling reference. What does this refer to in the PCL file even? 

Perhaps these need to be string literals like "bucket"?

```
  @Nullable List<String> ignoreChanges,
```

Zaid-Ajaj: > What does this refer to in the PCL file even?

I don't know to be honest 😅 programgen doesn't typecheck PCL programs but assumes they are already valid

t0yv0: Perhaps we need to ask around here, but it's plausible that PCL lets you write the string literals without quotes. In Java we do have to quote. The above code won't be right.

Zaid-Ajaj: You are right! expressions in `ignoreChanges` have to be stringified (asked in slack). That is, unless they are variable references 😓 that makes life extra complicated

Issue #525, opened 2022-05-02
Identical classes modulo case name in pulumi-azure-native v1.63.0


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It seems that we have two identical enums that differ only in the case of the class name in the schema.

This runs into issues for Mac OS X users, since Mac OS FS is not case-sensitive, running codegen on Mac OS X confuses the two files. One of these enums clobbers the other one.

```
anton@Antons-MacBook-Pro> jq '.types["azure-native:network:IpAllocationMethod"]' ~/schemas/pulumi-azure-native-v1.63.0.json                                        ~/pulumi-java
{
  "description": "Private IP address allocation method.",
  "type": "string",
  "enum": [
    {
      "value": "Static"
    },
    {
      "value": "Dynamic"
    }
  ]
}
anton@Antons-MacBook-Pro> jq '.types["azure-native:network:IPAllocationMethod"]' ~/schemas/pulumi-azure-native-v1.63.0.json                                        ~/pulumi-java
{
  "description": "The private IP address allocation method.",
  "type": "string",
  "enum": [
    {
      "value": "Static"
    },
    {
      "value": "Dynamic"
    }
  ]
}
```

There is a chance it is a schema issue that need to be fixed in pulumi-azure-native to unify the two named types.

pawelprazak:
Additional small complication (reported first by @iwahbe):
```
❯ gh repo clone pulumi/pulumi-java
Cloning into 'pulumi-java'...
remote: Enumerating objects: 1083345, done.
remote: Counting objects: 100% (1490/1490), done.
remote: Compressing objects: 100% (56/56), done.
remote: Total 1083345 (delta 1441), reused 1434 (delta 1434), pack-reused 1081855
Receiving objects: 100% (1083345/1083345), 416.15 MiB | 26.06 MiB/s, done.
Resolving deltas: 100% (1009849/1009849), done.
Updating files: 100% (72733/72733), done.
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'providers/pulumi-azure-native/sdk/java/src/main/java/com/pulumi/azurenative/network/enums/IPAllocationMethod.java'
[...]

PR #560, opened 2022-05-05
Provider update: azure native 1.64.0

t0yv0:
# Description

It looks like this is adding Java examples into Javadocs. This may create a problem with RAM in compile/javadoc publish, but let us see. 

It also seems the new version fixes enum name collision at:

```
 providers/pulumi-azure-native/sdk/java/src/main/java/com/pulumi/azurenative/network/enums/IPAllocationMethod.java
```

Fixes # (issue)

## Checklist


t0yv0:
Might be fixing https://github.com/pulumi/pulumi-java/issues/525 upstream - that is the problem is gone in the new schema. Looks like we have some other problems in here though.

t0yv0:
Not confirmed, #525 needed a separate fix.

pawelprazak (approved):
LGTM

PR #569, opened 2022-05-06
Fix 525 type collision modulo case

### LINKED ISSUE #525: Identical classes modulo case name in pulumi-azure-native v1.63.0

t0yv0:
# Description

Fixes #525 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi-java-gen/dedup.go
+++ b/pkg/cmd/pulumi-java-gen/dedup.go
@@ -60,8 +60,8 @@ func dedupTypes(spec *pschema.PackageSpec) (*pschema.PackageSpec, error) {
 
 	for oldToken, newToken := range renamedTypes {
 		eq := reflect.DeepEqual(
-			transformJsonTree(stripDescription, types[oldToken]),
-			transformJsonTree(stripDescription, types[newToken]),
+			transformJSONTree(stripDescription, types[oldToken]),
+			transformJSONTree(stripDescription, types[newToken]),
 		)
 		if eq {
 			fmt.Printf("WARN renaming %s to %s in the schema\n",
@@ -95,7 +95,7 @@ func dedupTypes(spec *pschema.PackageSpec) (*pschema.PackageSpec, error) {
 		return node
 	}
 
-	rawSchema = transformJsonTree(rewriteRefs, rawSchema)
+	rawSchema = transformJSONTree(rewriteRefs, rawSchema)
 
 	buf.Reset()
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi-java-gen/dedup.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+// Detects cases when identical types have similar names modulo case
+// such as `azure-native:network:IpAllocationMethod` vs
+// `azure-native:network:IPAllocationMethod`, deterministically picks
+// one of these names, and rewrites the schema as if there was only
+// one such type.
+func dedupTypes(spec *pschema.PackageSpec) (*pschema.PackageSpec, error) {
+	normalizedTokens := map[string]string{}
+	for typeToken := range spec.Types {
```
iwahbe, line 22: We do this for type names but not for resource or function tokens?

iwahbe (commented):
I think this works, but I'm concerned with the mechanism:
 It seems problematic to rewrite the raw schema before the schema is loaded, especially since we won't always control how the schema is loaded (TF bridge). I'm also concerned that it will prevent java from using our standard package loading infrastructure (`schema.Loader`). Is this programgen compatible?

t0yv0:
You're correct that programgen would continue emitting mis-spelled code for the affected type. I am inclined to ignore this for the moment. I still think this is the appropriate fix since the blast radius of the bug is exactly N=1 type in azure-native:

- no functions
- no resources
- Java provider generation do not use tfbridge, it uses the schema as input, where tfbrdige was used to make the schema
- https://github.com/pulumi/pulumi-azure-native/issues/1692 would be preferable to fix upstream, then this fix can be dropped

iwahbe (approved):
I buy that. Once azure-native is fixed we can rip this out and hopefully make these collisions illegal at the schema level. 
That may be a 4.0 level change.

PR #573, opened 2022-05-06
Fix typo and use GAV that exist in maven central

maxandersen:
# Description

Fix "coulnd" to "could"

Use pulumi 0.1.0 GAV version as otherwise you can't compile the test projects.


t0yv0 (approved):
Thanks for the PR!

PR #574, opened 2022-05-06
Move out templates

t0yv0:
# Description

Templates are now hosted at pulumi/templates. Try to use those instead of the copies in-repo, while retaining CI checks that see if anything we change in the SDK or providers break those templates.

Fixes # (issue)

## Checklist


dixler (approved):
LGTM

pawelprazak (approved):

Issue #557, opened 2022-05-05
CI setup not ideal for community PRs


t0yv0:
### What happened?

It turns out we rely on GitHub packages to communicate intermediate Java build artifacts between build phases. This is not great for community fork PRs - those do not get the GITHUB token secret divulged, making the CI verification fail.

Example: https://github.com/pulumi/pulumi-java/pull/550

### Steps to reproduce

Open a PR from a forked repo.

### Expected Behavior

CI verifies changes for merge-ability.

### Actual Behavior

CI fails.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
Adopted `/run-acceptance-tests` workflow from pulumi/pulumi.

PR #575, opened 2022-05-06
Add support for using JBang

maxandersen:
#Descripton

Adds support for jbang like there are for maven and gradle.

Advantage is you don't need to have extra files and deep structures if you are just doing some pulumi sripting.

Minimal requirement as implemented here is:

```
src/main.java
Pulumi.yaml
jbang.properties
```

Could even cut it down and look for `src/main.java` if no `jbang.properties` found.

Dependencies are done via `//DEPS`in the .java file, i.e.:

```
//DEPS com.pulumi:pulumi:${env.PULUMI_JAVA_SDK_VERSION:0.1.0}
```

or simply:

```
//DEPS com.pulumi:pulumi:0.1.0
```

notes:
Looks for `jbang.properties` as a marker similar to `pom.xml` and `settings.gradle`.
It is not required for jbang to work, thus we could also just look for a standard source file i.e. `src/main.java`

Also handles `binary: myFile.java` as a way to customize which of possible "scripts" to use.

Also allow using Groovy (.groovy) and Kotlin (.kt) via JBang, but that is untested for now.- Fix typo and use GAV that exist in maven central

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ provider.%.install:	provider.%.build
 # Run a custom integration test or example.
 # Example: make test_example.aws-java-webserver
 test_example.%:	bin/pulumi-language-java
-	cd tests/examples && PATH="${PATH}:${PWD}/bin" go test -run "TestExamples/^$*" -test.v
+	cd tests/examples && PATH="${PATH}:${PWD}/bin" go test -run "TestExamples/^$*$$" -test.v
 
 # Test a single template, e.g.:
 #     make test_template.java-gradle


maxandersen:
Two things I haven't figured out yet:

1) hooking up with the code gen feature. Not sure if relevant/affected

2) how/when "build" command is called by pulumi. Is there a flag or specific sequence of commands to have it call build?

AaronFriel:
Related to this, if we want to add a template: https://github.com/pulumi/templates/pull/283/files

maxandersen:
Yes - I'm happy to complete this just looking for some pointers to my question in the pr description. Thanks!

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -56,15 +57,35 @@ func main() {
...
 	case binary != "":
-		logging.V(3).Infof("language host asked to use specific JAR: `%s`", binary)
-		cmd, err := lookupPath("java")
-		if err != nil {
-			cmdutil.Exit(err)
-		}
-		javaExec, err = newJarExecutor(cmd, binary)
-		if err != nil {
-			cmdutil.Exit(err)
+		logging.V(3).Infof("language host asked to use specific file: `%s`", binary)
```
t0yv0, line 60: "language host asked to use a specific file"

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -56,15 +57,35 @@ func main() {
...
+			cmd, err := probeJBangExecutor()
+			if err != nil {
+				cmdutil.Exit(err)
+			}
+        	javaExec, err = newJBangExecutor(cmd, binary)
+       		if err != nil {
+                cmdutil.Exit(err)
+            }
+		default:
+			cmdutil.Exit(fmt.Errorf("Could not find way to use binary: `%s` with `%s`", binary, suffix))
```
t0yv0, line 84: "Cannot run binary %s with extension %s. Expecting a .jar or a JBang entry-point (.java, .kt, or .groovy)"

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -177,7 +198,7 @@ func (host *javaLanguageHost) determinePulumiPackages(
 	args := host.exec.pluginArgs
 	output, err := host.runJavaCommand(ctx, cmd, args)
 	if err != nil {
-		return nil, errors.Wrapf(err, "language host counld not run plugin discovery command successfully")
+		return nil, errors.Wrapf(err, "language host could not run plugin discovery command successfully")
```
t0yv0, line 201: Nice!

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
 	flag.StringVar(&root, "root", "", "Project root path to use")
 	flag.StringVar(&binary, "binary", "", "A relative or an absolute path to a JAR to execute")
```
t0yv0, line 36: "JAR or a JBang entry-point file to execute"

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -35,6 +35,7 @@ func main() {
...
 
+	logging.V(3).Infof("JBANG FTW!")
```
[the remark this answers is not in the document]
maxandersen: Ha - no we definitely don't :)

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -56,15 +57,35 @@ func main() {
...
+			cmd, err := lookupPath("java")
+			if err != nil {
+				cmdutil.Exit(err)
+			}
+			javaExec, err = newJarExecutor(cmd, binary)
+			if err != nil {
+				cmdutil.Exit(err)
+			}
+		case ".java", ".kt", ".groovy":
+			cmd, err := probeJBangExecutor()
```
t0yv0, line 75: Looks like this pathway is for letting users set `binary:` option in the Pulumi.yaml and have a non-standard entry-point file. Sounds great!

t0yv0:
Thank you for contributing this! I'm just learning about JBang here but I love the idea - it's a great match for short Pulumi programs. Taking it out for a spin shortly.

Please let us know if you can resolve the merge conflicts or need some help with that.

t0yv0:
Also - apologies for CI showing as red.  This is https://github.com/pulumi/pulumi-java/issues/557 - something I will be cleaning up next week.

pkg/cmd/pulumi-language-java/main.go
maxandersen: exactly

maxandersen:
> Please let us know if you can resolve the merge conflicts or need some help with that.

pushed fixes + rebase with main.

maxandersen:
> An executor that does nothing for the "build" command I think should work fine also. It will simply defer the dep fetching until `pulumi up` is called.

so i think it should just work - but didn't find a way to trigger the code thus far.

t0yv0:
Wonderful! Works on my machine, I'll find a way to hitch it up to CI and merge. Thank you!

t0yv0 (approved):

maxandersen:
Want to merge https://github.com/pulumi/templates/pull/283 too to get matching templates ?

t0yv0:
In principle yes, I'd wait until this change gets released next week though.

maxandersen:
I saw 0.3.0 was released but its release notes does not seem to include this change. Was something missing to make it in?

t0yv0:
Included in https://github.com/pulumi/pulumi-java/blob/main/CHANGELOG.md#030-2022-06-01 ? Where is it not propagating?

maxandersen:
https://github.com/pulumi/pulumi-java/releases latest release don't have it in the change log and it's still in the incoming change log as part of 0.4

PR #577, opened 2022-05-09
Git fix attempt

pawelprazak:
# Description

Could the git fix be that simple? I was able to rebase other branches on top successfully.

## Checklist


t0yv0:
I'm not sure I understand what we're fixing, what is this? The net diff of the two commits is a no-op. Is the code on `main` not working properly for you

pawelprazak:
I'm not sure either, but it worked locally, closing.

PR #581, opened 2022-05-10
Add doc comments to overloads of function invokes

Zaid-Ajaj:
# Description

Not all overloads of function invokes generate the doc comments, this PR fixes that

## Checklist


dixler:
did you
```sh
$ PULUMI_ACCEPT=true make codegen_tests
```

Zaid-Ajaj:
> did you `$ PULUMI_ACCEPT=true make codegen_tests`

 @dixler I have now! Fixed the issue and it should be good to go

pawelprazak:
According to the CI, the providers' codegen output has changed but was not versioned.

t0yv0 (approved):
Thank you!

PR #551, opened 2022-05-05
Provide compilation failure details

pawelprazak:
# Description

Fixes #538 by removing `-q`/`--quiet`

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #584, opened 2022-05-11
Fix #540 do not crash when plugin discovery fails

### LINKED ISSUE #540: Unfriendly error when the Java code does not compile
### What happened?

When there is a Java compilation error, we get a poor experience in how it is reported as the plugin discovery step fails. 

### Steps to reproduce

Introduce a compilation error into a working Java project, run `pulumi preview`.

### Expected Behavior

Pulumi reports compilation error in the user program.

### Actual Behavior

Pulumi reports `error: failed to discover plugin requirements: language host could not determine Pulumi packages: language host counld not run plugin discovery command successfully`

```[nix-shell:~/pulumi-java/tests/examples/azure-java-static-website]$ pulumi preview
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/azure-java-static-website/dev/previews/8399198f-5416-4c9a-9729-ca49e6691b7c

     Type                 Name                           Plan     Info
     pulumi:pulumi:Stack  azure-java-static-website-dev           14 messages
 
Diagnostics:
  pulumi:pulumi:Stack (azure-java-static-website-dev):
    /Users/anton/pulumi-java/tests/examples/azure-java-static-website/app/src/main/java/staticwebsite/App.java:1: error: reached end of file while parsing
    BAM
    ^
    1 error
    FAILURE: Build failed with an exception.
    * What went wrong:
    Execution failed for task ':app:compileJava'.
    > Compilation failed; see the compiler error output for details.
    * Try:
    > Run with --stacktrace option to get the stack trace.
    > Run with --info or --debug option to get more log output.
    > Run with --scan to get full insights.
    * Get more help at https://help.gradle.org
    BUILD FAILED in 3s
 
warning: A new version of Pulumi is available. To upgrade from version '3.30.0' to '3.31.0', visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
error: failed to discover plugin requirements: language host could not determine Pulumi packages: language host counld not run plugin discovery command successfully: './gradlew ./gradlew -q run --console=plain -PmainClass=com.pulumi.bootstrap.internal.Main --args=packages' exited with non-zero exit code: 1
```

### Versions used

pulumi-language-java 0.1.0
com.pulumi.pulumi 0.1.0
pulumi cli v3.30.0

### Additional context

_No response_

### Contributing

pawelprazak:
# Description

Based on https://github.com/pulumi/pulumi-java/pull/551#discussion_r866162571

Fixes #540 

## Checklist


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -177,7 +177,10 @@ func (host *javaLanguageHost) determinePulumiPackages(
 	args := host.exec.pluginArgs
 	output, err := host.runJavaCommand(ctx, cmd, args)
 	if err != nil {
-		return nil, errors.Wrapf(err, "language host could not run plugin discovery command successfully")
+		// Plugin determination is an advisory feature so it doe not need to escalate to an error.
+		logging.V(3).Infof("language host could not run plugin discovery command successfully, returning empty plugins; cause: %s", err)
+		logging.Warningf("Plugin autodiscovery failed. More information available at log level 3 and above")
```
pawelprazak, line 182: @t0yv0 I've added a warning, if you are sure we don't need it I'll remove it, but I wanted to avoid a situation where the plugin discovery fails and the user is confused why this blows up in runtime on a missing plugin, esp. important when there are no cached plugins on the users machine yet

pawelprazak:
CI problems: https://github.com/pulumi/pulumi-java/runs/6386726880?check_suite_focus=true#step:11:372

pkg/cmd/pulumi-language-java/main.go
t0yv0: Would it actually blow up on a missing plugin? My understanding is that it will still auto-install it at a later phase which is less desirable but still ok. Also what about programs that do not need plugins at all?

I feel that we should not confuse the user with the error message here, since the message makes it sound like a Pulumi error and does not point to how to resolve it.

In the vast majority of cases we hit this pathway there is a problem with the project, either the code does not compile or builder cannot resolve dependencies or some such thing. The message we show the user should convey that - even if it's unannotated builder output with the build command we ran so the user can reproduce. Since we don't parse the builder output we don't know at the Go level any more detail about the error.

I continue to be inclined this error message should be coming out of the main build pathway not the error discovery pathway (where I'd prefer just hiding things in INFO level), but I'm less firm on that; if the error message is not confusing it can come from here.

pawelprazak: thank you, I've removed the warnings

pawelprazak:
rebased, removed warnings

t0yv0 (approved):
Thanks, LGTM!

PR #588, opened 2022-05-12
Enhance pulumi-java-gen default gradle buildfiles

### LINKED ISSUE #583: Enable ability to add Java support to external repos
As part of the post Java preview launch, we now need to add the ability to allow 1st and 3rd part providers to add Java support to their providers 

we have decided on centralising all of the Java logic in a Central Java repo and building a generation binary that is available for download for use in our package repos

Currently, the Java-gen binary takes a config file necessary to create the SDK code. Most of those arguments are not going to be required as part of the generation within a specific provider repository. Therefore, we should allow the Java gen binary to accept cli arguments of provider name, version (reading from environment variable would be amazing) and a list of dependencies (when an MLC for example).

The Java gen binary should be able to generate the Graeme / maven build files as part of the codegen so they don’t need to be manually created

The process for Java gen would be to download the latest binary in each parameter (the binary should adhere to Semver!) and be in control of the dependencies needed for Java SDK generation

we then also need to understand the secrets and build steps needed to build + publish the Java binaries

t0yv0:
# Description

Fixes #583 

How I tested:

```
#!/usr/bin/env bash

set -euo pipefail

mkdir -p bin

(cd pkg && go build -o ../bin github.com/pulumi/pulumi-java/pkg/cmd/pulumi-java-gen)

./bin/pulumi-java-gen \
    -schema https://raw.githubusercontent.com/pulumi/pulumi-random/v4.6.0/provider/cmd/pulumi-resource-random/schema.json \
    -out providers/pulumi-random/sdk/java \
    -override override.json \
    -version 4.6.0

(cd providers/pulumi-random/sdk/java && gradle build)

export SIGNING_KEY=$(cat secret.key)
export SIGNING_PASSWORD=$(pass gpg/support@pulumi.com/passphrase)

export PUBLISH_REPO_URL=https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/
export PUBLISH_REPO_USERNAME=t0yv0
export PUBLISH_REPO_PASSWORD=$(pass issues.sonatype.org/t0yv0/anton@pulumi.com)

(cd providers/pulumi-random/sdk/java && gradle publish)

```

Given `override.json`:

```
{
    "buildFiles": "gradle",
    "packages": {
        "com.pulumi:pulumi": "0.0.1"
    }
}
```

The override.json could just be a part of the schema under `.language.java`. Unfortunately currently published schemas do not yet include Java options so mixing it in from the side seemed convenient.'

The `buildFiles: gradle` is important because build files generation is opt-in. We should really support Maven but at this point I'm a lot more familiar with Gradle so I started with it.

## Checklist

### DIFF
--- a/pkg/cmd/pulumi-java-gen/old.go
+++ b/pkg/cmd/pulumi-java-gen/old.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Old style main. The style of invocation wtih -config argument
+// Old style main. The style of invocation with -config argument
 // pointing to a YAML-encoded Config file will be phased out when all
 // provider builds move out of the pulumi/pulumi-java repo.
 
--- a/pkg/codegen/java/gen_test.go
+++ b/pkg/codegen/java/gen_test.go
@@ -3,7 +3,6 @@
 package java
 
 import (
-	"os"
 	"testing"
 
 	"github.com/pulumi/pulumi/pkg/v3/codegen"


pkg/codegen/java/build.gradle.template
```diff
@@ -0,0 +1,121 @@
...
+
+plugins {
+    id("signing")
+    id("java-library")
+    id("maven-publish")
+}
+
+version = "{{ .Version }}"
+
+def signingKey = System.getenv("SIGNING_KEY")
```
t0yv0, line 12: These are the environment variables that enable `gradle publish` to work.

pkg/codegen/java/templates_gradle.go
```diff
@@ -0,0 +1,112 @@
...
+	if _, got := ctx.Dependencies[jsr305]; !got {
+		ctx.Dependencies[jsr305] = "3.0.2"
+	}
+
+	if pkg.License == "Apache-2.0" {
+		ctx.LicenceName = "The Apache License, Version 2.0"
+		ctx.LicenceURL = "http://www.apache.org/licenses/LICENSE-2.0.txt"
+	}
+
+	if pkg.Publisher == "Pulumi" ||
```
t0yv0, line 86: Heavy defaulting here. Might need to provide a pathway for the user to set this metadata in `.language.java` blob when the publisher is not Pulumi.

t0yv0:
Looks like I need a bit more work to make it no-op for existing provider+CI setup in place here, but the new pathway can be reviewed already.

stack72:
@t0yv0 as the version of the binary and the version of the package in the override.json is the same, can we cut the need for passing this value? If so then we could remove the need for override.json completely and pass a build flag (--build=gradle) and that makes this even easier :)

t0yv0:
In terms of releasing this, any thoughts - should I just add this as an artifact to the GitHub releases for the repo? Or should perhaps bundle the code-generation facility inside pulumi-language-java? Conceptually at least it makes sense.

stack72:
> In terms of releasing this, any thoughts - should I just add this as an artifact to the GitHub releases for the repo? Or should perhaps bundle the code-generation facility inside pulumi-language-java? Conceptually at least it makes sense.

I'd be more of a fan of having a specific gen binary (much like tfgen is) for this as it keeps the language binary separate

stack72:
For me this is a +1 but this is more a review for @mikhailshilkov :)

t0yv0:
Sounds good! Pawel can also review for consistency Java-wise. I'd like to err on the side of checking in and releasing, fully expecting to iterate a few more times on this as y'all try it out.

t0yv0:
I'll for just adding this binary to the GoReleaser and GitHub releases.

mikhailshilkov (approved):
I skimmed through the changes and it LGTM. Let's make it work for all the providers :shipit:

PR #589, opened 2022-05-12
Bump awsx 1.0.0-beta.1 -> 1.0.0-beta.2

pawelprazak:
# Description

## Checklist


t0yv0 (approved):
Good, this updates to v1.0.0-beta.2 from source. Good to merge. We need a ticket to actually release this since there's still some manual steps involved documented on the wiki. Also seems like now there is v1.0.0-beta.3 also.

PR #591, opened 2022-05-16
Remove dependency on Mockito in SDK

### LINKED ISSUE #419: Remove SDK dependency on Mockito

pawelprazak:
# Description

- add manual mocks
- remove Mockito dep in SDK

Fixes #419

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockDeployment.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockDeployment.java
@@ -1,0 +1,156 @@
+package com.pulumi.deployment;
+
+import com.pulumi.core.Output;
+import com.pulumi.core.TypeShape;
+import com.pulumi.core.internal.annotations.InternalUse;
+import com.pulumi.deployment.CallOptions;
+import com.pulumi.deployment.Deployment;
+import com.pulumi.deployment.InvokeOptions;
+import com.pulumi.deployment.internal.DeploymentImpl;
+import com.pulumi.deployment.internal.DeploymentInstanceHolder;
+import com.pulumi.deployment.internal.DeploymentInternal;
+import com.pulumi.deployment.internal.Runner;
+import com.pulumi.resources.CallArgs;
+import com.pulumi.resources.InvokeArgs;
+import com.pulumi.resources.Resource;
+import com.pulumi.resources.ResourceArgs;
+import com.pulumi.resources.ResourceOptions;
+import com.pulumi.resources.Stack;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Simple mock of {@link Deployment} that does mostly nothing.
+ */
+@InternalUse
+public class MockDeployment extends DeploymentInstanceHolder implements Deployment, DeploymentInternal {
+    public final DeploymentImpl.DeploymentState state;
+
+    public MockDeployment(DeploymentImpl.DeploymentState state) {
+        this.state = requireNonNull(state);
+    }
+
+    @Nonnull
+    @Override
+    public String getStackName() {
+        return this.state.stackName;
+    }
+
+    @Nonnull
+    @Override
+    public String getProjectName() {
+        return this.state.projectName;
+    }
+
+    @Override
+    public boolean isDryRun() {
+        return this.state.isDryRun;
+    }
+
+    @Override
+    public <T> Output<T> invoke(String token, TypeShape<T> targetType, InvokeArgs args, @Nullable InvokeOptions options) {
+        return null;
+    }
+
+    @Override
+    public <T> Output<T> invoke(String token, TypeShape<T> targetType, InvokeArgs args) {
+        return null;
+    }
+
+    @Override
+    public <T> CompletableFuture<T> invokeAsync(String token, TypeShape<T> targetType, InvokeArgs args, InvokeOptions options) {
+        return null;
+    }
+
+    @Override
+    public <T> CompletableFuture<T> invokeAsync(String token, TypeShape<T> targetType, InvokeArgs args) {
+        return null;
+    }
+
+    @Override
+    public CompletableFuture<Void> invokeAsync(String token, InvokeArgs args, InvokeOptions options) {
+        return null;
+    }
+
+    @Override
+    public CompletableFuture<Void> invokeAsync(String token, InvokeArgs args) {
+        return null;
+    }
+
+    @Override
+    public <T> Output<T> call(String token, TypeShape<T> targetType, CallArgs args, @Nullable Resource self, @Nullable CallOptions options) {
+        return null;
+    }
+
+    @Override
+    public <T> Output<T> call(String token, TypeShape<T> targetType, CallArgs args, @Nullable Resource self) {
+        return null;
+    }
+
+    @Override
+    public <T> Output<T> call(String token, TypeShape<T> targetType, CallArgs args) {
+        return null;
+    }
+
+    @Override
+    public void call(String token, CallArgs args, @Nullable Resource self, @Nullable CallOptions options) {
+        // Empty
+    }
+
+    @Override
+    public void call(String token, CallArgs args, @Nullable Resource self) {
+        // Empty
+    }
+
+    @Override
+    public void call(String token, CallArgs args) {
+        // Empty
+    }
+
+    @Override
+    public DeploymentImpl.Config getConfig() {
+        return this.state.config;
+    }
+
+    @Override
+    public Optional<String> getConfig(String fullKey) {
+        return this.state.config.getConfig(fullKey);
+    }
+
+    @Override
+    public boolean isConfigSecret(String fullKey) {
+        return this.state.config.isConfigSecret(fullKey);
+    }
+
+    @Override
+    public Stack getStack() {
+        return null;
+    }
+
+    @Override
+    public void setStack(Stack stack) {
+        // Empty
+    }
+
+    @Override
+    public Runner getRunner() {
+        return this.state.runner;
+    }
+
+    @Override
+    public void readOrRegisterResource(Resource resource, boolean remote, Function<String, Resource> newDependency, ResourceArgs args, ResourceOptions opts, Resource.LazyFields lazy) {
+        // Empty
+    }
+
+    @Override
+    public void registerResourceOutputs(Resource resource, Output<Map<String, Output<?>>> outputs) {
+        // Empty
+    }
+}
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockRunner.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockRunner.java
@@ -1,0 +1,39 @@
+package com.pulumi.deployment;
+
+import com.pulumi.core.Output;
+import com.pulumi.deployment.internal.Runner;
+import com.pulumi.resources.Stack;
+import com.pulumi.resources.StackOptions;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+
+public class MockRunner implements Runner {
+
+    @Override
+    public List<Exception> getSwallowedExceptions() {
+        return null;
+    }
+
+    @Override
+    public <T> void registerTask(String description, CompletableFuture<T> task) {
+        // Empty
+    }
+
+    @Override
+    public CompletableFuture<Integer> runAsyncFuture(Supplier<CompletableFuture<Map<String, Output<?>>>> callback) {
+        return null;
+    }
+
+    @Override
+    public CompletableFuture<Integer> runAsyncFuture(Supplier<CompletableFuture<Map<String, Output<?>>>> callback, StackOptions options) {
+        return null;
+    }
+
+    @Override
+    public <T extends Stack> CompletableFuture<Integer> runAsync(Supplier<T> stackFactory) {
+        return null;
+    }
+}
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
@@ -7,15 +7,15 @@ import com.pulumi.Context;
 import com.pulumi.Log;
 import com.pulumi.core.Output;
 import com.pulumi.core.internal.Internal;
+import com.pulumi.deployment.MockDeployment;
 import com.pulumi.deployment.MockEngine;
 import com.pulumi.deployment.MockMonitor;
+import com.pulumi.deployment.MockRunner;
 import com.pulumi.deployment.Mocks;
 import com.pulumi.deployment.internal.DeploymentImpl.DefaultEngineLogger;
 import com.pulumi.exceptions.RunException;
 import com.pulumi.resources.Resource;
 import com.pulumi.resources.Stack;
-import org.mockito.ArgumentMatchers;
-import org.mockito.Mockito;
 
 import javax.annotation.Nullable;
 import java.util.List;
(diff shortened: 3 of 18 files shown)


sdk/java/pulumi/src/test/java/com/pulumi/core/UrnTest.java
```diff
@@ -20,7 +20,7 @@ public static void mockSetup() {
         mock = DeploymentTests.DeploymentMockBuilder.builder()
             .setMocks(new MocksTest.MyMocks())
             .setOptions(new TestOptions(false))
-            .setMockGlobalInstance();
+            .build();
```
t0yv0, line 23: Does it not actually set the global instance still? The previous name was more suggestive.

t0yv0:
CI failure seems unrelated; odd though. Can you add a no-op commit to try again on a different version? We can also delete this version manually from GPR repo.

```
* What went wrong:
Execution failed for task ':pulumi:publishGprPublicationToGitHubPackagesRepository'.
> Failed to publish publication 'gpr' to repository 'GitHubPackages'
   > Could not PUT 'https://maven.pkg.github.com/pulumi/pulumi-java/com/pulumi/pulumi/0.1.1-alpha.1652715321+66c4d102/pulumi-0.1.1-alpha.1652715321+66c4d102.jar.sha1'. Received status code 409 from server: Conflict
```

t0yv0:
I worry about usability. It would be lovely if this PR came with an *example* outside of the SDK project of how we expect the user to do the mocking. I am basically worried about the overhead there, or perhaps we are expecting the user to do mocking frameworks? The new MockDeployment class is for some reason Internal and in the test package again, so the user cannot rely on that being available. A new project under tests/examples would show exactly how to do it.

sdk/java/pulumi/src/test/java/com/pulumi/core/UrnTest.java
pawelprazak: yes, yes it does, a builder not ending with build feels weird, as a target API that I'm working on I was thinking about:
```
PulumiTest.withOptions(new TestOptions(false))
  .monitorMocks(new MocksTest.MyMocks())
  .useMockRunner()
  .build()
```
The fact that we set a global singleton is an implementation detail that is also true for the `Pulumi.run` itself, and not specific to the test API, so we should document it but making is as part of API does not feel like thinking ahead, esp. since we may change this in the future.

pawelprazak:
> I worry about usability. It would be lovely if this PR came with an _example_ outside of the SDK project of how we expect the user to do the mocking. I am basically worried about the overhead there, or perhaps we are expecting the user to do mocking frameworks? The new MockDeployment class is for some reason Internal and in the test package again, so the user cannot rely on that being available. A new project under tests/examples would show exactly how to do it.

I agree. the goal of this PR is not to address user facing testing and mocking, that will require a lot more work, I have a branch with ongoing work on that.

I've refactored the tests and the test framework and discovered a bunch of "hot points" and now I'm reimplementing it in a logical order an with improvements informed by the end result of this effort.

I wanted to avoid having a 2000 LOC added and 2100 LOC removed type of PR that would be unreviewable.

I've distilled a minimal subset of those changes to address this specific issue of Mockito dep.

sdk/java/pulumi/src/test/java/com/pulumi/core/UrnTest.java
t0yv0: OK then.

t0yv0:
> I agree. The goal of this PR is not to address user facing testing and mocking, that will require a lot more work, I have a branch with ongoing work on that.

User facing testing and mocking is possible with the code on `main` by copy-pasting our mocking examples and having those work. With these changes but without an example I'm not sure this feature set is preserved. But perhaps it is?

> I wanted to avoid having a 2000 LOC added and 2100 LOC removed type of PR that would be unreviewable.

Happy to review that and larger PRs, there's techniques to review commit-by-commit or stack PRs if needed. This really isn't a problem.

> I've distilled a minimal subset of those changes to address this specific issue of Mockito dep.

That's great! D'you mind withholding the merge until we have evidence users can still mock-test?

>  that will require a lot more work, 

Can we brainstorm on approaches that perhaps do not require so much work? Perhaps something along the lines of exposing the currently-private/internal mock classes to the users in the SDK would make mock testing still possible? Or there's some disadvantages there? Curious.

pawelprazak:
rebased on main

t0yv0:
Could you merge and fix the conflicts that'd be much appreciated!

As discussed before, I won't block this if there's a technical way that remains for users to write mock tests. I tried mock-testing a simple bucket-creating program using this branch. It's possible. It's not pretty, but it's a starting point:

```java
package myproject;

import com.google.common.collect.ImmutableMap;
import com.pulumi.Config;
import com.pulumi.Context;
import com.pulumi.aws.s3.Bucket;
import com.pulumi.context.LoggingContext;
import com.pulumi.core.Tuples;
import com.pulumi.core.internal.Internal;
import com.pulumi.deployment.MockCallArgs;
import com.pulumi.deployment.MockResourceArgs;
import com.pulumi.deployment.Mocks;
import com.pulumi.deployment.internal.DeploymentImpl;
import com.pulumi.deployment.internal.DeploymentTests;
import com.pulumi.deployment.internal.TestOptions;
import com.pulumi.resources.Stack;
import org.junit.jupiter.api.Test;

import java.util.Map;
import java.util.Objects;
import java.util.Optional;

import static org.assertj.core.api.Assertions.assertThat;

public class AppTest {
    @Test
    public void testFoo() {
        // Pain: can we make this simpler?
        var mock = DeploymentTests.DeploymentMockBuilder.builder()
                .setOptions(new TestOptions(false))
                .setMocks(new MyMocks())
                .deploymentFactory(DeploymentImpl::new)
                .build();

        // Pain: need to implement and supply TestContext, wire it.
        var result = mock.runAsync(() -> {
            var ctx = new TestContext();
            App.init(ctx);
            return CompletableFuture.completedFuture(ctx.getStackOutputs());
        }).join();

        assertThat(result.exceptions).isEmpty();
        assertThat(result.resources.size()).isEqualTo(2);

        // Pain: finding resources is manual.
        Bucket bucket = (Bucket) result.resources.stream().filter(r -> r instanceof Bucket).findFirst().get();

        // Pain: checking outputs is non-trivial.
        var bucketData = Internal.of(bucket.bucket()).getDataAsync().join();
        assertThat(bucketData.isKnown()).isTrue();
        assertThat(bucketData.getValueNullable())
                .isEqualTo("my-bucket");

        // Blocker: can find Stack but can't find Stack outputs.
        Stack stack = (Stack) result.resources.stream().filter(r -> r instanceof Stack).findFirst().get();
    }

    public static class MyMocks implements Mocks {
        @Override
        public CompletableFuture<Tuples.Tuple2<Optional<String>, Object>> newResourceAsync(MockResourceArgs args) {
            Objects.requireNonNull(args.type);
            switch (args.type) {
                case "aws:s3/bucket:Bucket":
                    return CompletableFuture.completedFuture(
                            Tuples.of(Optional.of("i-1234567890abcdef0"),
                                    Map.of("bucket", "my-bucket"))
                    );
                default:
                    throw new IllegalArgumentException(String.format("Unknown resource '%s'", args.type));
            }
        }

        @Override
        public CompletableFuture<Map<String, Object>> callAsync(MockCallArgs args) {
            return CompletableFuture.completedFuture(null);
        }
    }

    // TODO: use real context after refactoring test framework (and as a consequence DefaultRunner, and PulumiInternal)
    class TestContext implements Context {
        private final ImmutableMap.Builder<String, Output<?>> exports = ImmutableMap.builder();

        public Map<String, Output<?>> getStackOutputs() {
            return exports.build();
        }

        @Override
        public String projectName() {
            throw new IllegalStateException("Not implemented yet: projectName()");
        }

        @Override
        public String stackName() {
            throw new IllegalStateException("Not implemented yet: stackName()");
        }

        @Override
        public LoggingContext log() {
            throw new IllegalStateException("Not implemented yet: log()");
        }

        @Override
        public Context export(String name, Output<?> output) {
            Objects.requireNonNull(name, "The 'name' of an 'export' cannot be 'null'");
            Objects.requireNonNull(output, "The 'output' of an 'export' cannot be 'null'");
            this.exports.put(name, output);
            return this;
        }

        @Override
        public Config config() {
            throw new IllegalStateException("Not implemented yet: config()");
        }

        @Override
        public Config config(String name) {
            throw new IllegalStateException("Not implemented yet: config(String name)");
        }

        @Override
        public <T> Output<T> output(T value) {
            return Output.of(value);
        }
    }
}
```

t0yv0:
@pawelprazak possible to do another rebase? Sorry looks like tons of conflict..

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockDeployment.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Simple mock of {@link Deployment} that does mostly nothing.
+ */
+@InternalUse
+public class MockDeployment extends DeploymentInstanceHolder implements Deployment, DeploymentInternal {
```
pawelprazak, line 33: not sure if we'll need this class in the end, but right now it does exactly what a mockito mock would do

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockRunner.java
```diff
@@ -0,0 +1,39 @@
...
+import com.pulumi.deployment.internal.Runner;
+import com.pulumi.resources.Stack;
+import com.pulumi.resources.StackOptions;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+
+public class MockRunner implements Runner {
```
pawelprazak, line 13: not sure if we'll need this class in the end, but right now it does exactly what a mockito mock would do

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -296,48 +304,26 @@ private void initUnset() {
...
-                    this.state.config.getConfig((String) invocation.getArguments()[0])
-            );
-            Mockito.when(mock.isConfigSecret(ArgumentMatchers.anyString())).then(invocation ->
-                    this.state.config.isConfigSecret((String) invocation.getArguments()[0])
-            );
-            Mockito.when(mock.getRunner()).thenReturn(this.runner);
-
-            this.deployment = mock;
+            if (this.deploymentFactory == null) {
+                this.deploymentFactory = DeploymentImpl::new;
```
pawelprazak, line 318: I've made real runner and real deployment the default, this is what a user would like to use, we might want to use a MockDeployment internally but this feels like the right default

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/ConverterTests.java
```diff
@@ -226,7 +227,8 @@ void testNullInPreviewProducesFalseKnown() {
             DeploymentTests.DeploymentMockBuilder.builder()
                     .setMocks(new MocksTest.MyMocks())
                     .setOptions(new TestOptions(true))
-                    .setMockGlobalInstance();
+                    .deploymentFactory(MockDeployment::new)
```
pawelprazak, line 230: we choose to mock the deployment in cases where we know we don't need it, since it requires knowledge about internals I've decided this should not be the default anymore, so this is done explicitly now

pawelprazak (commented):
I've rewrote the PR on top of main (was easier than regular rebase).

This PR's goal is to only remove the Mockito deps, not write a new test framework.

The only semantic change is that now the user gets real deployment and runner by default, this is what I expect to be the only non-internal option in the future simplified public API.

t0yv0:
It's a lot cleaner for user mock testing since the changes landed on master. Nice! User does not have to implement Context anymore.

```java
package myproject;

import com.google.common.collect.ImmutableMap;
import com.pulumi.Config;
import com.pulumi.Context;
import com.pulumi.aws.s3.Bucket;
import com.pulumi.context.LoggingContext;
import com.pulumi.core.Tuples;
import com.pulumi.core.internal.Internal;
import com.pulumi.deployment.MockCallArgs;
import com.pulumi.deployment.MockResourceArgs;
import com.pulumi.deployment.Mocks;
import com.pulumi.deployment.internal.DeploymentImpl;
import com.pulumi.deployment.internal.DeploymentTests;
import com.pulumi.deployment.internal.TestOptions;
import com.pulumi.resources.Stack;
import org.junit.jupiter.api.Test;

import java.util.Map;
import java.util.Objects;
import java.util.Optional;

import static org.assertj.core.api.Assertions.assertThat;

public class AppTest {
    @Test
    public void testFoo() {
        // Pain: can we make this simpler?
        var mock = DeploymentTests.DeploymentMockBuilder.builder()
                .setOptions(new TestOptions(false))
                .setMocks(new MyMocks())
                // Can we just not customize deploymentFactory here?
                // .deploymentFactory(DeploymentImpl::new)
                .build();

        var result = mock.runTestAsync(App::init).join();

        assertThat(result.exceptions).isEmpty();
        assertThat(result.resources.size()).isEqualTo(2);

        // Pain: finding resources is manual.
        Bucket bucket = (Bucket) result.resources.stream().filter(r -> r instanceof Bucket).findFirst().get();

        // Pain: awaiting outputs is non-trivial.
        var bucketData = Internal.of(bucket.bucket()).getDataAsync().join();
        assertThat(bucketData.isKnown()).isTrue();
        assertThat(bucketData.getValueNullable())
                .isEqualTo("my-bucket");

        var stackOut = Internal.of(result.stackOutput("bucketName", String.class)).getDataAsync().join();
        assertThat(stackOut.getValueNullable()).isEqualTo("my-bucket");
    }

    public static class MyMocks implements Mocks {
        @Override
        public CompletableFuture<Tuples.Tuple2<Optional<String>, Object>> newResourceAsync(MockResourceArgs args) {
            Objects.requireNonNull(args.type);
            switch (args.type) {
                case "aws:s3/bucket:Bucket":
                    return CompletableFuture.completedFuture(
                            Tuples.of(Optional.of("i-1234567890abcdef0"),
                                    Map.of("bucket", "my-bucket"))
                    );
                default:
                    throw new IllegalArgumentException(String.format("Unknown resource '%s'", args.type));
            }
        }

        @Override
        public CompletableFuture<Map<String, Object>> callAsync(MockCallArgs args) {
            return CompletableFuture.completedFuture(null);
        }
    }
}
```

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockDeployment.java
t0yv0, line 33: Would it ever need to be public? Would users need to access it?

t0yv0 (approved):
Looks good - left a few questions.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockDeployment.java
pawelprazak: Those are great questions, I assume the answer is no, but I didn't want to bundle this decision with this PR.

I assume only `Mocks` will ever be exposed to users, and I'd love to refactor the package structure to refelct that.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockRunner.java
pawelprazak: Great that you've come to the same conclusions independently, this gives me confidence in this decision.

Yes, I totally agree we should make all mocks except the `Mocks` (monitor mocks interface and maybe a default impl `EmptyMocks`) internal.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -261,7 +261,14 @@ public DeploymentMockBuilder setLog(Log log) {
             return this;
         }
 
-        private void initUnset() {
+        public DeploymentMockBuilder deploymentFactory(
```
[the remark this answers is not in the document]
pawelprazak: Yes, only hard-core internal testing. User is unlikely to ever need that.

My proposal is to have `PulumiTest` interface with the public API and then it's implementation, `PulumiTestInternal` class would be used for internal tests that need to manipulate the internals to simulate various failure modes etc.

Issue #381, opened 2022-04-13
Fix generated package and class names for versioned APIs


pawelprazak:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently when generating providers we don't handle versioned APIs very well (kubernetes, google-native, others?).

Expectations:
- `io.pulumi.googlenative.accesscontextmanager_v1` -> `io.pulumi.googlenative.accesscontextmanager.v1`
- `Accesscontextmanager_v1Functions ` -> `AccesscontextmanagerFunctions`

I've seen this mechanism used to correct this type of provider specific problems:
https://github.com/VirtusLab/pulumi-kubernetes/commit/2e08c5e7ded01eee32832f9fbfefadad9a2a375c#diff-8dfd74ae4dfdfd04f5c226c8e65a2d3060d6761b45252ba7a757c22492d99311R42460

Underscores in packages should only be used to escape reserved keywords and when the name starts with a number, e.g.:
- `org.example.foo.class_.for_.void_`
- `org/example.bar._23games`

t0yv0:
What does the C# projection do and how does it do it?

Found this:

https://github.com/pulumi/pulumi-google-native/blob/351aaf086ff1e630cda07b54a7f623dce8c4a6db/sdk/dotnet/AccessContextManager/V1/Inputs/DevicePolicyArgs.cs

VS this:

https://github.com/pulumi/pulumi-java/tree/main/providers/pulumi-google-native/sdk/java/src/main/java/com/pulumi/googlenative/accesscontextmanager_v1

C# has this language-specific config:

```
anton@Antons-MacBook-Pro> jq .language.csharp ~/schemas/pulumi-kubernetes-v3.15.1.json

{
  "compatibility": "kubernetes20",
  "dictionaryConstructors": true,
  "namespaces": {
    "": "Provider",
    "admissionregistration.k8s.io/v1": "AdmissionRegistration.V1",
    "admissionregistration.k8s.io/v1beta1": "AdmissionRegistration.V1Beta1",
    "apiextensions": "ApiExtensions",
    "apiextensions.k8s.io/v1": "ApiExtensions.V1",
    "apiextensions.k8s.io/v1beta1": "ApiExtensions.V1Beta1",
    "apiregistration.k8s.io/v1": "ApiRegistration.V1",
[...]

PR #596, opened 2022-05-16
[programgen] Implement fully qualified imports

### LINKED ISSUE #547: [programgen] generate fully qualified imports from PCL programs
## Hello!

## Issue details

Right now, programgen writes these imports:
```java
import java.util.*;
import java.io.*;
import java.nio.*;
import com.pulumi.*;
```

It would great if it generated the fully qualified names of the resources that it is using. i.e. 
```java
import java.util.List;
import java.util.Map;
import java.nio.Paths;
import com.pulumi.aws.s3.Bucket;
import com.pulumi.aws.s3.BucketArgs;
// etc.
```

### Affected area/feature

programgen

Zaid-Ajaj:
# Description

Fixes #547 

This PR recursively searches resource and variable definitions from PCL and collects imports from them. Those imports are added at the start of the generated file (see diff)

Also small fix for `genApply` where it opts for for `.applyValue(...)` instead of `.apply(...)` because that is what is needed most of the time but sometimes `.apply(x -> x.y)` is needed when `y` is `Output<T>` but still need to check for that 

## Checklist

### DIFF
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -252,6 +252,7 @@ func ignoreModule(module string) bool {
 
 // Removes dashes and replaces slash with underscore
 func sanitizeImport(name string) string {
+	// TODO to be revised when https://github.com/pulumi/pulumi-java/issues/381 is resolved
 	// e.g. azure-native becomes azurenative
 	withoutDash := strings.ReplaceAll(name, "-", "")
 	// e.g. kubernetes.core/v1 becomes kubernetes.core_v1


pkg/codegen/java/gen_program.go
```diff
@@ -229,33 +229,334 @@ func containConfigVariables(nodes []pcl.Node) bool {
...
+								fullyQualifiedTypeName := objectType.Token
+								nameParts := strings.Split(fullyQualifiedTypeName, ":")
+								objectTypeName := names.Title(nameParts[len(nameParts)-1])
+								if objectType.IsInputShape() {
+									objectTypeName = objectTypeName + "Args"
+								}
+								imports = append(imports, pulumiInputImport(pkg, module, objectTypeName))
+								for _, importDef := range collectObjectImports(object, objectType) {
+									imports = append(imports, importDef)
+								}
```
iwahbe, line 352: It looks like this block repeats 5-6 times. Is there a reason why it couldn't be refactored out?

Zaid-Ajaj: Yeah I will try refactor into a separate method ✅

pkg/codegen/java/gen_program.go
```diff
@@ -229,33 +229,295 @@ func containConfigVariables(nodes []pcl.Node) bool {
...
+// Decides when to ignore a module in import definition
+func ignoreModule(module string) bool {
+	return module == "" || module == "index"
+}
+
+// Removes dashes and replaces slash with underscore
+func sanitizeImport(name string) string {
+	// e.g. azure-native becomes azurenative
+	withoutDash := strings.ReplaceAll(name, "-", "")
+	// e.g. kubernetes.core/v1 becomes kubernetes.core_v1
```
t0yv0, line 257: Link to https://github.com/pulumi/pulumi-java/issues/381 TODO as this might change.

```
// TODO[pulumi/pulumi-java#381]
```

t0yv0:
Sorry for the slow review.

> Also small fix for genApply where it opts for for .applyValue(...) instead of .apply(...) because that is what is needed most of the time but sometimes .apply(x -> x.y) is needed when y is Output<T> but still need to check for that

Separate PRs for this kind of stuff are super appreciated, though I understand the desire to piggy-back on the exact same PR!

The functionality looks good, much needed! Thank you!

On the implementation side there are a few aspects that make me significantly uneasy. I do not want to block on it but wanted to pick your brain, is this something we could refactor or there's aspects of the problem I don't understand.

1. Type mapping and naming. So you have a logical type, and you want to map it to a Java type. We have a ton of heuristics there in other non-java providers and I hate to go down the same pathway here. If we possibly could factor it out into a shared module so we could share this between sdk-gen and program-gen? We even had a proposal somewhere to store these mappings as part of the provider. Basically, here be dragons.  In sdk-gen this essentially maps schema.Type to TypeShape (which is built up of FQNs that give you fully qualified Java names). Does programgen model use schema.Type or it uses something else? Really interested in factoring this out, when it falls out of sync we gen invalid programs.

2. Import generation (less important). You're going down the pathway of walking the AST to find the type-set and compute imports from it. This in a way is more "pure" but with the caveat that the AST walker has to be bug-free and has to match the use pattern of the actual code generator itself. I have had very similar problem in sdk-generation. In all languages except Java, it also does essentially your approach here. I have since switched to a more "single-pass" approach in sdk-generation: it just generates the SDK into an in-memory byte buffer as it would and it uses https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/gen_classfile.go and https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/names/imports.go#L57 helper class; then at the end it concatenates the effective imports from the imports class with the generated SDK code proper. The advantage here is that imports are very precise - it will only be imported if it's referenced; and the imports class takes care of deciding to use long or short name, and track collisions. Would something like this be useful here as well?

3. Visitors - if you do stick with the approach of traversing the AST to compute imports, are there any pre-baked PCL AST visitors that the code could use instead of recursing manually? If those are maintained they'd be sure to cover all cases. But perhaps not applicable here.

t0yv0 (approved):

Zaid-Ajaj:
Hi @t0yv0, thanks a lot for reviewing the code 🙏  these are very much valid concerns so let us talk about them:

> In sdk-gen this essentially maps schema.Type to TypeShape (which is built up of FQNs that give you fully qualified Java names). Does programgen model use schema.Type or it uses something else?

program-gen does use `schema.Type`, more specifically `schema.ObjectType` and generates the imports from it and its properties that are also of type `schema.ObjectType` (the recursive bit). The import derivation is this bit:

```go
// objectType is *schema.ObjectType 

imports := make([]string, 0)
// add imports of the type itself
fullyQualifiedTypeName := objectType.Token
nameParts := strings.Split(fullyQualifiedTypeName, ":")
objectTypeName := names.Title(nameParts[len(nameParts)-1])
pkg := nameParts[0]
module := nameParts[1]
if objectType.IsInputShape() {
	if !strings.HasSuffix(objectTypeName, "Args") {
		objectTypeName = objectTypeName + "Args"
	}
	imports = append(imports, pulumiInputImport(pkg, module, objectTypeName))
} else {
	imports = append(imports, pulumiImport(pkg, module, objectTypeName))
}
```

> If we possibly could factor it out into a shared module so we could share this between sdk-gen and program-gen?

I wasn't able to find something that I could use right off the bat because the functions were part of some instance but I could have another look in sdk-gen. I too would love to the share the logic there 💯 

> You're going down the pathway of walking the AST to find the type-set and compute imports from it. This in a way is more "pure" but with the caveat that the AST walker has to be bug-free and has to match the use pattern of the actual code generator itself. [...] Would something like this be useful here as well?

I personally prefer the approach of the AST walker because it self-contained and easier to follow. Single-pass approach is definitely faster and more memory efficient though I don't think this is a problem here 

> Visitors - if you do stick with the approach of traversing the AST to compute imports, are there any pre-baked PCL AST visitors that the code could use instead of recursing manually?

Actually, this is a great point! Because there is a `VisitExpression` PCL function which I would have loved to use as follows: 
```go
imports := make([]string, 0)
for _, node := range nodes {
	diags := node.VisitExpressions(model.IdentityVisitor, func(expr model.Expression) (model.Expression, hcl.Diagnostics) {
		objectExpr, isObjectExpr := expr.(*model.ObjectConsExpression)
		if isObjectExpr {
			// objectExpr ::  model.ObjectConsExpression
                         // SOMEHOW EXTRACT schema.Type from objectExpr?
                         objectType := somehowGetType(objectExpr)
                         for _, importDef := range collectObjectImports(objectExpr, objectType) {
				imports = append(imports, importDef)
			}
		}

		return expr, nil
	})

	contract.Assert(len(diags) == 0)
}
```
But I wasn't able to resolve the schema.Type from an instance of type model.ObjectConsExpression 🤔 the current implementation gets the root resource type and traverses the types from there instead of resolving _on-demand_ from existing expressions. I will investigate this again ✅

pkg/codegen/java/gen_program.go
Zaid-Ajaj: Done

Zaid-Ajaj:
So I've been investigating the approach with `node.VisitExpression` to simplify the implementation I wrote but so far no luck. Specifically because the function `pcl.GetSchemaForType(model.Type)` isn't returning any type from the schema 🤔 I need this in order to retrieve the type for object expressions but it always returns `nil, false`

I believe it is not working because the input expressions themselves do not have any _type annotations_ which is what `pcl.GetSchemaForType` uses to map the types. I've talked to @iwahbe about this before and now that I am thinking about it, it is kinda weird that it only requires `model.Type` and nothing else (like the `pcl.Program` itself)

As for sharing naming logic, I didn't look too much into it because I was focusing on the above ☝️ so I will open an issue to track it

Zaid-Ajaj:
Unification of import naming will be tracked in #611

PR #599, opened 2022-05-18
Refactor: tests do not inherit from stack

t0yv0:
# Description

Toward making the Stack class `final` and removing completely the pattern from C# of declaring subclasses of Stack with `@Export` annotated fields. The key impediment to that is lots of tests written using this pattern. The PR refactors these tests to use `public static void init(Context ctx)` pattern of declaring stacks that is similar of what we recommend for production use now.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
@@ -10,7 +10,7 @@ import java.util.Map;
 import java.util.Objects;
 
 // TODO: use real context after refactoring test framework (and as a consequence DefaultRunner, and PulumiInternal)
-public class TestContext implements Context {
+class TestContext implements Context {
     private final ImmutableMap.Builder<String, Output<?>> exports = ImmutableMap.builder();
 
     public Map<String, Output<?>> getStackOutputs() {


sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -115,26 +114,69 @@ public CompletableFuture<TestAsyncResult> runAsync(Supplier<CompletableFuture<Ma
...
+
+        public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFactory) {
+            return this.runAsync(() -> {
+                var context = new TestContext();
+                stackFactory.accept(context);
+                return CompletableFuture.completedFuture(context.getStackOutputs());
+            });
+        }
+
+        public CompletableFuture<ImmutableList<Resource>> testAsync(Consumer<Context> stackFactory) {
```
t0yv0, line 135: Perhaps `runAsync`?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -115,26 +114,69 @@ public CompletableFuture<TestAsyncResult> runAsync(Supplier<CompletableFuture<Ma
...
+                            this.exceptions.size(), this.exceptions.stream()
+                                    .map(Throwable::getMessage)
+                                    .collect(Collectors.joining(", "))
+                    ));
+                }
             }
         }
     }
 
+    private static final class StackOutputNotFoundException extends RuntimeException {
```
t0yv0, line 174: This one is a bit weird, do we have any situation in main code (non-test) where we're trying to get an output that's not there? What do we throw there? Perhaps when querying a StackReference?

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
```diff
@@ -46,19 +47,26 @@ void testTerminatesEarlyOnException() {
...
+        public static void init(Context ctx) {
             Output.of(CompletableFuture.failedFuture(new RunException("Deliberate test error")));
-            this.slowOutput = Output.of(new CompletableFuture<Integer>()
-                    .completeOnTimeout(1, 60, TimeUnit.SECONDS));
+            ctx.export("slowOutput",Output.of(new CompletableFuture<Integer>()
+                    .completeOnTimeout(1, 60, TimeUnit.SECONDS)));
+        }
+
+        public static TerminatesEarlyOnExceptionStack of(DeploymentTests.DeploymentMock.TestAsyncResult result) {
+            return new TerminatesEarlyOnExceptionStack(result.getStackOutput("slowOutput"));
```
t0yv0, line 69: NOTE: we used to have reflection via `@Export` annotation help consolidate that the types and slots of "export" declarations align with the fields in the Stack object, which is helpful in the actual test code. I wanted to keep that typing benefit here but rolled boilerplate hand-written code to recover the types in the `.of()` methods. Not particularly pretty but gets this done.

sdk/java/pulumi/src/test/java/com/pulumi/resources/StackTest.java
```diff
@@ -1,49 +1,51 @@
...
+            ctx.export("implicitName", Output.of("buzz"));
         }
     }
 
     @Test
     void testValidStackInstantiationSucceeds() {
-        var result = run(ValidStack::new);
-        assertThat(result.t2).hasSize(3);
+        var result = run(ValidStack::init, ValidStack::new);
+        assertThat(result.t2).hasSize(2);
```
t0yv0, line 47: I was confused why hasSize(3) was there before. Sounds like only two outputs are there in the stack.

sdk/java/pulumi/src/test/java/com/pulumi/resources/StackTest.java
```diff
@@ -59,52 +61,36 @@ void testValidStackInstantiationSucceeds() {
...
-        var stack = factory.get();
-        Internal.from(stack).registerPropertyOutputs();
-
+        var result = mock.tryTestAsync(factory).join();
         //noinspection unchecked
         ArgumentCaptor<Output<Map<String, Output<?>>>> outputsCaptor = ArgumentCaptor.forClass(Output.class);
 
-        // TODO: is this OK that we're called twice?
-        verify(mock.deployment, atLeastOnce()).registerResourceOutputs(any(Resource.class), outputsCaptor.capture());
+        verify(mock.deployment, times(1))
```
t0yv0, line 89: Not called twice anymore, tighten to `times(1)`.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: looks like it is unused currently

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -115,26 +114,69 @@ public CompletableFuture<TestAsyncResult> runAsync(Supplier<CompletableFuture<Ma
...
+                                   ImmutableList<Exception> exceptions,
+                                   Map<String, Output<?>> stackOutputs) {
                 this.resources = resources;
                 this.exceptions = exceptions;
+                this.stackOutputs = stackOutputs;
+            }
+
+            public <T> Output<T> getStackOutput(String name) {
+                if (!this.stackOutputs.containsKey(name)) {
+                    return Output.of(CompletableFuture.failedFuture(new StackOutputNotFoundException(name)));
```
pawelprazak, line 157: this needs to be short lived, until the next refactor, looks like accidental complexity

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: I don't think so. We probably should throw in the StackReference, currently we return `null` from what I can see :|

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -341,4 +383,27 @@ public static Log mockLog(Logger logger) {
     public static Log mockLog(Logger logger, Supplier<Engine> engine) {
         return new Log(new DefaultEngineLogger(logger, () -> Mockito.mock(Runner.class), engine));
     }
+
+    private static final class OutputsCaptureHelper {
```
pawelprazak, line 387: what is this monstrosity?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -115,26 +114,69 @@ public CompletableFuture<TestAsyncResult> runAsync(Supplier<CompletableFuture<Ma
             }
             var mockEngine = (MockEngine) engine;
             var mockMonitor = (MockMonitor) monitor;
-            return this.runner.runAsyncFuture(callback)
+            var helper = new OutputsCaptureHelper(callback);
```
pawelprazak, line 117: this looks suspicious and not helpful contrary to the name ;)

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.pulumi.Config;
+import com.pulumi.Context;
+import com.pulumi.context.LoggingContext;
+import com.pulumi.core.Output;
+
+import java.util.Map;
+import java.util.Objects;
+
+public class TestContext implements Context {
```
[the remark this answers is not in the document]
pawelprazak: what is this class? looks like accidental complexity

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentExceptionTest.java
```diff
@@ -40,14 +41,14 @@ static void cleanup() {
...
-        assertThatThrownBy(() -> mock.testAsync(MyIncorrectStack::new).join())
+        assertThatThrownBy(() -> mock.testAsync(MyIncorrectStack::init).join())
                 .getRootCause()
                 .isInstanceOf(RunException.class)
                 .hasMessageContaining(DeliberateException.class.getName());
     }
 
-    public static class MyIncorrectStack extends Stack {
-        public MyIncorrectStack() {
+    public static class MyIncorrectStack {
```
pawelprazak, line 50: no need for a class, a method is enough `public static void myIncorrectStack(Context ctx)`

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
pawelprazak: I'd much rather we add good code, instead of even worse code then we're trying to refactor.
Getting those values from Stack instance would be a lot better IMO.

pawelprazak (commented):
I have my doubts about the added boilerplate.

I'll have a second round of review in the IDE and try to propose changes in a PR on top.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
t0yv0: How would that work?

We're inching toward a world where Stack is a final class and is only used in internally.

The only alternative here is a reflecting facility that hydrates a class out of `Map<String, Output<T>>` which we do not have in the codebase I think. I don't think adding such a facility would belong in this PR,  and if the only motivation is so that we keep the test suite, that is a weak motivation.

I think a good place to return to this is when re-designing the testing API for the users. This code here is fine for what it's doing.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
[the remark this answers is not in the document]
t0yv0: Not at all accidental, to run callbacks `void init(Context ctx)` in *test* context we need to pass something implementing the `Context` interface. It seemed to me the easiest is to implement TestContext interface, how else would we do this? Are you suggesting using ContextInternal instance here? It looked involved but maybe possible.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
t0yv0: I actually think something like this to be long-lived but perhaps renamed.  We need a clean design for mock testing API, and in this API the user should be able to easily pull out the stack exports to make assertions against. How would it look like if not using this API?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
t0yv0: It's an internal detail for keeping track of the `Map<String, Output<?>>` stack outputs coming out of the user-provided code, so that they can be exposed in the Result class. It's much easier to have it in a little class than inlining this into the call site but YMMV.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
pawelprazak: I might have a simple idea, but I need to test it first.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: This is clean, it's what I've done in my rerfactoring:
```
Stack.outputs()
Stack.output(name, type)
```

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: this looks like accidental complexity hiding bad architecture, there are better ways

t0yv0: Looks like the method that this supports has 1 call site. Let me see if this can just be deleted.

t0yv0: That worked. This is gone.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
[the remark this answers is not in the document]
pawelprazak: OK, after working with the code I understand now, it's cool idea to not have to modify too much.
When we have a full runner based on `PulumiInternal` this will not be needed, because we'll be able to supply a real one. IMO this is important for the user to be able to use testing framework almost in the same way as they'd use the normal API.
We might want to extend the `ContextInternal` with a `TestContext` in the future but I haven't found a use case for it yet.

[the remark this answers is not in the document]
t0yv0: Ok sound good, I'll put a bit of effort to removing this, but if it looks too deep, I'll just put a TODO here to remove this and link to an issue. We can at the very least easily make this a private class, not public.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: Not sure about the user, but useful internally for sure.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
pawelprazak: I've reduced the amount of those methods to 1 universal

pawelprazak:
My 2nd review: https://github.com/pulumi/pulumi-java/pull/601

pawelprazak (approved):

PR #601, opened 2022-05-18
Simplify test changes

pawelprazak:
# Description

Review for https://github.com/pulumi/pulumi-java/pull/599

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
@@ -1662,15 +1662,6 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
     @InternalUse
     @VisibleForTesting
     static class DefaultRunner implements Runner {
-        private static final int ProcessExitedSuccessfully = 0;
-        private static final int ProcessExitedBeforeLoggingUserActionableMessage = 1;
-        // Keep track if we already logged the information about an unhandled error to the user.
-        // If so, we end with a different exit code. The language host recognizes this and will not print
-        // any further messages to the user since we already took care of it.
-        //
-        // 32 was picked to be very unlikely to collide with any other error codes.
-        private static final int ProcessExitedAfterLoggingUserActionableMessage = 32;
-
         private final Logger standardLogger;
         private final EngineLogger engineLogger;
 
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
@@ -6,6 +6,7 @@ import com.google.common.collect.ImmutableSet;
 import com.pulumi.Context;
 import com.pulumi.Log;
 import com.pulumi.core.Output;
+import com.pulumi.core.internal.Internal;
 import com.pulumi.deployment.MockEngine;
 import com.pulumi.deployment.MockMonitor;
 import com.pulumi.deployment.Mocks;
@@ -17,15 +18,16 @@ import org.mockito.ArgumentMatchers;
 import org.mockito.Mockito;
 
 import javax.annotation.Nullable;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
-import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
 public class DeploymentTests {
@@ -78,14 +80,7 @@ public class DeploymentTests {
             this.config.setAllConfig(config, secretKeys);
         }
 
-        public <T extends Stack> CompletableFuture<ImmutableList<Resource>> testAsync(Supplier<T> stackFactory) {
-            return tryTestAsync(stackFactory).thenApply(r -> {
-                r.throwOnError();
-                return r.resources;
-            });
-        }
-
-        public <T extends Stack> CompletableFuture<TestAsyncResult> tryTestAsync(Supplier<T> stackFactory) {
+        public CompletableFuture<TestResult> runTestAsync(Consumer<Context> stackCallback) {
             if (!(engine instanceof MockEngine)) {
                 throw new IllegalStateException("Expected engine to be an instanceof MockEngine");
             }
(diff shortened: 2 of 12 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -78,32 +80,7 @@ public void overrideConfig(ImmutableMap<String, String> config, @Nullable Iterab
...
-                    .thenApply(stackObject -> new TestAsyncResult(
-                            ImmutableList.copyOf(mockMonitor.resources),
-                            mockEngine.getErrors().stream()
-                                    .map(RunException::new).collect(toImmutableList()),
-                            Resource.ResourceInternal.findOutputs(stackObject)
-                    ));
-        }
-
-        public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFactory) {
+        public CompletableFuture<TestResult> runTestAsync(Consumer<Context> stackCallback) {
```
pawelprazak, line 83: one universal method

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -112,63 +89,87 @@ public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFa
...
+                            ImmutableList.copyOf(mockEngine.getErrors()),
+                            outputs.apply(mockMonitor.resources)
+                    ));
         }
 
-        public static class TestAsyncResult {
-            public final ImmutableList<Resource> resources;
-            public final ImmutableList<Exception> exceptions;
+        public static class TestResult {
+            public final int exitCode;
```
pawelprazak, line 115: more info that is useful in tests

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -112,63 +89,87 @@ public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFa
...
+                this.errors = ImmutableList.copyOf(errors);
+                this.stackOutputs = ImmutableMap.copyOf(stackOutputs);
+            }
+
+            public Output<Object> stackOutput(String name) {
+                return stackOutput(name, Object.class);
             }
 
-            public <T> Output<T> getStackOutput(String name) {
+            public <T> Output<T> stackOutput(String name, Class<T> type) {
```
pawelprazak, line 138: safer casing

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -112,63 +89,87 @@ public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFa
...
+                    throw new IllegalArgumentException(String.format(
+                            "Cannot cast '%s' to the given type: '%s'",
+                            o.getClass().getTypeName(),
+                            type.getTypeName()
+                    ));
+                });
             }
 
-            public void throwOnError() {
+            public TestResult throwOnError() {
```
pawelprazak, line 160: returning this allows for more useful semantics

pawelprazak (commented):
less API surface are with the same functionality

pawelprazak:
I've added a cherry pick from main "Remove PULUMI_JAVA_TASK_TIMEOUT_IN_MILLIS" to unblock CI.
The original PR needs to be rabased and then this is no longer an issue.

The change that made the tests return real exceptions instead of synthetic ones, surfaced this issue.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentExceptionTest.java
```diff
@@ -40,22 +38,20 @@ static void cleanup() {
...
-
-        assertThatThrownBy(() -> mock.testAsync(MyIncorrectStack::init).join())
-                .getRootCause()
-                .isInstanceOf(RunException.class)
-                .hasMessageContaining(DeliberateException.class.getName());
-    }
-
-    public static class MyIncorrectStack {
-        public static void init(Context ctx) {
+        assertThatThrownBy(() -> mock.runTestAsync(ctx -> {
```
t0yv0, line 41: 👍 inlining here seems to be reasonable.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
```diff
@@ -31,43 +31,34 @@ public void cleanup() {
...
-        public static void init(Context ctx) {
-            Output.of(CompletableFuture.failedFuture(new RunException("Deliberate test error")));
-            ctx.export("slowOutput",Output.of(new CompletableFuture<Integer>()
-                    .completeOnTimeout(1, 60, TimeUnit.SECONDS)));
-        }
-
-        public static TerminatesEarlyOnExceptionStack of(DeploymentTests.DeploymentMock.TestAsyncResult result) {
-            return new TerminatesEarlyOnExceptionStack(result.getStackOutput("slowOutput"));
-        }
+        assertThat(Internal.of(result.stackOutput("slowOutput")).getDataAsync()).isNotCompleted();
```
t0yv0, line 59: This thinks that `result.stackOutput` is `Output<Object>` or `Output<?>` right? But that seems OK since the test does not care? This seems reasonable in this test.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
```diff
@@ -31,43 +31,34 @@ public void cleanup() {
...
+        assertThat(Internal.of(result.stackOutput("slowOutput")).getValueNullable()).isNotCompleted();
```
t0yv0, line 60: I think the two assertThat are testing the same thing, but make it look like something significant. Perhaps keep just one.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
```diff
@@ -54,9 +56,10 @@ void testCustomMocks() {
                 .setMocks(new MyMocks())
                 .setSpyGlobalInstance();
 
-        var resources = mock.testAsync(MyStack::init).join();
+        var result = mock.runTestAsync(MocksTest::myStack).join()
+                .throwOnError();
 
-        var instance = resources.stream()
+        var instance = result.resources.stream()
```
t0yv0, line 62: :+1:

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
```diff
@@ -112,10 +116,9 @@ void testStack() {
                 .setMocks(new MyMocks())
                 .setSpyGlobalInstance();
 
-        var result = mock.tryTestAsync(MyStack::init).join();
-        var stack = MyStack.of(result);
-        var ip = OutputTests.waitFor(stack.publicIp).getValueNullable();
-        assertThat(ip).isEqualTo("203.0.113.12");
+        var result = mock.runTestAsync(MocksTest::myStack).join();
+        var publicIp = waitForValue(result.stackOutput("publicIp", String.class));
```
t0yv0, line 120: Ok here we need to tell it it's a `String.class`, but does not seem too bad.

sdk/java/pulumi/src/test/java/com/pulumi/resources/StackTest.java
```diff
@@ -26,71 +23,38 @@
...
-                OutputTests.waitFor(result.t2.get("implicitName"))
-        );
-    }
-
-    private static class NullOutputStack {
-        public static void init(Context ctx) {
-            Output<String> foo = null;
-            ctx.export("foo", foo);
-        }
+        var result = run(ctx -> {
```
t0yv0, line 28: 👍 the old distinction makes no sense anymore, nice.

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/ResourceRefPropertyTest.java
```diff
@@ -30,13 +30,22 @@
...
+import static com.pulumi.core.OutputTests.waitForValue;
 import static com.pulumi.deployment.internal.DeploymentTests.cleanupDeploymentMocks;
 import static com.pulumi.serialization.internal.ConverterTests.deserializeFromValue;
 import static com.pulumi.serialization.internal.ConverterTests.serializeToValueAsync;
 import static org.assertj.core.api.Assertions.assertThat;
 
 class ResourceRefPropertyTest {
 
+    private static final Class<ImmutableMap<String, String>> ValuesClass =
+            TypeShape.<ImmutableMap<String, String>>builder(ImmutableMap.class)
```
t0yv0, line 43: TypeShape, really :) Interesting. Java can't do it with STDLIB alone?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentTests.java
```diff
@@ -112,63 +89,87 @@ public CompletableFuture<TestAsyncResult> tryTestAsync(Consumer<Context> stackFa
...
+                            ))
+                    ));
                 }
-                return (Output<T>) this.stackOutputs.get(name);
+                var output = this.stackOutputs.get(name);
+                return output.applyValue(o -> {
+                    if (type.isAssignableFrom(o.getClass())) {
+                        return type.cast(o);
+                    }
+                    throw new IllegalArgumentException(String.format(
```
t0yv0, line 152: I see, nice.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -1663,15 +1663,6 @@ public static DeploymentState fromEnvironment() {
     @InternalUse
     @VisibleForTesting
     static class DefaultRunner implements Runner {
-        private static final int ProcessExitedSuccessfully = 0;
```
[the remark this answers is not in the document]
t0yv0: Not a fan of exposing this in public API for long-term maintenance - I don't think user programs should care, beyond pass/fail. Unless Pulumi itself somehow special-handles 32 vs 1, in which case linking to relevant docs would be nice here.

t0yv0 (approved):
I like this for the most part, makes sense! I will miss my little `MyStack.of()` trick but in these examples there was not enough stack shape reuse and not enough fields in stacks for it to make sense. After all the user can reinvent it as needed. `MyStackResult.of(TestResult)` that "types" the results of calling `stackOutput` repeatedly. If needed. So this is great.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestContext.java
```diff
@@ -9,6 +9,7 @@
 import java.util.Map;
 import java.util.Objects;
 
+// TODO: use real context after refactoring test framework (and as a consequence DefaultRunner, and PulumiInternal)
```
t0yv0, line 12: Let me try to make this private, might just work.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
pawelprazak: yes, you are right, the second one "contains" the first one

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentRunnerTest.java
pawelprazak: yes: `public Output<Object> stackOutput(String name)`

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
pawelprazak: yes, this avoids nasty bugs, Java serializers usually use this kind of API

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/ResourceRefPropertyTest.java
pawelprazak: I've checked it does not compile.

PR #606, opened 2022-05-20
Bump default pulumi sdk version

t0yv0:
# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/providers/build-templates/build.gradle
+++ b/providers/build-templates/build.gradle
@@ -44,7 +44,7 @@ repositories {
 
 def prerelease = version.contains("+")
 
-def defaultPulumiJavaSdkVersion = "0.1.0";
+def defaultPulumiJavaSdkVersion = "0.2.1";
 
 def pulumiJavaSdkVersion = prerelease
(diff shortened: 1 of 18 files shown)


t0yv0:
Having some difficulty with flaky CI but I think we should do this as a pre-req to pubishing new providers.

pawelprazak (approved):

pawelprazak:
apropos bumping versions, we should bump grpc/guava/gson and other deps

t0yv0:
> apropos bumping versions, we should bump grpc/guava/gson and other deps

Very true, we need some kind of process/reminder to regularly do that, regular maintenance.

For this particular one, we should automate it, it is annoying to do.

pawelprazak:
> > apropos bumping versions, we should bump grpc/guava/gson and other deps
> 
> Very true, we need some kind of process/reminder to regularly do that, regular maintenance.
> 
> For this particular one, we should automate it, it is annoying to do.

looks like an option: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

PR #608, opened 2022-05-20
Stabilize CI with retries

### LINKED ISSUE #585: CI flaky due to passing artifacts via GPR
### What happened?

Build phases are chained to publish artifacts to GPR repo and consume from there to exchange artifacts. Periodically builds fail with "artifact not found". 

### Steps to reproduce

Hard to reproduce, but periodically builds fail.

### Expected Behavior

Builds are reliable.

### Actual Behavior

`gradle publish` fails for one of the artifacts with 409, for example:

```
* What went wrong:
Execution failed for task ':publishGprPublicationToGitHubPackagesRepository'.
> Failed to publish publication 'gpr' to repository 'GitHubPackages'
   > Could not PUT 'https://maven.pkg.github.com/pulumi/pulumi-java/com/pulumi/gcp/6.24.0-alpha.1653054928+813ba117/gcp-6.24.0-alpha.1653054928+813ba117.jar.md5'. Received status code 409 from server: Conflict
```

Despite the failure GPR creates the release with the given version, but the release is only partial. It has only a subset of the files expected from a normal release. Typically missing the POM file.

Subsequently, a dependent build fails to find this artifact in GPR:

```
    Execution failed for task ':app:compileJava'.
    > Could not resolve all files for configuration ':app:compileClasspath'.
       > Could not find com.pulumi:gcp:6.24.0-alpha.1653054928+813ba117.
         Searched in the following locations:
           - file:/home/runner/.m2/repository/com/pulumi/gcp/6.24.0-alpha.1653054928+813ba117/gcp-6.24.0-alpha.1653054928+813ba117.pom
           - https://maven-central.storage-download.googleapis.com/maven2/com/pulumi/gcp/6.24.0-alpha.1653054928+813ba117/gcp-6.24.0-alpha.1653054928+813ba117.pom
           - https://repo.maven.apache.org/maven2/com/pulumi/gcp/6.24.0-alpha.1653054928+813ba117/gcp-6.24.0-alpha.1653054928+813ba117.pom
           - https://maven.pkg.github.com/pulumi/pulumi-java/com/pulumi/gcp/6.24.0-alpha.1653054928+813ba117/gcp-6.24.0-alpha.1653054928+813ba117.pom
         Required by:
             project :app
```

### Versions used

_No response_

### Additional context

Possibly related: https://github.community/t/409-from-server-conflict-occurrs-when-publishing-to-github-packages/172806/16

This could be a bug in `gradle publish` or a bug in GPR.  Have not observed this happening yet with release builds going to Maven Central, although those use `gradle pubilsh` also.

We could potentially sidestep the issue entirely by propagating the build artifacts between build phases using means other than GPR repo.

### Contributing

t0yv0:
# Description

See if we can stabilize CI by retrying publish. 

Fixes #585

## Checklist

### DIFF
--- a/build/cmd/pulumi-java-builder/main.go
+++ b/build/cmd/pulumi-java-builder/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"log"
 	"os"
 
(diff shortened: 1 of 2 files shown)


pawelprazak (approved):
this could probably be written in gradel itself, but golang is very readable and is already done so don't thing there is a need to re-do it in Gradle

t0yv0:
Correct, the proper fix would probably be a patch for the gradle maven plugin. I rushed this in just because Go is familiar, going with "done is better than perfect" here, but we should definitely revisit if there's more maintenance on this. Also perhaps in the future we will not be using GPR repository at all, or at least not as an intermediary between build phases, so I'm not sure how long-lived this will be. So for now any fix that works like this one seems ok.

PR #612, opened 2022-05-23
[codegen] Output-versioned function invokes

### LINKED ISSUE #163: Port Output -versioned function invokes
## Hello!

## Issue details

Currently invoking a function accepts plain arguments and returns a CompleteableFuture. We instead want it to accept builders that overload T and `Output<T>` in the arguments, and returns  an `Output<Result>` as the result.

For an example illustrating the design consider this snippet from azure template. It is now possible to pass Output values to invokes directly:

### Design

```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {
       return StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                                                       .resourceGroupName(resourceGroupName)
                                                       .accountName(accountName)
                                                       .build())
            .applyValue(r -> r.keys().get(0).value())
            .asSecret();
    }
```

Before the change one had to manually await the Output args and project the resulting CompletableFuture to an Output.

```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {
        return Output.tuple(resourceGroupName, accountName).apply(tuple -> {
            var actualResourceGroupName = tuple.t1;
            var actualAccountName = tuple.t2;
            var invokeResult = StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                    .resourceGroupName(actualResourceGroupName)
                    .accountName(actualAccountName)
                    .build(), InvokeOptions.Empty);
            return Output.of(invokeResult)
                    .applyValue(r -> r.keys().get(0).value())
                    .asSecret();
        });
```

### Breaking Changes

When Java SDKs for providers such as azure-native update to the version of codegen implementing this feature, it will be a breaking change. To upgrade your code, you have two options:

1. rewrite the code to use the new form; note for example that StorageFunctions.listStorageAccountKeys returns an Output instead of a CompletableFuture, and ListStorageAccountKeysArgs now accepts Output values.

2. rewrite the code that uses the old form; it is now renamed with a Plain suffix. So for our example, you could use `StorageFunctions.listStorageAccountKeysPlain` that returns a CompletableFuture as before. 

### Follow up

- [x] https://github.com/pulumi/pulumi-java/issues/670
- https://github.com/pulumi/pulumi-java/issues/679

Zaid-Ajaj:
# Description

Fixes pulumi/pulumi#163

This PR is my initial attempt at getting function invokes to return `Output<ResultType>` instead of `CompletableFuture<ResultType>` 

My thought at first was to generate _syntax sugar_ around existing function invokes that return `CompletableFuture<ResultType>` and wrap them using `Output.all`,  `Output.apply` and `Output.of(CompletableFuture)` but then I saw that the SDK itself has the function that can be used immediately:
```java
Deployment.getInstance().invoke(....) // returns Output<T>
```
Questions are
 - Should both types of function invokes co-exist? if we have functions that return `Output<T>` is there any need for `CompletableFuture` alternative?
 - Am I on the right track with this implementation? 

TODO
 - Generate code from all providers
 - Remove workaround in programgen that wraps function invokes with `Output.of(....)`

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,18 +1,14 @@
 ### Improvements
 
-- Fix #547: Implement fully qualified imports for generated programs from PCL
-[#596](https://github.com/pulumi/pulumi-java/pull/596)
+- [codegen] **Breaking**
+  [#163](https://github.com/pulumi/pulumi-java/issues/163): function
+  invokes now accept `Output<T>` in their arguments and return
+  `Output<T>` instead of `CompletableFuture<T>`
+  [#612](https://github.com/pulumi/pulumi-java/pull/612).
 
-- Fix #163: function invokes now return `Output<T>` instead of `CompletableFuture<T>`
-[#612](https://github.com/pulumi/pulumi-java/pull/612)
-
-- Fix #419: Remove SDK dependency on Mockito
-
-- Support for using [jbang](https://jbang.dev)
-
-- Stack resource is now considered internal and cannot be directly instantiated by the user. 
-  The TestResult returned in tests is now an interface and no longer exposes fields.
+- [sdk] `Stack` resource is now considered internal and cannot be
+  directly instantiated by the user. Instead of inheriting from
+  `Stack`, please use the `Pulumi.run` form to define resources in
+  your stack as shown in the README.
 
 ### Bug Fixes
-
-Fix #627 - sanitize codegen sdk version input in pulumi-java-gen


pkg/codegen/java/gen.go
```diff
@@ -507,6 +507,11 @@ func (pt *plainType) genInputType(ctx *classFileContext) error {
 			false,               // outer optional
 			false,               // inputless overload
 		)
+
+		if isOutput, _ := propTypes[i].UnOutput(); !isOutput && propertiesAsOutputs {
```
pawelprazak, line 511: this code will be used in only one place and is not universal, moreover it is similar to code already in typeString

pawelprazak (commented):
IMHO this change shows architectural issues already in code that should be addressed as a separate issue.

pawelprazak:
> Should both types of function invokes co-exist? if we have functions that return Output<T> is there any need for CompletableFuture alternative?

this is a very good question, I second this

pkg/codegen/java/gen.go
```diff
@@ -1559,8 +1564,12 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
 					baseClass:             "com.pulumi.resources.InvokeArgs",
 					propertyTypeQualifier: inputsQualifier,
 					properties:            fun.Inputs.Properties,
+					args:                  true,
 				}
-				return args.genInputType(ctx)
+
+				// make sure the input properties are outputs
+				propertiesAsOutputs := true
```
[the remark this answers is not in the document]
Zaid-Ajaj: I kept like this because when reading the code outside of IDE, you don't see the intent but I guess it is better to keep these things as part of `plainType`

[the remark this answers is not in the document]
pawelprazak: this is the problem with flags :)

pkg/codegen/java/gen.go
```diff
@@ -490,7 +490,7 @@ func (pt *plainType) genInputProperty(ctx *classFileContext, prop *schema.Proper
 	return nil
 }
 
-func (pt *plainType) genInputType(ctx *classFileContext) error {
+func (pt *plainType) genInputType(ctx *classFileContext, propertiesAsOutputs bool) error {
```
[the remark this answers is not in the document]
pawelprazak: Probably, maybe. Ideally it would be localized to where it is actually used, but that is probably a bigger refactoring and there's no point in making it part of this PR.

The least damage we can do to an already difficult to maintain codebase is probably our best bet at this moment.

I'm sorry I don't have a more constructive comment, maybe others will have a better idea.

providers/pulumi-aws/sdk/java/src/main/java/com/pulumi/aws/acm/AcmFunctions.java
```diff
@@ -97,7 +97,7 @@ public static CompletableFuture<GetCertificateResult> getCertificate(GetCertific
      * ```
      * 
      */
-    public static CompletableFuture<GetCertificateResult> getCertificate(GetCertificateArgs args, InvokeOptions options) {
-        return Deployment.getInstance().invokeAsync("aws:acm/getCertificate:getCertificate", TypeShape.of(GetCertificateResult.class), args, Utilities.withVersion(options));
+    public static Output<GetCertificateResult> getCertificate(GetCertificateArgs args, InvokeOptions options) {
+        return Deployment.getInstance().invoke("aws:acm/getCertificate:getCertificate", TypeShape.of(GetCertificateResult.class), args, Utilities.withVersion(options));
```
pawelprazak, line 101: yes, good find! I've added this method for this purpose

pkg/codegen/java/gen.go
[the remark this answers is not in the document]
pawelprazak: I've looked at the code again and I don't have an idea for a constructive comment with a specific action,
so feel free to ignore me. Sorry about that.

pkg/codegen/java/gen.go
pawelprazak: as noted above, since I don't have an idea how to address my concern, feel free to ignore this comment, sorry about that

pawelprazak:
A kind reminder, to clear the codegen test errors we currently need to manually commit the changes to the generated test code.

Running `PULUMI_ACCEPT=true; make codegen_tests` and committing the result should clear the current CI error IIRC.

Sorry for the inconvenience, we probably should automate this somehow in the future.

Zaid-Ajaj:
Hi @pawelprazak, been trying to debug what is going on here to no avail 😓 I ran the command you sent but the tests keep failing. It is showing code diffs telling that the generated code has changed ... well, that's what the PR is supposed to do 😅 any leads?

pawelprazak:
@Zaid-Ajaj I'm sorry you're having troubles with this, if you'd like we could pair on this over zoom. 
Feel free to reach out to me on Slack.

Zaid-Ajaj:
@pawelprazak thank you! Tomorrow I will have another look and reach out if I am still stuck 🙏

Zaid-Ajaj:
NOTE to self:
 - Remember to `PUMULI_ACCEPT=true make codegen_test` 
 - Having an outdated build of the SDK in your local maven build will cause the tests to fail unless you `make install_sdk` first before running the tests

Zaid-Ajaj:
Alright so this is really frustrating to work with, after having changed how function invokes are implemented, regenerated code from providers _and_ installed them locally, the test examples failed in the CI. For example
```
make test_example.gcp-java-gke-hello-world
```
would fail in the CI but if I run it _locally_ it tells me that everything is good while using a older version of the provider (gcp in this case) 

This means I was in the unfortunate state of ✅ locally and 🔴 on CI and had to rewrite the function invokes in the test examples by hand to make them work in CI but locally it was red squiggles all over the place.

Somehow even when I install a provider after regenerating it, the test examples were still using a cached build somewhere on my machine. I would expect that `make test_example.gcp-java-gke-hello-world` would take care of removing cached builds of providers. 

The test examples are now all green which is good but still need to update `azure-java` template. Specifically, the following 
```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {
        return Output.tuple(resourceGroupName, accountName).apply(tuple -> {
            var actualResourceGroupName = tuple.t1;
            var actualAccountName = tuple.t2;
            var invokeResult = StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                    .resourceGroupName(actualResourceGroupName)
                    .accountName(actualAccountName)
                    .build(), InvokeOptions.Empty);
            return Output.of(invokeResult)
                    .applyValue(r -> r.keys().get(0).value())
                    .asSecret();
        });
    }
```
should change into:
```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {

            var invokeResult = StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                    .resourceGroupName(resourceGroupName)
                    .accountName(accountName)
                    .build());

            return invokeResult
                    .applyValue(result -> result.keys().get(0).value())
                    .asSecret();
    }
```
I guess I should update that in the templates repo and then update the templates submodule from here

Zaid-Ajaj:
NOTE: function invokes here also suffer from [pulumi/pulumi#9593](https://github.com/pulumi/pulumi/issues/9593) which is why wrapping with `Output.tuple(...).apply(...)` is still necessary sometimes when doing a `pulumi preview`

Zaid-Ajaj:
TODOs
 - [x] Generate code for all providers
 - [x] Refactor program-gen to use output-versioned invokes
 - [ ] Disable testing templates because these are out of date with latest providers
 - [ ] Bring back function invokes that return `CompletableFuture<T>` as there is no (easy) way to obtain `T` from `Output<T>` and use it in control flow unless from within `apply` and `applyValue` (thanks @mikhailshilkov for the note)

t0yv0:
> Should both types of function invokes co-exist? if we have functions that return Output<T> is there any need for CompletableFuture alternative?

Both must coexist, unless we opt to expose `Output.join() : Output<T> -> T`. There are scenarios where invokes need to run in plain context and lifting them to work on Output modality is too much pain on the user.

t0yv0:
> My thought at first was to generate syntax sugar around existing function invokes that return CompletableFuture<ResultType> and wrap them using Output.all, Output.apply and Output.of(CompletableFuture) but then I saw that the SDK itself has the function that can be used immediately:

There's some leeway here, I think in most of the SDKs we went with syntactic sugar implementation, but there was some blocker making it too difficult/impossible in the C# SDK case.

Ah I remember what it was exactly. FPlainArgs and FInputArgs types differ in deeply nested Outputs. There was no way in the C# SDK to convert `FInputArgs` to `Output<FPlainArgs` that would be required by the "syntactic sugar". And implementing the transformation at code-generation time was distasteful to me (mildly complex).

Java implementation started the port from C# so some of the same considerations can apply here.

t0yv0:
Yes, we need an update in templates repo unfortunately. Yes, the Makefile does not understand dependencies correctly which is quite frustrating. We could try working this out that'd help everyone. Providers team is experimenting and adopting with Task util instead of make also if that helps https://github.com/go-task/task

One wrench here is that the provider team is taking over the provider code trees and moving them out of the repo, so the capability to "test my changes on provider X" will have to be able to work around the source being somewhere else. I'm open to input how to make this all better, it clearly is a time sink right now.

t0yv0:
MOre review comments in https://github.com/pulumi/pulumi-java/pull/636 - since this current PR breaks GH UI

pkg/codegen/java/gen.go
```diff
@@ -1604,31 +1596,29 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
 		// Emit the args and result types, if any.
 		if fun.Inputs != nil {
 			// Generate "{Function}Args" class for invokes that return Output<T>
+			// Notice here using fun.Inputs.InputShape.Properties which use the shape which accepting Outputs
```
t0yv0, line 1615: 👍 

There's a bit of misnomer here. In other languages there is an `Input<T>` type, canonically (in TypeScript) it's an untagged union `T | Output<T>`. In Java we made the controversial decision to not model that type at all, so we only have `Output<T>` - but the cross-language Go framework is still calling that `InputShape` - which is fine. at the meta-level `InputShape` will inject `InputType` wrappers, that will be translated to `Output<T>` during the Java specific codegen pass.

pkg/codegen/java/gen_test.go
```diff
@@ -35,6 +35,10 @@ var javaSpecificTests = []*test.SDKTest{
 		Directory:   "mini-awsx",
 		Description: "Regression tests extracted from trying to codegen awsx",
 	},
+	{
```
t0yv0, line 38: Hmmm, this one is a bit surprising to me. My mental model was that we borrow tests from the common test suite and then add Java-specific regression tests here. Like `mini-awsnative` is Java specific test. But it would seem that `output-funcs-edgeorder` should be coming from pulumi/pulumi? It's not there?

t0yv0: That is I was expecting it to be in test.PulumiPulumiSDKTests?

t0yv0: I'll take care of sorting this out later, not blocking.

t0yv0:
Actually @Zaid-Ajaj I have tried testing the azure-template with modifications, and it's not working. Unfortunately that's blocking. We have a bug in here somewhere.

Here is the modification I've made to take advantage of the new features:

```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {
        return StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                                                       .resourceGroupName(resourceGroupName)
                                                       .accountName(accountName)
                                                       .build())
            .applyValue(r -> r.keys().get(0).value())
            .asSecret();
    }
```

``` modified:   azure-java/src/main/java/myproject/App.java```

Here's how I tested the change:

```
make install_sdk
make provider.azure-native.install
make test_template.azure-java

```

```
    io.grpc.StatusRuntimeException: UNKNOWN: invocation of azure-native:storage:listStorageAccountKeys returned an error: request failed /subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/{resourceGroupName}/providers/Microsoft.Storage/storageAccounts/{accountName}/listKeys: autorest/azure: Service returned an error. Status=404 Code="ResourceGroupNotFound" Message="Resource group '{resourceGroupName}' could not be found."
```

**Comment by t0yv0** [comment]
> I guess I should update that in the templates repo and then update the templates submodule from here

Here's a reason we can't do that - updating the templates repo would BREAK the templates for all Pulumi users since these changes are not released yet. Hence my suggestion of having a branch in that repo tracking `pulumi-java/main` that we sync to master whenever we release.

**Comment by Zaid-Ajaj** [comment]
> Actually @Zaid-Ajaj I have tried testing the azure-template with modifications, and it's not working. Unfortunately that's blocking. We have a bug in here somewhere.

@t0yv0 I've actually seen this error before, it is a manifestation of [pulumi/pulumi#7030](https://github.com/pulumi/pulumi/issues/7030) where function invokes fail when populating their parameters with instances `Output<T>` during `pulumi preview` 

[pulumi/pulumi#9593](https://github.com/pulumi/pulumi/issues/9593) Also demontrates the same behaviour

The suggested workaround during a `pulumi preview` seems to be guarding the function invoke with an `unknownOutput.apply(unused -> invokeHere())` where `unknownOutput` can be any ID of any resource since these unknown during a preview. 

Another workaround in the template code is to guard the function invoke with an `if` statement that checks if we are executing a dry run
```java
    private static Output<String> getStorageAccountPrimaryKey(Output<String> resourceGroupName,
                                                              Output<String> accountName) {

        if (Deployment.getInstance().isDryRun())
        {
             return Output.secret("*** StorageAccountPrimaryKey ***");
        } 

        return StorageFunctions.listStorageAccountKeys(ListStorageAccountKeysArgs.builder()
                                                       .resourceGroupName(resourceGroupName)
                                                       .accountName(accountName)
                                                       .build())
            .applyValue(r -> r.keys().get(0).value())
            .asSecret();
    }
```
The final workaround at runtime is to make function invokes smarter and expanding their logic as follows:
```java
public static Output<ArbitFuncResult> arbitFunc(ArbitFuncArgs args)
{
    if (Deployment.getInstance().isDryRun())
    {
         // return Output<ArbitFuncResult> but it is unknown?
    }

    // execute invoke as usual... 
}
```
I think making the generated function invokes smarter is the way to go so that users don't have to worry changing their code to adapt to preview/update runs. What do you think?

t0yv0:
My preferred fix for *this* issue pulumi/pulumi-java/pull/671 which is also what we have in C#, Python, etc.

The suggestion to not invoke in preview is interesting but I can see some downsides, well, namely that invokes don't work in preview. Bookmarking it for further discussion. This would be a change across all SDKs not just Java.

I also would like to step through the ref'd issues myself to get a better understanding:

- https://github.com/pulumi/pulumi/issues/7030
- https://github.com/pulumi/pulumi/issues/9593

t0yv0:
Confirming that I can pass the test with #671  merged here.

```
make install_sdk
make provider.azure-native.install
export PULUMI_JAVA_SDK_VERSION=0.0.1 # ouch, otherwise template test does not use the in-source SDK 
make test_template.azure-java
```

tests/examples/aws-java-webserver/app/src/main/java/webserver/App.java
```diff
@@ -23,7 +23,7 @@ public static void main(String[] args) {
 
     public static void stack(Context ctx) {
         final var amiResult = Ec2Functions.getAmi(GetAmiArgs.builder()
-                .filters(GetAmiFilter.builder()
+                .filters(GetAmiFilterArgs.builder()
```
t0yv0, line 26: Any idea why is this changing? This feels like a breaking change that was not called for.

t0yv0: Figure out - this is because GetAmiArgs is now InputShape, so all the nested types get named Args. Perfect.

t0yv0 (approved):

PR #624, opened 2022-05-23
[codegen] Use PackageReferences in program gen.

pgavlin:
Change the program generator to use schema.PackageReference values rather
than *schema.Package values. The former load types, resources, and
functions as they are needed rather than up-front, which can drastically
reduce the peak memory consumption of the executor.


pkg/codegen/java/gen_program.go
```diff
@@ -146,7 +146,11 @@ func GenerateProgram(program *pcl.Program) (map[string][]byte, hcl.Diagnostics,
 
 	// Import Java-specific schema info.
 	// FIX ME: not sure what this is doing...
-	for _, p := range program.Packages() {
+	packages, err := program.PackageSnapshots()
```
t0yv0, line 149: Since we're here, remove the FIX ME above with a brief comment on what it's doing? It looks like we need it somehow but I'm confused. I tried looking at the source behind ImportLanguages. Can't quite understand what's going on here. 

Sounds like the Java Importer implementation is mostly a no-op pass-through.  Do some languages hook into this interface to do something custom?

// Importer implements schema.Language for Java


CC @Zaid-Ajaj

pkg/codegen/java/gen_program.go
```diff
@@ -568,12 +572,15 @@ func makeResourceName(baseName string, suffix string) string {
...
 func (g *generator) findFunctionSchema(function string) (*schema.Function, bool) {
-	for _, pkg := range g.program.Packages() {
-		if pkg.Functions != nil {
-			for _, functionSchema := range pkg.Functions {
-				if functionSchema != nil && strings.HasSuffix(functionSchema.Token, names.LowerCamelCase(function)) {
-					return functionSchema, true
+	function = names.LowerCamelCase(function)
+	for _, pkg := range g.program.PackageReferences() {
+		for it := pkg.Functions().Range(); it.Next(); {
+			if strings.HasSuffix(it.Token(), function) {
```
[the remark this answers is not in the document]
Zaid-Ajaj: I believe because `it.Token()` contains the `pkg:module:functionName` so here the check determines whether `function` is `functionName` (from the token)

pkg/codegen/java/gen_program.go
Zaid-Ajaj: I think the language importer is supposed to load java-specific schema information such as property name overrides but I haven't seen a schema that actually makes use of it (and programgen doesn't either)

Zaid-Ajaj:
@t0yv0 I'll have a look tomorrow 👍

Zaid-Ajaj (approved):
> Might just need to skip enum-reference-python codegen test.

@t0yv0 Indeed, that was the issue. I made sure the test is skipped and rebased to master. Now all is good

PR #632, opened 2022-05-27
Better CI for PRs from forked repos

t0yv0:
# Description

Fixes # (issue)

## Checklist


t0yv0:
CC @dixler - I will check in for now as I don't think I can fully test till it's on main; but the idea here is to copy the workflow with accepting tests as we have on pulumi/pulumi. Let me know if you have any review comments I'll fixup in a follow up.

Issue #548, opened 2022-05-04
Post-launch cleanup


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

- [x] migrate java examples in https://github.com/pulumi/examples from Gradle to Maven
     - [x] https://github.com/pulumi/examples/pull/1228
- [x] remove examples that migrated to https://github.com/pulumi/examples and to compensate example test coverage, find a way to test latest provider/sdk changes on top of example code from pulumi/examples
  - [x] https://github.com/pulumi/pulumi-java/pull/672 
- [ ] https://github.com/pulumi/pulumi-java/issues/685 separately, add java examples to CI in https://github.com/pulumi/examples - this would test latest release versions
- [x] https://github.com/pulumi/pulumi-java/pull/574
  - [x] remove templates that migrated to https://github.com/pulumi/templates
  - [x] to compensate template test coverage, find a way to test latest provider/sdk changes on top of templates from pulumi/templates
- [x] https://github.com/pulumi/templates/pull/287 separately, add templates to CI in https://github.com/pulumi/templates - this would test against latest released versions, and find things like https://github.com/pulumi/templates/pull/276 - we had passing CI but shipped an essentially broken template
  - [x] debug windows hanging

t0yv0:
CC @danielrbradley had some thoughts on example organization - happy to course correct.

CC @AaronFriel had objections to submodules, what would be a workaround? Could maintain a copy of examples in-repo with a script to re-fetch/re-copy them from examples/examples?

t0yv0:
After a quick discussion with Daniel and Paul I think the understanding is that while the future of examples ownership is still being decided, it should be OK for me to proceed to drop the example source code from pulumi-java and use the source code from `pulumi/examples` for those examples that are duplicated. 

We can play it by ear down the line which examples live where.

I'm hitching up automation to make sure the SDK (and currently provider) changes get tested by those examples as we currently need that coverage. So the automation can do examples from either this repo or pulumi/examples.

git modules currently are used to share source, but it's an implementation detail, per @AaronFriel we can drop that later.

t0yv0:
Closing this out leaving https://github.com/pulumi/pulumi-java/issues/685 for later.

PR #634, opened 2022-05-27
Bump versions

### LINKED ISSUE #623: Bump SDK dependencies versions
## Hello!

## Issue details

Update SDK dependencies.
Consider automating with dependabot.

pawelprazak:
# Description

Fixes #623 

## Checklist

### DIFF
--- a/sdk/java/pulumi/build.gradle
+++ b/sdk/java/pulumi/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id("signing")
     id("java-library")
     id("maven-publish")
-    id("com.google.protobuf") version "0.8.16"
+    id("com.google.protobuf") version "0.8.18"
 }
 
 version = System.getenv("PULUMI_JAVA_SDK_VERSION") ?: "0.0.1"
@@ -22,8 +22,8 @@ repositories {
     mavenLocal()
 }
 
-def grpcVersion = "1.38.0"
-def protobufVersion = "3.12.0"
+def grpcVersion = "1.46.0"
+def protobufVersion = "3.19.2"
 def protocVersion = protobufVersion
 
 dependencies {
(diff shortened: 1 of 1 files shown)


CHANGELOG_PENDING.md
```diff
@@ -9,4 +9,5 @@
 
 ### Bug Fixes
 
-Fix #627 - sanitize codegen sdk version input in pulumi-java-gen
+Fix #627: sanitize codegen sdk version input in pulumi-java-gen
+Fix #623: update SDK versions, update dependencies versions
```
t0yv0, line 13: This kind of change may be omitted from the changelog.

tests/examples/aws-java-webserver/app/build.gradle
```diff
@@ -25,7 +25,7 @@ repositories {
     }
 }
 
-var pulumiJavaSdkVersion = System.getenv("PULUMI_JAVA_SDK_VERSION") ?: "0.0.1"
+var pulumiJavaSdkVersion = System.getenv("PULUMI_JAVA_SDK_VERSION") ?: "0.2.1"
```
t0yv0, line 28: This actually may break something intentional. We use 0.0.1 to mark the "Current version in repo" so that `make install_sdk` works and we can reference that from examples as the default; like with this change `make test_example.xyz` will go and test against the released 0.2.1 version not your changes.

I wonder how to organize this better.

tests/examples/minimal/pom.xml
```diff
@@ -15,21 +15,23 @@
         <maven.compiler.release>11</maven.compiler.release>
         <mainClass>${project.groupId}.${project.artifactId}.App</mainClass>
         <mainArgs/>
+        <pulumiSdkVersion>0.2.1</pulumiSdkVersion>
+        <pulumiRandomVersion>4.6.0</pulumiRandomVersion>
```
t0yv0, line 19: Does this need pulumiRandomVersion?

tests/examples/minimal/pom.xml
```diff
@@ -39,7 +41,7 @@
         <dependency>
             <groupId>com.pulumi</groupId>
             <artifactId>pulumi</artifactId>
-            <version>${env.PULUMI_JAVA_SDK_VERSION}</version>
+            <version>${pulumiSdkVersion}</version>
```
t0yv0, line 44: I'm a bit scared messing with these

t0yv0:
Is there a programmatic way to upgrade a dependency in a POM file? Like a CLI interface? So I could say "mvn something update-pulumi-sdk-dep 1.2.3"? If we could do that we could simplify some strange regex code and move out the dependency shenanigans out of the example code and into the testing harness, which would be better.

In the real examples we want people to use we actually use floating references at the moment:
https://github.com/pulumi/examples/blob/master/aws-java-webserver/app/build.gradle#L14

I was going to work on deleting the examples from here anyway as part of https://github.com/pulumi/pulumi-java/issues/548 but I'm waiting a bit on the team to make a decision where the examples should go.

tests/examples/aws-java-webserver/app/build.gradle
pawelprazak: hmm, that's unfortunate because it breaks testing locally. IMO current version should be always generated from pulumicli (hash + dirty in case of active debug, maybe even some kind of build number that will change on very compile) so things are consistent and does not require intimate knowledge of our build system

tests/examples/minimal/pom.xml
pawelprazak: why, this is very easy, this is the canonical way of doing this version hack

pawelprazak:
> Is there a programmatic way to upgrade a dependency in a POM file? Like a CLI interface? So I could say "mvn something update-pulumi-sdk-dep 1.2.3"? If we could do that we could simplify some strange regex code and move out the dependency shenanigans out of the example code and into the testing harness, which would be better.
> 

I'm not aware of a mechanism that would do that, but maybe there is one.

tests/examples/minimal/pom.xml
pawelprazak: sorry, my bad

tests/examples/aws-java-webserver/app/build.gradle
pawelprazak: I'll remove this change then. I thought it was an omission.

pawelprazak:
> > Is there a programmatic way to upgrade a dependency in a POM file? Like a CLI interface? So I could say "mvn something update-pulumi-sdk-dep 1.2.3"? If we could do that we could simplify some strange regex code and move out the dependency shenanigans out of the example code and into the testing harness, which would be better.
> 
> I'm not aware of a mechanism that would do that, but maybe there is one.

Funny thing, actually I've used to know this, but forgot. Revising my old project reminded me about some maven magic:
https://github.com/pawelprazak/java-extended/blob/master/update_version.sh#L31

t0yv0 (approved):
Thanks this one is easy approve!

PR #636, opened 2022-05-31
output versioned function invokes

t0yv0:
# Description

@Zaid-Ajaj opening this up to help us use GitHub UI for review; the PR you had checked in providers/ changes which breaks the ability to use GitHub UI. We have some scripts to help check in that with "one provider per commit" and we use a stacked PR for now to workaround. Anyhow, I'll use this PR for now to comment. It's identical to yours just without providers/ and `pr.yml` change (might need to add that, I see why we need it).

Fixes # (issue)

## Checklist


pkg/codegen/java/gen.go
```diff
@@ -1515,50 +1534,101 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
...
 		if fun.Inputs != nil {
+			// Generate "{Function}Args" class for invokes that return Output<T>
 			if err := addClass(inputsPkg, argsClass, func(ctx *classFileContext) error {
 				args := &plainType{
 					mod:                   mod,
 					name:                  ctx.className.String(),
 					baseClass:             "com.pulumi.resources.InvokeArgs",
 					propertyTypeQualifier: inputsQualifier,
 					properties:            fun.Inputs.Properties,
+					propertiesAsOutputs:   true,
```
t0yv0, line 1614: This doesn't really work I'm afraid think for nested types.. the proeprties

```
f(FArgs)
class FArgs {
  A Output<AType>;
}
class AType {
  B Output<Integer>; // want this to vary also, `Integer` vs `Output<Integer>`
}
```

t0yv0: I am wondering if the InputShape facility can just work for us here. Like this:

https://github.com/pulumi/pulumi/blob/ecf65dab2055f1c187d03a2d90aa200167785f76/pkg/codegen/dotnet/gen.go#L1474

You use `fun.Inputs.InputShape` and that automagically does the underlying conversion of properties to Output-wrapped properties deeply - that's my hope at least. Could we explore using that and removing `propertiesAsOutputs:   true,` and manual attempts to wrap in `Output` here?

pkg/codegen/java/gen.go
```diff
@@ -1515,50 +1534,101 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
 			returnType = ctx.imports.Ref(names.Void)
 		}
 
+		// default method name returns Output<ReturnType>
 		methodName := tokenToFunctionName(fun.Token)
 
+		// another "plain" overload will return CompletableFuture<ReturnType>
+		plainMethodName := methodName + "Plain"
```
t0yv0, line 1541: Could we use the same method name (no suffix)? In theory the overload resolution would let us do that right? Wondering if that's a good idea but want to at least consider explicitly.

pkg/codegen/java/gen.go
```diff
@@ -1498,14 +1506,25 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
...
 		inputsPkg := javaPkg.Dot(names.Ident("inputs"))
+
+		// Generating "{Function}Args" class for invokes that return Output<T>
+		// builders for this class allow outputs as inputs for their properties
 		argsClass := names.Ident(tokenToName(fun.Token) + "Args")
 		argsFQN := inputsPkg.Dot(argsClass)
 
+		// Generating "{Function}PlainArgs" class for invokes that return CompletableFuture<T>
+		// builders for this class only allow non-outputs as inputs for their properties
+		plainArgsClass := names.Ident(tokenToName(fun.Token) + "PlainArgs")
```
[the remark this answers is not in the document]
t0yv0: Are there any other ways to do it that aren't breaking? Just curious. I can't think of any that's clean, want the `"FArgs"` to stand for input-ty args.

pkg/codegen/java/gen.go
```diff
@@ -1515,50 +1534,101 @@ func (mod *modContext) genFunctions(ctx *classFileContext, addClass addClassMeth
...
+			fprintf(w, "    public static %s<%s> %s() {\n",
+				ctx.ref(names.CompletableFuture), returnType, plainMethodName)
+			fprintf(w,
+				"        return %s(%s.Empty, %s.Empty);\n",
+				plainMethodName, plainArgsType, invokeOptions)
+			fprintf(w, "    }\n")
 		}
 
-		// Add args only invoke
+		// Output version: add args only invoke
```
t0yv0, line 1567: Maybe gate by `NeedsOutputVersion` like in all other SDKs? 

https://github.com/pulumi/pulumi/blob/ecf65dab2055f1c187d03a2d90aa200167785f76/pkg/codegen/dotnet/gen.go#L1465

pkg/codegen/java/gen.go
Zaid-Ajaj: In theory you could but it will not _always_ work because if you have a function without arguments, then overloads won't compile since only the return type is different:
```java
// this won't work
Output<Result> funcName();
CompletableFuture<Result> funcName();
```

Zaid-Ajaj: > Right. In all the other SDKs we don't generate the Output-versioned invoke if there are no arguments

> I recall what made me not worry about this case too much is that there's barely any examples of such invokes across our provider base, though they are theoretically possible ofc.

Actually I thought having both versions is much more uniform:
 - It is clear which functions return what family of types
 - It makes it easier for program-gen to assume that there is always an output-returning function invoke regardless of whether they have arguments or not, so that programgen doesn't have to wrap those with `Output.of`

Zaid-Ajaj:
@t0yv0 I believe this bullet point can be done in a separate task:

> Hand-port runtime/mock tests for output-funcs. This may sound tedious but doing it mechanically for all languages has been worth the time for me in the past, every attempt found new bugs. https://github.com/pulumi/pulumi/tree/master/pkg/codegen/testing/test/testdata/output-funcs/dotnet-extras/Tests

Other than that, I will enable the tests and see if anything breaks + adding nested properties that are objects (which should be wrapped in output as well) 

Will reach out if I get stuck

t0yv0:
Going back to that PR.

PR #637, opened 2022-05-31
Fix 568: name collision preventing OCI provider gen

### LINKED ISSUE #568: Codegen name collision prevents clean oci provider codegen
### What happened?

Workaround applied was manually removing "oci:ApiGateway/getApiValidation:getApiValidation" from schema.json.

Using a vanilla schema results in a name collision preventing the generated code from compiling.

### Steps to reproduce

`make provider.oci.build` after using the standard schema.

### Expected Behavior

OCI provider builds.

### Actual Behavior

OCI provider build breaks.

### Versions used

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
# Description

There are a few ways forward here, this fix is the most conservative and surgical that just fixes 568. It's a partial name disambiguation that renames FResult to FInvokeResult when it suspects a conflict. 

Other SDK generators notably Go contain more sophisticated machinery for disambiguating name conflicts. We could upgrade to something similar. 

The trouble with sophisticated auto-naming is that upgrades that need bw-compat may get interesting and unexpected diffs from auto-naming, depending how often that happens it can be quite annoying.

Or else we could take a breaking change and possibly stash away function output types as inner classes or some such - it really should be a rare situation where a user needs to import the class for a function result directly (perhaps some sort of testing situation), more often than not this class is just not imported but used implicitly.

Curious what we think about this approach, I think what goes for it is that blast radius is fairly small, it's rarely that this code runs.

Fixes #568 

## Checklist

### DIFF
--- a/providers/pulumi-oci/pulumi-java-gen.yaml
+++ b/providers/pulumi-oci/pulumi-java-gen.yaml
@@ -1,8 +1,6 @@
 artifactID: oci
 version: 0.1.0
 schema: https://raw.githubusercontent.com/pulumi/pulumi-oci/v0.1.0/provider/cmd/pulumi-resource-oci/schema.json
-# Schema without "oci:ApiGateway/getApiValidation:getApiValidation":
-# schema: ./providers/pulumi-oci/schema.json
 out: sdk/java
 versionFile: src/main/resources/com/pulumi/oci/version.txt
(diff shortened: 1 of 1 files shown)


pawelprazak (approved):
I appreciate the surgical nature of this fix.

PR #639, opened 2022-05-31
Fix 420

### LINKED ISSUE #420: Codegen ignores Secret flag on properties with prompt types
## Hello!

## Issue details

Currently in providers such as AWS we have properties that are specified as secret, but are projected by codegen into promptly typed (not `Output<T>`-wrapped) accessors. As a result there is no guarantee that consuming code understands they are secret. We should fix this, including a breaking change to project these properties as `Output<T>`-wrapped properties so we can set the secret flag on them.

Since https://github.com/pulumi/pulumi-jvm/pull/414 this is a codegen warning but we can tighten to an error and fix for practical cases.

t0yv0:
# Description

This is about `WARN: secret property %s (%s) does not have an Output type` warnings.

Reviewing the warning I found that we were missing the proper if condition so we had a lot of spurious warnings. This PR reduces those and improves the location of the message by printing which type it is from. 

Real warnings are only from azure-native. Checking how C# and TypeScript do this, it's a problem there also. I will open an issue in pulumi/pulumi detailing this as an overall issue. Perhaps this is okay or perhaps we need to warn about schemas with a non-Output type for secret properties.

Fixes #420 

## Checklist

### DIFF
--- a/pkg/codegen/java/gen_defaults.go
+++ b/pkg/codegen/java/gen_defaults.go
@@ -57,7 +57,9 @@ func (dg *defaultsGen) defaultValueExpr(
 //
 // arg if non-empty is an expr of type targetType pointing to a
 // user-provided nullable argument.
-
+//
+// propContext is free text only used in WARN statements to
+// contextualize the property.
 func (dg *defaultsGen) builderExpr(
 	propContext string,
 	prop *schema.Property,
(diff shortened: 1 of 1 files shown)


pkg/codegen/java/gen_defaults.go
```diff
@@ -73,10 +79,10 @@ func (dg *defaultsGen) builderExpr(
...
 			builderTransformCode = ".output()"
 		}
-	} else {
+	} else if prop.Secret {
 		// TODO this should be handled or an error
-		fmt.Printf("WARN: secret propety does not have an Output type: %v\n",
-			prop.Name)
+		fmt.Printf("WARN: secret propety %s (%s) does not have an Output type\n",
+			prop.Name, propContext)
 	}
```
pawelprazak, line 86: Is it safe to ignore the `else` here? The code before changes warned about all unrecognized cases.

If it is safe to ignore, maybe adding an explicit ignore would be a nice improvement for readability:
```
} else {
  // ... description of why we ignore this branch ...
}
```

Also we might consider throwing an error instead. 

I'm not sure what is the best course of action here, but just wanted to point out that after the change we silently ignore more then before the change.

pawelprazak (approved):
Looks good, thank you! 
Had just one question.

pkg/codegen/java/gen_defaults.go
t0yv0: I think so. 

All the other warnings were false warnings. If a Config property for example has type String but is not marked secret (very common), this is all good, we should not warn. 

I like the the idea of an no-op else, let me try that. 

We cannot throw on this since this would cost us the pulumi-azure-native provider. It needs this until there is a deeper fix. So it has to be a warning for now.

t0yv0: >  after the change we silently ignore more then before the change.

Intentionally, yes! Those were spurious warnings. It was warning on AWS for example where all is well.

Issue #626, opened 2022-05-24
Refactor internals for easier testing and less complexity


pawelprazak:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

**This issue by nature probably will not have a complete description since we might learn new things while implementing and I might have forgotten some more things that should be added.**

I'll do my best to keep this list up to date as we learn more.

The goals:
- allow for **adding features and writing tests easily**
- reduce internal complexity for less costly maintenance
- allow for maximum reuse and coherence between CLI, Automation and Tests

Non-goals:
- we do not want to diverge too much from TS and Go and other implementations
- we **do not want any big changes to the public API**

Problems and solutions are described below.

Focused mainly on Deployment:
- `DeploymentImpl` reads environment variables directly instead being configured using the constructor, this creates problems with testing and violates single responsibility principle - proposed solution is to read the environment variables in `PulumiInternal` and pass an "options object" (`RuntimeContext`?) with the result along with other dependencies to the constructor (this will also help with implementing the Automation API)
- DeploymentImpl creates its own dependencies internally and that does not allow for easy testing - proposed solution is to use manual dependency injection and pass all dependencies as parameters to the constructor, this allows for cleaner `PulumiInternal` and for a cleaner testing framework with less mocking
- `DeploymentImpl` holds its dependencies as nested classes, that makes the file big and hard to work with, encourages tight coupling between separate parts, does not allow for easy use in testing framework - proposed solution is to move all nested internal classes to package `com.pulumi.runtime.internal` and also extract interfaces for all of those classes (e.g Call, Invoke, etc.), as an additional benefit this will allow for easier understanding of the underlying interaction with the engine
- `Deployment` is being accessed through a singleton in many places, deep in the code structure, this makes testing harder requiring deployment instance in places where a unit test would be more appropriate - proposed solution is to reduce the amount of calls to the singleton by passing the instance from the edges of the code
- `Deployment` as a class name suggests that it is a thing, when in reality it is a state associated with an action, this is misleading - proposed solution is a rename to `DeploymentContext` and a move to package `com.pulumi.context.internal`
- Stack can probably be passed as a `DeploymentImpl` constructor parameter instead of through a setter, resulting in simplicity and immutability
[...]

Issue #553, opened 2022-05-05
Add mock (unit) testing Java example


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

This would illustrate how to do https://www.pulumi.com/docs/guides/testing/ for Java.  See the Unit Testing section that mentions examples in other languages. We would first add an examples and then add Java reference to tag page. Note that APIs used for "unit" testing Pulumi programs need to become stable, so we might need to review these APIs before taking a further commitment.

@pawelprazak mentioned that mockito dependency is unfortunate, perhaps that can be avoided.

pawelprazak:
Note to self: fix #419 after this one.

pawelprazak:
API that I'm working on:
```
PulumiTest.withOptions(new TestOptions(false))
  .monitorMocks(new MocksTest.MyMocks())
  .useMockRunner() // vs .useRealRunner()
  .build()
```

```
    /**
     * Run a Pulumi test stack callback asynchronously and return an exit code.
     *
     * @param stack the stack to run in Pulumi test runtime
     * @return a future exit code from Pulumi test runtime after running the stack
     */
    CompletableFuture<Integer> runAsync(Consumer<Context> stack);

    /**
     * Run a Pulumi test stack callback asynchronously and return {@link TestResult}.
     *
     * @param stack the stack to run in Pulumi test runtime
     * @return a future {@link TestResult} from Pulumi test runtime after running the stack
     */
    CompletableFuture<TestResult> runTestAsync(Consumer<Context> stack);
```

```
public class TestResult {

    private final int exitCode;
    private final List<Exception> exceptions;
    private final List<Resource> resources;
    private final List<String> errors;
    private final Stack stack;
    ...
}
```

There will also be `PulumiTestInternal` with a lot more API surface area for internal testing purposes, in fact, `PulumiTest` is just an interface that limits the internal one to the externally useful subset.

I'm also leaning towards having `.useRealRunner()` as the default and `.useMockRunner()` as an option, but I don't have a strong opinion on that one.

The fact that we set a global singleton is an implementation detail, and it is also true for the `Pulumi.run` itself, and not specific to the test API, so we should document it but making is as part of API does not feel like thinking ahead, esp. since we may change this in the future.

pawelprazak:
[...]

PR #642, opened 2022-06-01
Simplify Stack and Runner

pawelprazak:
# Description

This PR focuses on decoupling Stack and Runner, and as a consequence, simplifies both.

Goals:
- decouple Stack from Runner so that Runner knows nothing about Stack
- simplify Runner interface
- maintain "short-circuiting semantics" of the runner (and running the user's code in a context of a stack in general)

Non-goals:
- do not alter significantly other parts of the code, e.g. the testing API, try to stay on target

Changes:
- decouple Stack from Runner
- improve test results API
- make the Stack initialization clear (and gathered in one place)
- simplify and generify the Runner/DefaultRunner
- simplify Stack and move it to an internal package
- remove unused code

I believe we will benefit from those changes when adding features like Stack transformations and Automation API and also when writing internal tests (simpler tests, more things can be tested now, more things can be tested in isolation).

This will also help to polish the test framework.

Part of #626

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
@@ -1706,21 +1706,25 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
             registerTask("DefaultRunner#runAsync", valueFuture);
             // loop starts after the callback
             return valueFuture
-                    .thenCompose(__ -> whileRunningAsync())
-                    .handle((code, throwable) -> {
+                    .thenCompose(value -> whileRunningAsync().thenApply(__ -> value))
+                    .handle((value, throwable) -> {
                         if (throwable != null) {
                             return handleExceptionAsync(throwable).thenApply(errorCode ->
                                     new Result<>(
                                             errorCode,
                                             ImmutableList.copyOf(this.swallowedExceptions),
-                                            Optional.ofNullable(valueFuture.join()) // the future is already complete
+                                            Optional.ofNullable(value)
                                     )
                             );
                         }
+                        // Getting error information from a logger is slightly ugly, but that's what C# implementation does
+                        var code = this.engineLogger.hasLoggedErrors()
+                                ? ProcessExitedBeforeLoggingUserActionableMessage
+                                : ProcessExitedSuccessfully;
                         return CompletableFuture.completedFuture(new Result<>(
                                 code,
                                 ImmutableList.copyOf(this.swallowedExceptions),
-                                Optional.ofNullable(valueFuture.join()) // the future is already complete
+                                Optional.ofNullable(value)
                         ));
                     })
                     .thenCompose(Function.identity()); // we return a future from logging, and we need to flat-map here
@@ -1755,12 +1759,7 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
         // 2. Any task throws an exception.
         // So the resulting semantics is that we complete
         // when remaining count is zero, or when an exception is thrown.
-        private CompletableFuture<Integer> whileRunningAsync() {
-            // Getting error information from a logger is slightly ugly, but that's what C# implementation does
-            Supplier<Integer> exitCode = () -> this.engineLogger.hasLoggedErrors()
-                    ? ProcessExitedBeforeLoggingUserActionableMessage
-                    : ProcessExitedSuccessfully;
-
+        private CompletableFuture<Void> whileRunningAsync() {
             // Wait for every task and remove from inFlightTasks when completed
             Consumer<CompletableFuture<Void>> handleCompletion = (task) -> {
                 try {
@@ -1794,7 +1793,7 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
                 }
             };
 
-            Supplier<CompletableFuture<Integer>> loopUntilDone = () -> {
+            Supplier<CompletableFuture<Void>> loopUntilDone = () -> {
                 // Keep looping as long as there are outstanding tasks that are still running.
                 while (inFlightTasks.size() > 0) {
                     this.standardLogger.log(Level.FINEST, String.format("Remaining tasks [%s]: %s", inFlightTasks.size(), inFlightTasks));
@@ -1811,14 +1810,13 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
                                 // will attempt again in the next iteration
                             }
                         } catch (Exception e) {
-                            return handleExceptionAsync(e);
+                            return CompletableFuture.failedFuture(e);
                         }
                     }
                 }
 
-                // There were no more tasks we were waiting on.
-                // Quit out, reporting if we had any errors or not.
-                return CompletableFuture.completedFuture(exitCode.get());
+                // There were no more tasks we were waiting on. Quit out.
+                return CompletableFuture.completedFuture((Void) null);
             };
             return CompletableFuture.supplyAsync(loopUntilDone).thenCompose(f -> f);
         }


sdk/java/pulumi/src/main/java/com/pulumi/context/internal/ContextInternal.java
```diff
@@ -30,15 +30,14 @@ public ContextInternal(
             String stackName,
             LoggingContextInternal logging,
             ConfigContextInternal config,
-            OutputContextInternal outputs,
-            Map<String, Output<?>> exports
```
pawelprazak, line 34: I've realized that I cannot think of any scenario where this would be pre-set. Please correct me if I'm wrong.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -1700,43 +1699,31 @@ public DefaultRunner(Logger standardLogger, EngineLogger engineLogger) {
...
-
-            return whileRunningAsync();
+        public <T> CompletableFuture<Result<T>> runAsync(Supplier<T> callback) {
+            var valueFuture = CompletableFuture.supplyAsync(callback);
+            // run the callback asynchronously in the context of the error handler
+            registerTask("DefaultRunner#runAsync", valueFuture);
+            // loop starts after the callback
+            return valueFuture
+                    .thenCompose(__ -> whileRunningAsync())
+                    .handle((code, throwable) -> {
```
pawelprazak, line 1710: I wanted to make sure we always handle exceptions.

sdk/java/pulumi/src/main/java/com/pulumi/internal/PulumiInternal.java
```diff
@@ -42,28 +44,35 @@ public static PulumiInternal fromEnvironment() {
...
-                        }
-                )
+    public CompletableFuture<Integer> runAsync(Consumer<Context> stackCallback) {
+        return runAsyncResult(stackCallback).thenApply(r -> r.exitCode());
+    }
+
+    protected CompletableFuture<Result<Stack>> runAsyncResult(Consumer<Context> stackCallback) {
+        // Stack must be created and set globally before running any user code
+        return runner.runAsync(
+                () -> Stack.factory(
```
pawelprazak, line 66: Wanted to make it clear what is the sequence of events.

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
```diff
@@ -502,61 +499,6 @@ static public ProviderResource getProvider(Resource resource, String moduleMembe
...
-         * Finds all {@link Output} fields annotated with {@link Export},
-         * validates the annotation presence and content,
-         * validates non-null-ness of fields,
-         * ensures the field type is {@link Output}
-         * and uses reflection to get the references to them.
-         * Returns a map of export names and output references.
-         * Not to be confused with {@link com.pulumi.core.internal.OutputCompletionSource#from(com.pulumi.resources.Resource)}
-         */
-        @InternalUse
-        public static Map<String, Output<?>> findOutputs(Object object) {
```
pawelprazak, line 516: No longer used thanks to previous refactorings.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static java.util.Objects.requireNonNull;
+
+@InternalUse
+@ParametersAreNonnullByDefault
+public final class Stack extends ComponentResource {
```
pawelprazak, line 21: I wanted to radically simplify this class so there is no magic in the constructor, apart from the inherited magic)

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+    public Output<Map<String, Output<?>>> getOutputs() {
+        return this.outputs;
+    }
+
+    @InternalUse
+    public static Function<Supplier<Map<String, Output<?>>>, Stack> factory(
+            String projectName,
+            String stackName,
+            List<ResourceTransformation> resourceTransformations
+    ) {
```
pawelprazak, line 76: Wanted to make it clear what is the sequence of events and to have the whole init sequence in one place to help with maintaining and debugging in the future.

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -0,0 +1,45 @@
...
+import com.pulumi.internal.PulumiInternal;
+import com.pulumi.test.PulumiTest;
+import com.pulumi.test.TestResult;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+public class PulumiTestInternal extends PulumiInternal implements PulumiTest {
```
pawelprazak, line 18: This will get completed in the future.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentExceptionTest.java
```diff
@@ -38,14 +38,16 @@ static void cleanup() {
     @Test
     void testUrnFutureDoesNotHangOnException() {
         mock.standardLogger.setLevel(Level.OFF);
-        assertThatThrownBy(() -> mock.runTestAsync(ctx -> {
+        var result = mock.runTestAsync(ctx -> {
```
pawelprazak, line 41: the tests no longer throw before `join`

sdk/java/pulumi/src/test/java/com/pulumi/resources/internal/StackTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+        //noinspection unchecked
+        verify(mock.deployment, times(1))
+                .registerResourceOutputs(any(Resource.class), any(Output.class));
+
+        assertThat(result.outputs()).isEmpty();
+    }
+
+    @Test
+    void testImmediateStackReference() {
```
pawelprazak, line 111: new test casse

sdk/java/pulumi/src/main/java/com/pulumi/context/internal/ContextInternal.java
t0yv0: I'm not sure... at this point I'm leaning to "if it's not tested it does not exist" but we do need to build up some test coverage for how users interact with this.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -1700,43 +1699,31 @@ public DefaultRunner(Logger standardLogger, EngineLogger engineLogger) {
...
+                        if (throwable != null) {
+                            return handleExceptionAsync(throwable).thenApply(errorCode ->
+                                    new Result<>(
+                                            errorCode,
+                                            ImmutableList.copyOf(this.swallowedExceptions),
+                                            Optional.ofNullable(valueFuture.join()) // the future is already complete
```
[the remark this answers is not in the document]
t0yv0: It completed with an exception though, why do you think it would have a value? Naively I'd expect `.join()` to throw here but I'm not sure, better try it out and see.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
t0yv0: Doesn't registerTask handle exceptions also? 

This might be ok but a bit convoluted to reason about.

t0yv0: Let me check where runAsync is called.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -1700,43 +1699,31 @@ public DefaultRunner(Logger standardLogger, EngineLogger engineLogger) {
...
+                                    )
+                            );
+                        }
+                        return CompletableFuture.completedFuture(new Result<>(
+                                code,
+                                ImmutableList.copyOf(this.swallowedExceptions),
+                                Optional.ofNullable(valueFuture.join()) // the future is already complete
```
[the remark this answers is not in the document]
t0yv0: Here, yes.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
t0yv0: OK. This implements the project entry point. I think this isn't right. This code makes the `CompletableFuture<Result<T>>` ALWAYS wait for `valueFuture` which removes the feature of early termination due to an out-of-band exception somewhere in the underlying net of futures. That is, our protocol up to now, was that if any exception happens anywhere in the Output chain, we bubble up and stop. This is the point of registerTask and whileRunningAsync.

Is this an intentional change to the semantics?

I wonder what happened with the test that tested for it.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+    @InternalUse
+    public static final String RootPulumiStackTypeName = "pulumi:pulumi:Stack";
+
+    /**
+     * The outputs of this stack, if the <code>callback</code> callback exited normally.
+     */
+    private final Output<Map<String, Output<?>>> outputs;
+
+    /**
+     * Create a Stack with stack outputs defined in derived class constructor.
```
t0yv0, line 37: The comment is weird, what is "derived class constructor"? Stack is now final right?

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+        // - set the Stack instance in the global state (user code will try to reach it via Resource constructor)
+        // - register the Stack outputs in the Pulumi engine
+        // - call the user code callback (asynchronously to avoid any additional cyclic dependency problems)
+        // - complete the Stack initialization (by completing the lazyFuture)
+        return (Supplier<Map<String, Output<?>>> callback) -> {
+            // incomplete future that will be completed after the Stack is constructed
+            var lazyFuture = new CompletableFuture<Map<String, Output<?>>>();
+            // incomplete outputs that will be completed when lazyFuture completes
+            var lazyOutputs = Output.of(lazyFuture);
+            // Stack must be the first Resource to be constructed,
```
t0yv0, line 88: Not a fan of commenting every line, that's diminishing returns for comments the code speaks for itself ideally at this level.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+            // before any user code attempts to construct any Resource
+            var stack = new Stack(projectName, stackName, resourceTransformations, lazyOutputs);
```
t0yv0, line 90: I have some doubts this sequencing works, I worry about a race. Does `new Stack()` launch `RegisterResource` or some other async call in a bg thread? Or it doesn't because it's a Component? Need to read it carefully. 

It seems that it *does* call RegisterResource to the ResourceMonitor gRPC service.

Now that will race with `deploymentInstance.setStack(stack)` and  `DeploymentInternal.getInstance().registerResourceOutputs(stack, lazyOutputs);`. This may or may not be okay? Would something call `deploymentInstance.getStack()` before the RegisterResource thing?

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
```diff
@@ -0,0 +1,109 @@
...
+            // register the new Stack with the engine
+            DeploymentInternal.getInstance().setStack(stack);
+            // register the Stack outputs with the engine
+            DeploymentInternal.getInstance().registerResourceOutputs(stack, lazyOutputs);
+            // run the user code callback wrapped in a future (note that future is eager)
+            var callbackFuture = CompletableFuture.supplyAsync(callback);
+            DeploymentInternal.getInstance().getRunner().registerTask("callback", callbackFuture);
```
t0yv0, line 97: Perhaps share `DeploymentInternal.getInstance()` or even make it a param so we query the global less.

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * A test result information.
+ */
+public class TestResult {
+    private final int exitCode;
+    private final List<Resource> resources;
+    private final List<Exception> exceptions;
+    private final List<String> errors;
```
t0yv0, line 21: What's the difference between errors and exceptions?

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
+
+/**
+ * A test result information.
+ */
+public class TestResult {
+    private final int exitCode;
```
t0yv0, line 18: Why do we need to test exit codes?

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentExceptionTest.java
t0yv0: That seems actually good. Consistent.

t0yv0:
This is mostly good, as usual it would benefit from breaking down into smaller PRs but we can review as is.

1. I have some doubts about early termination in Runner that might be blocking the PR for me above; can we talk through that? You believe it works as expected? 

2. Ideally we need CHANGELOG_PENDING entry outlining all the changes to user-facing functionality; it's hard for me to even see which one is that, but at the very least: explain that Stack class is removed and why, and explain that the mock testing interface is changed and how.

sdk/java/pulumi/src/main/java/com/pulumi/context/internal/ContextInternal.java
pawelprazak: > "if it's not tested it does not exist"

I like this. Since this is an internal API, this constructor, then it can be easily added back :)

sdk/java/pulumi/src/main/java/com/pulumi/deployment/MockRunner.java
```diff
@@ -1,39 +1,19 @@
...
-    }
-
-    @Override
-    public CompletableFuture<Integer> runAsyncFuture(Supplier<CompletableFuture<Map<String, Output<?>>>> callback, StackOptions options) {
-        return null;
-    }
-
-    @Override
-    public <T extends Stack> CompletableFuture<Integer> runAsync(Supplier<T> stackFactory) {
+    public <T> CompletableFuture<Result<T>> runAsync(Supplier<T> callback) {
```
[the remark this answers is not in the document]
pawelprazak: > Why is this class public and marked internal etc? Do you see users interacting with it at any point?
I wanted to move it to an internal package. I see no reason for it to be a public API. (ofc the public keyword will have to be there to be visible from other internal packages.

I didn't want to make too many changes in too many places at once. I plan to address the testing API in #553

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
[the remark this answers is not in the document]
pawelprazak: Hmm, that's a good point, does it throw, it should throw on `join`.
I can't exclude a situation where both the value and an exception is present, but not sure if it is a useful state. 
Maybe simplifying this to always return Optional.empty() could be a better option, but I was worried this will make some tests impossible/difficult to write.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
pawelprazak: > Is this an intentional change to the semantics?
No, it is not intentional. I was under the impression, this maintains the semantics, but I'm not sure anymore.

> I wonder what happened with the test that tested for it.
I'd guess it either does not test this conditions well enough or the semantics didn't actually changed.

I was under the impression that `valueFuture` would short circuit on an exception and not even call the `.thenCompose(__ -> whileRunningAsync())` and go straight to `handle`.

Let me try to come up with a test for those short-circuiting semantincs.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/Runner.java
```diff
@@ -1,14 +1,18 @@
...
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Supplier;
 
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A task {@link CompletableFuture} runner.
+ */
+@InternalUse
 public interface Runner {
```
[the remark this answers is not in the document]
pawelprazak: > 1. why does it deal with exit codes at all - at this level?
yes, this is confusing for me as well, I was thinking about moving the error handling out of the runner, or at least out of the mian loop and do it once with a `.handle` not multiple times

> 2. why does the caller need to get swallowed exceptions, [...]
I've removed `getSwallowedExceptions` and have put the exceptions into the `Result`. Right now it is needed for the tests only. Not sure if universally useful. The functionality was not originally there, I've ported it from C# while trying to catch up with upstream changes.

> 4. [...]
I'd love to be able to simplify the signature even more: `CompletableFuture<Void> run()` I'm just not sure if we can guarantee the stack will be created before user code is called just by registering both and making the user's code execution lazy/wait for stack

Then we could make the error handling on the boundary in PulumiInternal, instead of the Runner. 
This way a Runner would do just one thing and be super easy to test and reason about.

I just have a feeling it's not gonna be so easy, but we might try anyway :D

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: I agree, ideally code should speak for itself. After struggling to understand the sequence and the reasons behind it myself I wanted to help everyone else who might need to do the same in the future.
Would rewriting the comments into a one big description on to be more readable for you?

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: Great points. My understanding is that `getStack()` will only be called for all other resources then the `Stack resource.

Currently there are two places where `getStack` is called:
- (read or register resource) `prepareResourceAsync` -> `getRootResourceAsync` - if a resource has no parent and is not a Stack
- Resource constructor - if a resource has no parent and is not a Stack

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
pawelprazak: errors are error messages from logger, exceptions are exception instances

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
```diff
@@ -118,7 +118,7 @@ void testStack() {
                 .build();
 
         var result = mock.runTestAsync(MocksTest::myStack).join();
-        var publicIp = waitForValue(result.stackOutput("publicIp", String.class));
```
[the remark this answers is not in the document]
pawelprazak: if we want to resonate with `pulumi stack output` a natural thing IMO would be to have:

```
PulumiTest.run(ctx -> ...)
    .stack.output("foo", String.class);
```

So `Stack` in the Result itself.

pawelprazak:
> smaller PRs but we can review as is

Sorry about the volume of changes, I couldn't make smaller PRs to make them sensible - they would appear rather random I'm afraid without the context.

Maybe stacking multiple smaller PR would improve things? Although not convinced about the benefits of smaller PR vs having multiple commits in a PR, eps. when untangling things, they tend to touch many places in the code.

I'll keep that in mind thou, maybe next time I'll find a better way to make a PR smaller and easier to review.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: Regarding outputs registration, AFAIU, `lazyOutputs` being not completed will block the registration until completed via `whenComplete`.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
pawelprazak: After adding more test cases I was not successful in triggering the `if (throwable != null)` branch.

It turns out it might be redundant with the code in the loop itself. I wonder if we could remove the handler form the loop and handle exceptions here instead maintaining the semantics.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: done

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
pawelprazak: we terminate on a logged error even if there were no exceptions (behaviour inherited from C# impl)

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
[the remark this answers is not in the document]
pawelprazak: After moving error handling to one place, the join is no longer needed. I'm glad this small hack of mine got deleted :)

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
pawelprazak: Yes, after moving the handler to only be present in `handle` all tests pass and it looks equivalent (in behaviour) to me looking at the code.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: fixed

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
pawelprazak: I've moved the comments/clarify, hope it helps.

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
pawelprazak: Yes, after giving this a though, exit codes are special case that works only for the CLI use and are not very useful to the user anyway. Moving the exit codes out of runner and removing from test result seam like a good idea.
Thank you for spotting this inconsistency.

sdk/java/pulumi/src/main/java/com/pulumi/resources/internal/Stack.java
t0yv0: Good! In that case we should be OK here!

sdk/java/pulumi/src/main/java/com/pulumi/test/TestResult.java
t0yv0: Gotcha, so this basically captures what was logged.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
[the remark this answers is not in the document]
t0yv0: I like it. Good mnemonic!

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -1706,21 +1706,25 @@ public <T> CompletableFuture<Result<T>> runAsync(Supplier<T> callback) {
             registerTask("DefaultRunner#runAsync", valueFuture);
             // loop starts after the callback
             return valueFuture
-                    .thenCompose(__ -> whileRunningAsync())
-                    .handle((code, throwable) -> {
+                    .thenCompose(value -> whileRunningAsync().thenApply(__ -> value))
```
t0yv0, line 1709: Why not this:

```
whileRunningAsync().thenCompose(_ -> valueFuture)
```

Or use `thenCombine` API.

What I am trying to see is that when `valueFuture` takes a long time, but there's an exception out of band, this exception short-circuits `valueFuture` instead of waiting on it. Let me try to make a test.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/Runner.java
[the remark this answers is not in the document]
pawelprazak: I agree, I've gathered the feedback regarding possible future directions in #668 for further considerations.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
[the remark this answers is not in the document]
pawelprazak: Added to #668 for reference.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
pawelprazak: I'm slightly inclined to leave as-is, because this looks more confusing in my opinion:
`valueFuture.thenCompose(__ -> whileRunningAsync().thenCompose(___ -> valueFuture))` 
than this
`valueFuture.thenCompose(value -> whileRunningAsync().thenApply(__ -> value))`

Unless I'm missing something, they are equivalent in behaviour, right? (Tests are green in both cases)

PR #666, opened 2022-06-03
Fix complex condition

t0yv0:
# Description

It seems that this logic is needed to short-circuit tests that are not allowed for community PRs and dependabot PRs, but it accidentally short-circuits tests on `main` also, so the tests that check every commit on `main` do not run the full set of tests (providers and so on). This is unfortunate. The PR aims to fix that by expanding the condition.

Example: https://github.com/pulumi/pulumi-java/actions/runs/2436022397

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -102,8 +102,13 @@ jobs:
       - name: Publish Pulumi Java SDK
         run: make sdk.publish
         # See Note 1.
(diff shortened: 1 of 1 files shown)


t0yv0:
Needs https://github.com/pulumi/pulumi-java/pull/667 first possibly.

dixler (approved):

PR #669, opened 2022-06-06
Bump aws-apigateway provider to 0.0.9

pawelprazak:
# Description

## Checklist


stack72 (approved):

t0yv0:
Do we need to release as well to Maven Central? We need a ticket for that.

pawelprazak:
We can do it at earliest convenience, it does not fix a known issue so don't think we need to hurry.

PR #671, opened 2022-06-06
Fix #670

### LINKED ISSUE #670: Output-versioned function invoke fails to preview in C#/.NET
### What happened?

Original issue:
https://github.com/pulumi/pulumi/issues/8322

Current `invoke` method in DeploymentImpl has the same problem as the C# (and Python) before the fix. If Invoke is called with unknowns, it currently tries to send those unknowns to the resource monitor; the monitor does not understand those and we get obscure errors. The desired behavior is to short-circuit and not call the monitor at all if there are any unknowns in the Invoke args, instead returning an Unknown output.

*Note*: this is exactly the behavior we'd get if we implemented Output versioned invokes as wrappers:

```
Output.tuple(arg1, arg2, arg3).apply(tup -> Output.of(Func.invokePlain(..tup...)))`
```

In that implementation, invokePlain does not even execute if any of the args have unknowns. Unfortunately working around limitations of the ser/de framework, current implementation of Output-versioned invokes uses a DeploymentImpl primitive instead of the wrapper, and the primitive carries this bug.

### Steps to reproduce

See https://github.com/pulumi/pulumi/issues/8322

Also reproduces on our WIP PR https://github.com/pulumi/pulumi-java/pull/612 - one of our template tests for Azure caches this.

### Expected Behavior

Do not call Invoke with unknowns.

### Actual Behavior

Calls Invoke with unknowns.

### Versions used

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
# Description

Fixes #670 

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentInvokeTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentInvokeTest.java
@@ -130,7 +130,7 @@ public class DeploymentInvokeTest {
             var args = new IdentityArgs(unk);
             ctx.export("out", IdentityFunctions.invokeIdentity(args, new InvokeOptions()));
         }).join().throwOnError();
-        assertThat(OutputTests.waitFor(result.stackOutput("out")).isKnown()).isFalse();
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj (approved):
Looks great and the tests are clear to follow, thanks for looking into this 🙏

PR #672, opened 2022-06-06
Test extern examples from pulumi/examples

t0yv0:
# Description

- Deletes examples that have a copy in `pulumi/examples`
- Tests currently built SDK and Providers against those examples by fetching them via submodule
- Since we have breaking unreleased changes on main, uses a branch  adapted to the changes - `pulumi-java/main`
- Adds testing for the graal example that was previously excluded from the tests
- Fixes a bug where some templates and examples did not get tested against the right SDK/prov versions

Fixes # (issue)

## Checklist


tests/internal/pin_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+package tests
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFixupGradleVersion(t *testing.T) {
```
pawelprazak, line 11: maybe https://github.com/ben-manes/gradle-versions-plugin could help with this instead of a custom tool, but I haven't checked, just a thought

Issue #594, opened 2022-05-16
Deserialization error while trying to use pulumi-awsgateway


slnowak:
### What happened?

When trying to run a minimalistic example with `aws-apigateway`, pulumi throws the following error:
```
➜  java git:(master) ✗ pulumi up -y
Please choose a stack, or create a new one: dev
Previewing update (dev)

View Live: https://app.pulumi.com/slnowak/iac-workshop-lambda-api-gateway/dev/previews/c164ea18-8eca-49fd-a031-75c4ad640d60

     Type                             Name                                 Plan       Info
 +   pulumi:pulumi:Stack              iac-workshop-lambda-api-gateway-dev  create     2 errors; 16 messages
 +   ├─ aws:iam:Role                  lambdaRole                           create     
 +   ├─ aws:iam:RolePolicyAttachment  basicExecutionRole                   create     
 +   ├─ aws:lambda:Function           helloWorldFunction                   create     
 +   └─ apigateway:index:RestAPI      helloWorldApi                        create     
 +      ├─ aws:apigateway:RestApi     helloWorldApi                        create     
 +      ├─ aws:apigateway:Deployment  helloWorldApi                        create     
 +      ├─ aws:lambda:Permission      helloWorldApi-fa520765               create     
 +      └─ aws:apigateway:Stage       helloWorldApi                        create     
 
Diagnostics:
  pulumi:pulumi:Stack (iac-workshop-lambda-api-gateway-dev):
    error: Running program [PID: 13747](/usr/local/opt/openjdk/libexec/openjdk.jdk/Contents/Home/bin/java -classpath /usr/local/Cellar/maven/3.8.5/libexec/boot/plexus-classworlds-2.6.0.jar -Dclassworlds.conf=/usr/local/Cellar/maven/3.8.5/libexec/bin/m2.conf -Dmaven.home=/usr/local/Cellar/maven/3.8.5/libexec -Dlibrary.jansi.path=/usr/local/Cellar/maven/3.8.5/libexec/lib/jansi-native -Dmaven.multiModuleProjectDirectory=/Users/snowak/work/internal/pulumi/python_course/lambda/module2/java org.codehaus.plexus.classworlds.launcher.Launcher --quiet --no-transfer-progress compile exec:java) failed with an unhandled exception:
[...]

PR #678, opened 2022-06-07
Cleanup Resource dependencies code

pawelprazak:
# Description

Code style and readability. No behaviour changed.

Small cleanup while working on #594 

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
@@ -85,7 +85,6 @@ import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
@@ -884,8 +883,8 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
 
                                                                                 log.excessive("Processing a remote ComponentResource: t=%s, name=%s, custom=%s, remote=%s", type, name, custom, remote);
 
-                                                                                BiFunction<Optional<ProviderResource>, List<ProviderResource>, Map<String, CompletableFuture<String>>> convertProviders =
-                                                                                        (provider, providers) -> Stream.concat(provider.stream(), providers.stream())
+                                                                                Function<List<ProviderResource>, Map<String, CompletableFuture<String>>> convertProviders =
+                                                                                        (providers) -> providers.stream()
                                                                                                 .map(Internal::from)
                                                                                                 .collect(toMap(
                                                                                                         ProviderResource.ProviderResourceInternal::getPackage,
@@ -893,7 +892,7 @@ public class DeploymentImpl extends DeploymentInstanceHolder implements Deployme
                                                                                                 ));
 
                                                                                 providerFutures = CompletableFutures.allOf(
-                                                                                        convertProviders.apply(componentOpts.getProvider(), componentOpts.getProviders())
+                                                                                        convertProviders.apply(componentOpts.getProviders())
                                                                                 ).thenApply(ImmutableMap::copyOf);
                                                                             } else {
                                                                                 providerFutures = CompletableFuture.completedFuture(ImmutableMap.of());


sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -854,35 +858,43 @@ private CompletableFuture<PrepareResult> prepareResourceAsync(
...
 
+                                                    // If no parent was provided, parent to the root resource.
                                                     var parentUrn = options.getParent().isPresent()
                                                             ? Internal.of(options.getParent().get().getUrn()).getValueOptional()
                                                             : this.rootResource.getRootResourceAsync(type);
                                                     return parentUrn.thenCompose(
                                                             (Optional<String> pUrn) -> {
                                                                 log.excessive("Got parent urn: t=%s, name=%s, custom=%s, remote=%s", type, name, custom, remote);
-                                                                var providerRef = custom
-                                                                        ? CompletableFutures.flipOptional(options.getProvider().map(p -> Internal.from(p).getRegistrationId()))
```
pawelprazak, line 864: The ternary operator was not very readable here, too many things going on at once IMO.

Use of `final` local var guarantees the same behaviour (set var exactly once).

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -854,35 +858,43 @@ private CompletableFuture<PrepareResult> prepareResourceAsync(
...
+                                                                            // For remote resources, merge any provider opts into a single dict,
+                                                                            // and then create a new dict with all the resolved provider refs.
                                                                             final CompletableFuture<ImmutableMap<String, String>> providerFutures;
                                                                             if (remote && options instanceof ComponentResourceOptions) {
                                                                                 var componentOpts = (ComponentResourceOptions) options;
 
-                                                                                // TODO: C# had the following logic here:
-                                                                                //          "If only the Provider opt is set, move it to the Providers list for further processing."
-                                                                                //       But our ComponentResourceOptions should guarantee the desired semantics.
```
pawelprazak, line 875: I've decided to implement this merging anyway, just in case.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -854,35 +858,43 @@ private CompletableFuture<PrepareResult> prepareResourceAsync(
...
-                                                                                //       It would be great to add more tests and maybe remove 'provider' in favour of 'providers' only
+                                                                                log.excessive("Processing a remote ComponentResource: t=%s, name=%s, custom=%s, remote=%s", type, name, custom, remote);
+
+                                                                                BiFunction<Optional<ProviderResource>, List<ProviderResource>, Map<String, CompletableFuture<String>>> convertProviders =
+                                                                                        (provider, providers) -> Stream.concat(provider.stream(), providers.stream())
```
pawelprazak, line 888: Prepending `provider.stream()` to `providers.stream()` is the only behavioral change in this PR, but it should not matter since we handle this in Resource constructor if I'm not mistaken.

I've decided that adding this for completeness is probably not a bad idea.

t0yv0: Unfortunately it's possibly a semantic change - I don't understand this change therefore cannot approve this; we'd need tests to clarify what's going on, and possibly reference to another SDK to show the desired behavior this change is trying to achieve.

What would be really helpful is to detangle this. It seems that:

- we are computing RegisterResourceRequest Providers field (this is our output) 
- we have as an input ComponentResourceOptions.getProviders
- there is tricky interaction with `provider` field, so possibly that is an input as well
- not clear what's responsibility of this code vs resource constructor - can you clarify that comment? 
-  ealing with this at 12 level of indent is not helpful

Let me ask @iwahbe about related work on provider vs provider and where we landed with that. I found https://docs.google.com/document/d/1Z0P-XU5AW-dhwKQCWHEiau_7kIAsF9j4CUFaPlMlGj4/edit#heading=h.4brqwrn4u47v but I need to check what the state of the project is.

t0yv0: I've filed a new issue that represents better how Ian's work actually ended unifying provider/providers across SDKs. There's some good stuff here. I suggest we treat it as a separate work item and do thoroughly when time permits. Tyvm. The issue is #681

t0yv0 (approved):
Thank you, looks strictly a no-op now, good.

PR #682, opened 2022-06-08
Fix 576 adding multi-project Gradle support

### LINKED ISSUE #576: Gradle multi-project support
## Hello!

## Issue details
A "mono-repo" approach is currently not supported (at least ootb) through grade as `pulumi` is invoking everything from the current folder instead of using Gradle with correct args such as the subproject name.

A simple setup containing two separate pulumi projects with different stacks. They are defined as two separate subprojects `:services:app-cluster` and `:services:mgmt-cluster`.

*root settings.gradle(.kts)*
```
rootProject.name = "iac"
include(
    "libs:modules",
    "services:app-cluster",
    "services:mgmt-cluster",
    "testing:commons",
)
```

*Simplified dir structure*
```
settings.gradle
build.gradle
services/
  -> app-cluster/
       * build.gradle
       * Pulumi.yaml
       * Pulumi.dev.yaml
       * Pulumi.staging.yaml
       * Pulumi.production.yaml
   -> mgmt-cluster/
       * build.gradle
       * Pulumi.yaml
       * Pulumi.production.yaml
```

You shouldn't need to have any `settings.gradle` inside of the subproject directories but trying to run pulumi inside one of the subprojects you get the following error message.
```
error: failed to load language plugin java: could not read plugin [/home/nickjn92/.pulumi/bin/pulumi-language-java] stdout: EOF
```

Also it might be classic pebcak but I'm currently getting errors trying to run this by adding `gradle.settings` to each subproject and running `pulumi up` from those directories
```
  pulumi:pulumi:Stack (iac-app-cluster-kotlin-dev):
    Exception in thread "main" java.lang.IllegalArgumentException: Program run without the Pulumi engine available; re-run using the `pulumi` CLI
...
Caused by: java.lang.IllegalArgumentException: expected environment variable: 'PULUMI_MONITOR', got: ...

```

t0yv0:
# Description

Factors out executors and improves mutli-project gradle executor detection.

Fixes #576 

## Checklist

### DIFF
--- a/pkg/internal/executors/executor_test.go
+++ b/pkg/internal/executors/executor_test.go
@@ -18,7 +18,7 @@ func TestGradleSimple(t *testing.T) {
 			"settings.gradle":  {},
 			"app/build.gradle": {},
 		})
-	exec, err := ConfigureExecutor(JavaExecutorOptions{WD: fsys})
+	exec, err := NewJavaExecutor(JavaExecutorOptions{WD: fsys})
 	assert.NoError(t, err)
 	assert.Equal(t, "/usr/bin/gradle", exec.Cmd)
 	assert.Equal(t, ".", exec.Dir)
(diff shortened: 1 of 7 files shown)


pkg/internal/executors/executor_test.go
```diff
@@ -0,0 +1,252 @@
...
+import (
+	"testing"
+	"testing/fstest"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi-java/pkg/internal/fsys"
+)
+
+func TestGradleSimple(t *testing.T) {
```
pawelprazak, line 14: nice! unit test!

pawelprazak (approved):

PR #683, opened 2022-06-09
Add unit testing examples

pawelprazak:
# Description

Fixes #553

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,7 @@
 ### Improvements
 
+- [sdk] #553: Added `com.pulumi.test` package with new idiomatic Java API for [unit testing Pulumi programs]
+  (https://www.pulumi.com/docs/guides/testing/#unit-testing) against mocks. 
+  See ([example](https://github.com/pulumi/pulumi-java/tree/main/tests/examples/testing-unit-java)) of using the new API.
+
 ### Bug Fixes


pawelprazak:
I'm sorry to drop this many changes at once, but I don't see how I can slice it up in a meaningful way.
If it helps to review, I could group subsets of those commits into multiple PRs stacked on top of eachother.

CHANGELOG_PENDING.md
```diff
@@ -13,4 +13,9 @@
 
 - Fix #478: New API Pulumi.withOptions adds support for stack transformations
 
+- [sdk] PulumiTest and PulumiTestInternal replaces DeploymentTests,
```
pawelprazak, line 16: not sure if we should consider this breaking change or not, but the public API for tests has changed significantly

sdk/java/pulumi/src/main/java/com/pulumi/core/function/ServiceProvider.java
```diff
@@ -1,7 +0,0 @@
-package com.pulumi.core.function;
-
-// FIXME: some core c# interface, will probably not be needed, not in this form at lest,
-//        looks like some kind of injection
-public interface ServiceProvider {
```
pawelprazak, line 5: an old misunderstanding while porting initial PoC, somehow survived this long ;)

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/ReadOrRegisterResource.java
```diff
@@ -0,0 +1,14 @@
+package com.pulumi.deployment.internal;
+
+import com.pulumi.resources.Resource;
+import com.pulumi.resources.ResourceArgs;
+import com.pulumi.resources.ResourceOptions;
+
+import java.util.function.Function;
+
+public interface ReadOrRegisterResource {
```
pawelprazak, line 9: adding targeted interfaces helps with the overall architecture by keeping coupling at minimum

sdk/java/pulumi/src/main/java/com/pulumi/test/Mocks.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Invoked when the program needs to call a provider to load data (e.g., to retrieve an existing resource).
+     * <p>
+     * Default implementation returns an empty map.
+     *
+     * @param args arguments containing call information
+     * @return the call invocation result, can be either a POCO or a dictionary bag.
+     */
+    default CompletableFuture<Map<String, Object>> callAsync(CallArgs args) {
```
pawelprazak, line 35: I've provided a default implementation that does nothing so the users don't have to implement it when they don't need it. Was thinking about making this throw instead, but was not sure. WDYT?

sdk/java/pulumi/src/main/java/com/pulumi/test/Mocks.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Invoked when a new resource is created by the program.
+     * <p>
+     *
+     * @param args arguments containing resource information
+     * @return A resource identifier and resource state. State can be either a POJO
+     * or a dictionary bag. The returned ID may be null for component resources.
+     */
+    CompletableFuture<ResourceResult> newResourceAsync(ResourceArgs args);
```
pawelprazak, line 25: A new cleaner signature without `Tuple2`

sdk/java/pulumi/src/main/java/com/pulumi/test/Mocks.java
```diff
@@ -0,0 +1,163 @@
...
+     * @return the call invocation result, can be either a POCO or a dictionary bag.
+     */
+    default CompletableFuture<Map<String, Object>> callAsync(CallArgs args) {
+        return CompletableFuture.completedFuture(Map.of());
+    }
+
+    /**
+     * ResourceArgs for use in {@link #newResourceAsync(ResourceArgs)}
+     */
+    class ResourceArgs {
```
pawelprazak, line 42: moved from a separate file, this way it does not need to be imported and can have a shorter name

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
```diff
@@ -10,12 +14,127 @@
...
+    /**
+     * Run a Pulumi test stack callback, wait for result and return {@link TestResult}.
+     *
+     * @param stack the stack to run in Pulumi test runtime
+     * @return a {@link TestResult} from Pulumi test runtime after running the stack
+     * @see #runTestAsync(Consumer)
+     * @see #withOptions(TestOptions)
+     * @see #withMocks(Mocks)
+     */
+    static TestResult runTest(Consumer<Context> stack) {
```
pawelprazak, line 26: `PulumiTest` echoes the `Pulumi` interface semantics, with the static methods for DSL-style of API.

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
```diff
@@ -10,12 +14,127 @@
...
+     * @return the underlying future value of the given {@link Output}
+     */
+    static <T> CompletableFuture<T> extractValueAsync(Output<T> output) {
+        return Internal.of(output).getValueNullable();
+    }
+
+    /**
+     * Pulumi test entrypoint operations.
+     */
+    interface API {
```
pawelprazak, line 100: I was thinking about renaming this interface to `DSL` (here and in `Pulumi` ofc), WDYT?

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
```diff
@@ -10,12 +14,127 @@
...
+        TestResult runTest(Consumer<Context> stack);
 
+        /**
+         * Run a Pulumi test stack callback asynchronously and return {@link TestResult}.
+         *
+         * @param stack the stack to run in Pulumi test runtime
+         * @return a future {@link TestResult} from Pulumi test runtime after running the stack
+         */
+        CompletableFuture<TestResult> runTestAsync(Consumer<Context> stack);
+    }
```
pawelprazak, line 139: Would providing a `builder()` as well be useful to the users?
We use builder internally for all of the tests, mostly because our internal tests are more complex.

I didn't want to add the builder here without being sure it will be useful for the users.

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -1,34 +1,139 @@
...
+     */
+    @InternalUse
+    public MockMonitor monitor() {
+        return this.monitor;
+    }
+
+    /**
+     * @return return the {@link ReadOrRegisterResource} used by the test
+     */
+    public ReadOrRegisterResource readOrRegisterResource() {
```
pawelprazak, line 116: used by StackTest at the moment, but I can see how this method of exposing targeted internal interfaces to the internal tests can give us scalability and freedom to refactor in the future if needed

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -42,4 +147,274 @@ public CompletableFuture<TestResult> runTestAsync(
...
+    }
+
+    @InternalUse
+    @CanIgnoreReturnValue
+    public static <T> OutputData<T> extractOutputData(Output<T> output) {
+        return Internal.of(output).getDataAsync().join();
+    }
+
+    @InternalUse
+    public static final class APIInternal implements PulumiTest.API {
```
pawelprazak, line 167: this is where we combine the builder with the operations for a DSL used as part of the top level API

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -42,4 +147,274 @@ public CompletableFuture<TestResult> runTestAsync(
...
+         * @return a {@link PulumiTestInternal} instance created from this {@link Builder}
+         */
+        public PulumiTestInternal build() {
+            if (this.options == null) {
+                this.options = TestOptions.Empty;
+            }
+            if (this.standardLogger == null) {
+                this.standardLogger = defaultLogger();
+            }
+            // FIXME: this instance is different then the one in DeploymentState
```
pawelprazak, line 313: some unfortunate consequences of the current way of initializing the Deployment, we'll address those separately

tests/examples/testing-unit-java/src/test/java/com/pulumi/example/unittest/WebserverInfraTest.java
```diff
@@ -0,0 +1,82 @@
...
+                .runTest(WebserverInfra::stack);
+
+        var instance = result.resources().stream()
+                .filter(r -> r instanceof Instance)
+                .map(r -> (Instance) r)
+                .findFirst();
+        assertThat(instance).isPresent();
+
+        return Stream.of(
+                dynamicTest("instance has name tag", () -> {
```
pawelprazak, line 41: I've used dynamic tests to be able to call run only once, I can easily rewrite this to be invoked many times like in the C# test if we don't want to use advanced features of JUnit 5 for some reason.

IMO this makes the test cleaner but I'm biased ;)

tests/examples/testing-unit-java/src/test/java/com/pulumi/example/unittest/WebserverInfraTest.java
```diff
@@ -0,0 +1,82 @@
...
+                                ).acceptsAll(ingress);
+                            });
+                })
+        );
+    }
+
+    // Mock the engine state.
+    public static class MyMocks implements Mocks {
+        @Override
+        public CompletableFuture<ResourceResult> newResourceAsync(ResourceArgs args) {
```
pawelprazak, line 76: Mocks are now easier to implement and look at :)

tests/examples/testing-unit-java/src/test/java/com/pulumi/example/unittest/WebserverInfraTest.java
t0yv0: I don't know. We should try to have the least surprising code for the Java user, I don't know what that would be. If this is idiomatic JUnit then OK. The idea is to be as much familiar to the people reading this as possible; there's a lot of Pulumi specific content to absorb and we want to be as simple as possible in every other aspect.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/TestOptions.java
```diff
@@ -15,48 +16,42 @@ public class TestOptions {
...
-     * @param preview is the test a preview or a normal execution
+     * @param projectName             the test project name to use
+     * @param stackName               the test stack name to use
+     * @param preview                 is the test a preview or a normal execution
+     * @param resourceTransformations the test stack resource transformations
      */
-    public TestOptions(@Nullable String projectName, @Nullable String stackName, boolean preview) {
-        this.projectName = projectName != null ? projectName : "project";
-        this.stackName = stackName != null ? stackName : "stack";
+    public TestOptions(
```
t0yv0: This change is good but breaking. I am guessing we don't have much code out there relying on TestOptions yet so this is ok just worth highlighting in the CHANGELOG that we have a new API here.

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
```diff
@@ -13,17 +16,68 @@
...
      */
-    CompletableFuture<TestResult> runTestAsync(Consumer<Context> stack);
+    static PulumiTest.API withMocks(Mocks mocks) {
+        return new PulumiTestInternal.APIInternal().withMocks(mocks);
+    }
 
+    /**
+     * Cleanup internal test state. <b>Must be called</b> after every run of a Pulumi test program.
+     */
+    static void cleanup() {
```
t0yv0, line 72: This one is unfortunate. Resetting globals?

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
t0yv0: Nope, it's an abbreviation some users might not know. The current name is not great but let's not change names unless there's a clear win.

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
t0yv0: Not sure, perhaps wait for user feedback.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/ReadOrRegisterResource.java
t0yv0: I'm confused where this is going. 

The architecture defines a ResourceMonitor service in gRPC. The "real" implementation is hosted inside `pulumi CLI`.

```
// ResourceMonitor is the interface a source uses to talk back to the planning monitor orchestrating the execution.
service ResourceMonitor {
    rpc SupportsFeature(SupportsFeatureRequest) returns (SupportsFeatureResponse) {}
    rpc Invoke(ResourceInvokeRequest) returns (InvokeResponse) {}
    rpc StreamInvoke(ResourceInvokeRequest) returns (stream InvokeResponse) {}
    rpc Call(CallRequest) returns (CallResponse) {}
    rpc ReadResource(ReadResourceRequest) returns (ReadResourceResponse) {}
    rpc RegisterResource(RegisterResourceRequest) returns (RegisterResourceResponse) {}
    rpc RegisterResourceOutputs(RegisterResourceOutputsRequest) returns (google.protobuf.Empty) {}
}
```

Is it still true that users are not expected to interact with it?

My high-level expectation is that in the context of testing against mocks, users implement the Mocks interface and Java SDK uses a test double ResourceMonitor that calls the users's mocks when implementing ReadResource or RegisterResource or Invoke/etc, and does something reasonable for everything else.

sdk/java/pulumi/src/test/java/com/pulumi/resources/internal/StackTest.java
```diff
@@ -109,19 +106,27 @@ void testStackWithNullOutputsThrows() {
...
+        var test = PulumiTestInternal.builder()
+                .mocks(monitorMocks)
                 .deploymentFactory(state -> Mockito.spy(new DeploymentImpl(state)))
-                .setMocks(monitorMocks)
                 .build();
-        monitorMocks.setMock(mock);
 
-        var result = mock.runTestAsync(ctx -> {
-            assertThat(((MockMonitor) mock.monitor).resources)
+        monitorMocks.setStack(() -> {
```
t0yv0, line 111: Lots of places use this pattern of finding resources by type. I think we might use a helper method in the testing framework.

sdk/java/pulumi/src/main/java/com/pulumi/test/Mocks.java
```diff
@@ -1,6 +1,8 @@
-package com.pulumi.deployment;
+package com.pulumi.test;
```
t0yv0, line 1: This is reasonable packaging, thank you.

sdk/java/pulumi/src/main/java/com/pulumi/test/Mocks.java
```diff
@@ -17,16 +18,146 @@ public interface Mocks {
...
+    @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+    class ResourceResult {
+        /**
+         * A resource identifier
+         */
+        public final Optional<String> id;
+        /**
+         * A resource state
+         */
+        public final Object state;
```
t0yv0, line 110: State - is that the right word to describe that this object is? 

Is it really resource outputs?

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
```diff
@@ -278,63 +277,53 @@ public static void myStack(Context ctx) {
...
 
         @Override
-        public CompletableFuture<Tuples.Tuple2<Optional<String>, Object>> newResourceAsync(MockResourceArgs args) {
+        public CompletableFuture<ResourceResult> newResourceAsync(ResourceArgs args) {
             requireNonNull(args.type);
             switch (args.type) {
                 case "aws:ec2/instance:Instance":
                     return CompletableFuture.completedFuture(
-                            Tuples.of(Optional.of("i-1234567890abcdef0"), ImmutableMap.of("publicIp", "203.0.113.12"))
+                            ResourceResult.of(Optional.of("i-1234567890abcdef0"), ImmutableMap.of("publicIp", "203.0.113.12"))
```
t0yv0, line 285: Perhaps a builder for ResourceResult? Builders never end it seems, but it could make this code more readable.

CHANGELOG_PENDING.md
t0yv0: Yes. I'm on the fence here, I guess it is not breaking since we did not sufficiently advertise how to use the old API. So perhaps we just announce this as the new API release.

Suggest something like this.

```
- [sdk] #553: Added `com.pulumi.test` package with new idomatic Java API for [unit testing Pulumi programs]
(https://www.pulumi.com/docs/guides/testing/#unit-testing) against mocks. See ([example](<url-to-example here>)) of  using the new API.

```

t0yv0 (approved):
LGTM overall just had some questions.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/ReadOrRegisterResource.java
pawelprazak: this is internal interface, used for granular dependency management and to allow loose coupling

tests/examples/testing-unit-java/src/test/java/com/pulumi/example/unittest/WebserverInfraTest.java
pawelprazak: the simplest but less clean approach would be to call `run` multiple times like c# does, but JUnit 5 gives this advanced ability to use dynamic tests

sdk/java/pulumi/src/main/java/com/pulumi/test/PulumiTest.java
pawelprazak: yes, the globals :(

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
pawelprazak: my thinking was that a factory method is slightly preferred since there is only 2 args and it forces setting of the required parameter in compile time

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/ReadOrRegisterResource.java
t0yv0: Ah ok then; I don't have such a strong opinion on the internals. If this is not for the user, that's OK. Bear in mind that all ResourceMonitor methods are conceptually coupled, `RegisterResource` and `ReadResource` work against the same state machine, they're not independent, but you probaly know that. Whatever works best.

tests/examples/testing-unit-java/src/test/java/com/pulumi/example/unittest/WebserverInfraTest.java
t0yv0: Slight perference for the simplest apprioach but OK to leave as is since you say this is standard JUnit 5.

sdk/java/pulumi/src/test/java/com/pulumi/deployment/MocksTest.java
t0yv0: It just doesn't name the args, and their type is `Optional<String>` for example, no way to say it's an ID. But the docs probably say that. Just a nit, feel free to ignore.

pawelprazak:
rebased, updated changelog according to feedback

PR #686, opened 2022-06-20
Add replace on change tests

### LINKED ISSUE #82: Cover replaceOnChanges with a test
## Hello!

## Issue details

Resource option `replaceOnChanges` functionality does not seem to be covered by tests, it would be great to add.

We unfortunately do not seem to have good prior art of testing it in pulumi/pulumi.

I found this: sdk/nodejs/tests/runtime/langhost/cases/066.replace_on_changes/index.js

Technically what we're testing here is that:

- the user can express replaceOnChanges option
- and in it propagates as such to the engine

So perhaps we can get a way with a mocks-based test that only exercises the SDK, though a full integration test is also an option here.

pawelprazak:
# Description

- adds a monitor decorator to the internal test api to allow for more invasive testing
- adds a replace-on-change unit test with mocks

Fixes #82 

On top of #683 

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/resources/ResourceTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/resources/ResourceTest.java
@@ -1,44 +1,52 @@
 package com.pulumi.resources;
 
 import com.google.common.collect.ImmutableMap;
-import com.pulumi.Context;
 import com.pulumi.core.internal.Internal;
+import com.pulumi.resources.internal.Stack;
 import com.pulumi.test.Mocks;
 import com.pulumi.test.TestOptions;
 import com.pulumi.test.internal.PulumiTestInternal;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import pulumirpc.Resource.RegisterResourceRequest;
 
-import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
 import static com.pulumi.test.PulumiTest.extractValue;
+import static java.util.Objects.requireNonNull;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.times;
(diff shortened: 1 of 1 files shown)


pawelprazak:
Will be easier to review after #683 is merged.

pawelprazak:
I wanted to write an integration test as well, but I'm not sure how to do it.

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -300,6 +306,19 @@ public Builder deploymentFactory(
             return this;
         }
 
+        /**
+         * Decorate the {@link MockMonitor} instance for this test.
+         *
+         * @param monitorDecorator the decorator, by default {@link Function#identity()}
+         * @return this Pulumi test {@link Builder}
+         */
+        public Builder monitorDecorator(
```
pawelprazak, line 315: allows for adding Mockito.spy or any other decorator for invasive unit tests

sdk/java/pulumi/src/test/java/com/pulumi/resources/ResourceTest.java
```diff
@@ -54,35 +62,61 @@ void testProviderPropagation() {
...
+        assertThat(result.resources())
+                .hasSize(2)
+                .hasExactlyElementsOfTypes(Stack.class, MyResource.class);
+
+        var resourceCaptor = ArgumentCaptor.forClass(Resource.class);
+        var requestCaptor = ArgumentCaptor.forClass(RegisterResourceRequest.class);
+        verify(test.monitor(), times(2)).registerResourceAsync(resourceCaptor.capture(), requestCaptor.capture());
+
+        assertThat(resourceCaptor.getAllValues().get(1)).isInstanceOf(MyResource.class);
+        assertThat(requestCaptor.getAllValues().get(1).getReplaceOnChangesList())
```
t0yv0, line 86: OK interesting. I'm a bit less familiar with the mockist testing but I think this makes sense - this is saying that `monitor` receives a `registerResourceAsync` call with `RegisterResourceRequest` containing the "foo". So essentially it tests that "foo" propagates from `CustomResourceOptions` to `RegisterResourceRequest`. 

This seem reasonable since I guess SDK does not do anything else with the option is just propagates it - we do not need to test that the engine respects it. 

Thank you!

t0yv0 (approved):

PR #705, opened 2022-06-28
Add support for setting java-gen version via EnvVar

stack72:
# Description

Fixes # (issue)

## Checklist


pkg/cmd/pulumi-java-gen/command.go
```diff
@@ -79,8 +79,6 @@ See https://www.pulumi.com/docs/guides/pulumi-packages/schema/#language-specific
 	cmd.Flags().StringVar(&versionArg, "version", "",
 		"semantic version for the generated package")
 
-	contract.AssertNoError(cmd.MarkFlagRequired("version"))
```
t0yv0, line 82: Thinking out loud here. Currently there are resources (plugin json and version.txt) that are generated by pulumi-java-gen, that must be present. So perhaps --version is still required in a sense? Not passing it would generate code that compiles but fails when you use it.

stack72: that would still cause a diff on every checkin - as the build.gradle file would change

stack72: Unless we make it mandatory for those situations AND require it to be an env var for the gradle build template and it will never use the value passed via the env var

stack72: Ok, I have added Viper and allowed the flag to be set via the same env var - so it can either be set via `--version` or `PACKAGE_VERSION` and no functionality will change

t0yv0: Hey @stack72 I was looking at this yesterday and I think the cleanest approach here would be to make sure the versioned plugin.json and version.txt resources are generated by Gradle and not by pulumi-java-gen. That way the codegen and build phases are truly separate, that is if Gradle thinks the version is 1.2.3-alpha then it is *indeed* that in the generated artifact.

```
pulumi-java-gen   # Phase 1: no --version, generates source code
gradle -Pversion=1.2.3 # that's the version
```

I have a PR coming up.

But I can also look at your latest changes maybe that's good enough.

stack72:
Superceded via #709

PR #720, opened 2022-07-05
Support custom jumbo type

### LINKED ISSUE #390: Deserialization should not rely on classes with constructors having large parameter counts
## Hello!

## Issue details

Deserialization should not rely on classes with constructors having large parameter counts.

Code generation for Java SDK classes that need to be deserialized from protobuf currently relies on constructors with many parameters (one per desired property) internally. This breaks down when it hits hard limits on Java parameter counts. 

Earlier when we worked around the problem of not being able to compile providers with large parameter counts, we introduced a special pathway in the SDK code generator for the special case of "jumbo" constructors. This does not completely solve the problem however. While you can construct these Jumbo objects, tryConvertObjectInner errors when facing them.

Desired complete fix would:

- [x] fix the deserializer in the Java SDK to not require one parameter per property in any method or constructor of the generated classes, so as to not hit the Java limit
- [ ] fix the codegen to align with the deserializer expectations
- [ ] remove branching on parameter count from codegen code
- [ ] have no breaking changes for the users (it seems the problem can be solved internally)
- [ ] fix or avoid "`Converter::tryConvertObjectInner` errors when calling jumbo constructors"  found earlier

pawelprazak:
# Description

Fixes #390 

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/JumboCustomType.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/JumboCustomType.java
@@ -1,0 +1,3111 @@
+package com.pulumi.serialization.internal;
+
+import com.pulumi.core.annotations.CustomType;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+import java.util.Optional;
+
+@CustomType
+public final class JumboCustomType {
+    private @Nullable String bar1;
+    private @Nullable String bar10;
+    private @Nullable String bar100;
+    private @Nullable String bar101;
+    private @Nullable String bar102;
+    private @Nullable String bar103;
+    private @Nullable String bar104;
+    private @Nullable String bar105;
+    private @Nullable String bar106;
+    private @Nullable String bar107;
+    private @Nullable String bar108;
+    private @Nullable String bar109;
+    private @Nullable String bar11;
+    private @Nullable String bar110;
+    private @Nullable String bar111;
+    private @Nullable String bar112;
+    private @Nullable String bar113;
+    private @Nullable String bar114;
+    private @Nullable String bar115;
+    private @Nullable String bar116;
+    private @Nullable String bar117;
+    private @Nullable String bar118;
+    private @Nullable String bar119;
+    private @Nullable String bar12;
+    private @Nullable String bar120;
+    private @Nullable String bar121;
+    private @Nullable String bar122;
+    private @Nullable String bar123;
+    private @Nullable String bar124;
+    private @Nullable String bar125;
+    private @Nullable String bar126;
+    private @Nullable String bar127;
+    private @Nullable String bar128;
+    private @Nullable String bar129;
+    private @Nullable String bar13;
+    private @Nullable String bar130;
+    private @Nullable String bar131;
+    private @Nullable String bar132;
+    private @Nullable String bar133;
+    private @Nullable String bar134;
+    private @Nullable String bar135;
+    private @Nullable String bar136;
+    private @Nullable String bar137;
+    private @Nullable String bar138;
+    private @Nullable String bar139;
+    private @Nullable String bar14;
+    private @Nullable String bar140;
+    private @Nullable String bar141;
+    private @Nullable String bar142;
+    private @Nullable String bar143;
+    private @Nullable String bar144;
+    private @Nullable String bar145;
+    private @Nullable String bar146;
+    private @Nullable String bar147;
+    private @Nullable String bar148;
+    private @Nullable String bar149;
+    private @Nullable String bar15;
+    private @Nullable String bar150;
+    private @Nullable String bar151;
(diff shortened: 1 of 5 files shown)


pawelprazak:
Rebased.

WDYT, should I add more tests, or it this good enough?

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/JumboCustomType.java
```diff
@@ -0,0 +1,3111 @@
...
+    	      this.bar93 = defaults.bar93;
+    	      this.bar94 = defaults.bar94;
+    	      this.bar95 = defaults.bar95;
+    	      this.bar96 = defaults.bar96;
+    	      this.bar97 = defaults.bar97;
+    	      this.bar98 = defaults.bar98;
+    	      this.bar99 = defaults.bar99;
+        }
+
+        @CustomType.Setter
```
t0yv0, line 1568: Why does it need both @CustomType.Setter and @CustomType.Parameter - curious?

CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,5 @@
 ### Improvements
 
+- [codegen] #390: support resources with over 255 parameters
```
t0yv0, line 3: Trying to remember if this was unique to resources - I think the issue applies to all generated classes not just resources.

t0yv0:
I would expect to see changes to pkg/codegen/java/gen.go - are those pending or what's the plan here?

- removal of `genJumboOutputType`
- removal of `IsJumbo`

t0yv0:
Java code seems reasonable. We can tune it for performance later if the ser/de ever becomes a problem - I'm guessing reflective invocations are not the fastest.

CHANGELOG_PENDING.md
pawelprazak: The mechanism being changed works only for custom resources specifically. Anything annotated with `@CustomType`.

pawelprazak:
> I would expect to see changes to pkg/codegen/java/gen.go - are those pending or what's the plan here?
> 
> * removal of `genJumboOutputType`
> * removal of `IsJumbo`

I've assumed we don't want to break current providers and I've planned to remove the constructor-based code, after we release both the new codegen and regenerate all of the (affected) providers.

As a result, I plan to do as follows, in 4 stages:
- SDK change (this PR), release the SDK
- update codegen to use the new mechanism
- regenerate and release providers with the new codegen
- remove the old mechanism from codegen and from SDK

pawelprazak:
updated after rebase

CHANGELOG_PENDING.md
t0yv0:
```
/**
 * Annotation used by a Pulumi Cloud Provider Package to mark complex types used for a Resource
 * output property.
 * <p>
 * A complex type must have a single constructor
 * marked with the @see {@link CustomType.Constructor} annotation.
 */
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
public @interface CustomType {
```

Ok. Judging by the comment, the types annotated with this are not modeling Pulumi resources. They are the types used to describe outputs of resources. It sounds like the fix is doing the right thing but the CHANGELOG entry is misleading, that's no problem, I can update later on as it comes time to release.

t0yv0 (approved):

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/JumboCustomType.java
t0yv0: Please be mindful of gigantic generated code sizes that remain a large problem for the speed of Pulumi experience. If something can be solved with fewer API elements we should probably consider that as it puts less pressure on the generated code. I don't have the numbers so perhaps this is negligible, but you could theoretically take into account that setters will only have 1 parameter, and record the name of said parameter in `@CustomType.Setter`. You could also skip recording the name if it happens to match a reflectable name. 

That said I'm good with merging this for now, just consider generated code size when making further changes in the future.

t0yv0:
LGTM but could we please track all the remaining work? In #390 feel free to edit the checklist as needed. Thank you!

sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/JumboCustomType.java
pawelprazak: > You could also skip recording the name if it happens to match a reflectable name.

Unfortunately this is not possible, we change the name of the setter in codegen in some cases.

Yes, you are right, I could reuse the same annotation. I'll make a followup PR with this.

pawelprazak: the follow up: https://github.com/pulumi/pulumi-java/pull/730

PR #724, opened 2022-07-06
Update rewriteconversions api to match branch on pulumi/pulumi

AaronFriel:
See:
- https://github.com/pulumi/pulumi/pull/10047


t0yv0 (approved):
LGTM if tests pass. There was one test failing on `main` that we might need to comment out.

PR #736, opened 2022-07-12
Fix CHANGELOG_PENDING

t0yv0:
# Description

Removed this as nothing changed yet for the users since codegen change needs to be completed:

- [codegen] #390: support resources with over 255 parameters

Removed this since this has not been released in here and ties into an existing feature:

- [pulumi-google-native] #381: all Java packages changed to be more intuitive, unfortunately this is a breaking change

Fixes # (issue)

## Checklist


t0yv0:
FYI @pawelprazak

PR #744, opened 2022-07-14
Custom jumpo type sunset

pawelprazak:
# Description

Use `@CustomType.Builder` in codegen.

Part of #390 

## Checklist

### DIFF
--- a/pkg/codegen/testing/test/testdata/azure-native-nested-types/java/src/main/java/com/pulumi/azurenative/documentdb/outputs/CompositePathResponse.java
+++ b/pkg/codegen/testing/test/testdata/azure-native-nested-types/java/src/main/java/com/pulumi/azurenative/documentdb/outputs/CompositePathResponse.java
@@ -15,21 +15,14 @@ public final class CompositePathResponse {
      * @return Sort order for composite paths.
      * 
      */
-    private final @Nullable String order;
+    private @Nullable String order;
     /**
      * @return The path for which the indexing behavior applies to. Index paths typically start with root and end with wildcard (/path/*)
      * 
      */
-    private final @Nullable String path;
+    private @Nullable String path;
 
-    @CustomType.Constructor
-    private CompositePathResponse(
-        @CustomType.Parameter("order") @Nullable String order,
-        @CustomType.Parameter("path") @Nullable String path) {
-        this.order = order;
-        this.path = path;
-    }
-
+    private CompositePathResponse() {}
     /**
      * @return Sort order for composite paths.
      * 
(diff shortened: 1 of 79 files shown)


pkg/codegen/java/gen.go
```diff
@@ -365,31 +365,6 @@ func (mod *modContext) typeStringForEnumType(enumType *schema.EnumType) TypeShap
 	return TypeShape{Type: fqn}
 }
 
-// Returns a constructor for an empty instance of type `t`.
-//
-// optionalAsNull is used in conjunction with the `typeString` parameter
-// `outerOptional` to ensure types line up.
-// In general, `outerOptional` <=> `!optionalAsNull`.
-func emptyTypeInitializer(ctx *classFileContext, t schema.Type, optionalAsNull bool) string {
```
t0yv0, line 373: This seems like a win to me! I had doubts this is fully corret in all cases.

pkg/codegen/testing/test/testdata/azure-native-nested-types/java/src/main/java/com/pulumi/azurenative/documentdb/outputs/CompositePathResponse.java
```diff
@@ -52,30 +45,32 @@ public static Builder builder() {
...
             return this;
         }
+        @CustomType.Setter
         public Builder path(@Nullable String path) {
             this.path = path;
             return this;
-        }        public CompositePathResponse build() {
-            return new CompositePathResponse(order, path);
+        }
+        public CompositePathResponse build() {
```
t0yv0, line 69: Looking good! So this removed the situation with "too many constructors". 

Observing that since class fields are no longer final, it's possible to shorten the builder code further by keeping an instance of the class.

```
    @CustomType.Builder
    public static final class Builder {
        private @Nullable CompositePathResponse $;
    }
    @CustomType.Setter
    public Builder path(@Nullable String path) {
         $.path = path;
         return this;
    }
    public CompositePathResponse build() {
        return $;
    }
```

pkg/codegen/testing/test/testdata/jumbo-resources/schema.json
```diff
@@ -773,6 +773,9 @@
         },
         "bar256": {
           "type": "string"
+        },
+        "default": {
```
t0yv0, line 777: Curious why add this one? To test name mangling?

pkg/codegen/testing/test/testdata/azure-native-nested-types/java/src/main/java/com/pulumi/azurenative/documentdb/outputs/CompositePathResponse.java
pawelprazak: yes, correct, I've noticed it too, just didn't want to make unrelated changes

pkg/codegen/testing/test/testdata/jumbo-resources/schema.json
pawelprazak: yes, I wanted to test the `default` -> `default_` and that one or the other will appear in all the correct places

Issue #749, opened 2022-07-19
Schema Package Overrides cause implementation dependecies and build failure


stack72:
The package overrides for kubernetes - https://github.com/pulumi/pulumi-kubernetes/pull/2055 are showing as implementation dependencies in the build.gradle file in pulumi-java-gen 5.0.0

```
dependencies {
    implementation("admissionregistration.k8s.io/v1:admissionregistration.v1")
    implementation("admissionregistration.k8s.io/v1beta1:admissionregistration.v1beta1")
    implementation("apiextensions.k8s.io/v1:apiextensions.v1")
    implementation("apiextensions.k8s.io/v1beta1:apiextensions.v1beta1")
    implementation("apiregistration.k8s.io/v1:apiregistration.v1")
    implementation("apiregistration.k8s.io/v1beta1:apiregistration.v1beta1")
    implementation("apps/v1:apps.v1")
    implementation("apps/v1beta1:apps.v1beta1")
    implementation("apps/v1beta2:apps.v1beta2")
    implementation("auditregistration.k8s.io/v1alpha1:auditregistration.v1alpha1")
    implementation("authentication.k8s.io/v1:authentication.v1")
    implementation("authentication.k8s.io/v1beta1:authentication.v1beta1")
    implementation("authorization.k8s.io/v1:authorization.v1")
```

This is causing build failures as:

```
make java_sdk
/Users/stack72/code/go/src/github.com/pulumi/pulumi-kubernetes/bin/pulumi-java-gen generate --schema provider/cmd/pulumi-resource-kubernetes/schema.json --out sdk/java --build gradle-nexus
cd sdk/java/ && \
		echo "module fake_java_module // Exclude this directory from Go tools\n\ngo 1.17" > go.mod && \
		gradle --console=plain build
> Task :compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find admissionregistration.k8s.io/v1:admissionregistration.v1:.
     Required by:
         project :
   > Could not find admissionregistration.k8s.io/v1beta1:admissionregistration.v1beta1:.
     Required by:
         project :
   > Could not find apiextensions.k8s.io/v1:apiextensions.v1:.
     Required by:
         project :
   > Could not find apiextensions.k8s.io/v1beta1:apiextensions.v1beta1:.
     Required by:
         project :
```

[...]

PR #751, opened 2022-07-19
Fix #739 - disambiguate Packages and Dependencies

### LINKED ISSUE #739: Confused meaning of "Packages" field in .languages.java options
### What happened?

We accidentally overloaded two meanings on the `packages` field in the language specific configuration for the schemas, see: 

```
// PackageInfo represents a Java language-specific info for a package.
type PackageInfo struct {
	Packages    map[string]string `json:"packages,omitempty"`
	BasePackage string            `json:"basePackage"`

	// If set to "gradle" generates a basic set of Gradle build files.
	BuildFiles string `json:"buildFiles"`
}
```

One use case is https://github.com/pulumi/pulumi-kubernetes/pull/2055/files#diff-8dfd74ae4dfdfd04f5c226c8e65a2d3060d6761b45252ba7a757c22492d99311R206 controlling friendly names for the generated versioned packages.

```
            "packages": {
                "admissionregistration.k8s.io/v1": "admissionregistration.v1",
            }
```

Another use case is specifying dependencies on other packages:

https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/templates_gradle.go#L75

You might want to say here:

```
            "packages": {
                "com.pulumi:azuread": "5.22.0",
            }
```

And have `pulumi-java-gen` generate a build file that includes a dependency on that package.

We should disambiguate this and add doc strings to explain what the options do.

I think the right thing to do is keep packages for the original use case, and fixup the version use case to use a different field.

### Steps to reproduce

-

### Expected Behavior

-

### Actual Behavior

-

### Versions used

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
# Description

Fixes #739, #749

## Checklist

### DIFF
--- a/pkg/codegen/java/templates_gradle_test.go
+++ b/pkg/codegen/java/templates_gradle_test.go
@@ -27,7 +27,7 @@ func TestNewGradleTemplateContext(t *testing.T) {
 	assert.Equal(t, "support@pulumi.com", tctx.DeveloperEmail)
 	assert.Equal(t, "The Apache License, Version 2.0", tctx.LicenceName)
 	assert.Equal(t, "http://www.apache.org/licenses/LICENSE-2.0.txt", tctx.LicenceURL)
-	assert.Equal(t, info.Packages, tctx.Dependencies)
+	assert.Equal(t, info.Dependencies, tctx.Dependencies)
 }
 
 func TestGenGradleProject(t *testing.T) {
@@ -54,6 +54,6 @@ func eksExample() (*schema.Package, *PackageInfo) {
 		"com.pulumi:aws":        "5.4.0",
 		"com.pulumi:kubernetes": "3.19.1",
 	}
-	info := &PackageInfo{Packages: deps}
+	info := &PackageInfo{Dependencies: deps}
 	return pkg, info
 }


stack72 (approved):
This is way better :) Thank you!!

pkg/codegen/java/importer.go
```diff
@@ -21,56 +21,11 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 )
 
-const defaultBasePackage = "com.pulumi."
```
pawelprazak, line 24: nice!

pkg/codegen/java/package_info.go
```diff
@@ -0,0 +1,128 @@
...
+// These options affect how GeneratePackage generates and lays out
+// Java SDKs for Pulumi Packages.
+//
+// The options are typically set in the .language.java section of
+// the Schema, but can also be overridden during code generation:
+//
+//     pulumi-java-gen generate --override overrides.json
+//
+// See https://www.pulumi.com/docs/guides/pulumi-packages/schema/#language-specific-extensions
+type PackageInfo struct {
```
pawelprazak, line 31: thank you for documenting this so well :)

pkg/codegen/java/gen_test.go
```diff
@@ -124,9 +124,8 @@ func TestGeneratePackage(t *testing.T) {
 func generatePackage(tool string, pkg *schema.Package, extraFiles map[string][]byte) (map[string][]byte, error) {
 	pkgInfo := PackageInfo{
 		BuildFiles: "gradle",
-		Packages: map[string]string{
+		Dependencies: map[string]string{
```
pawelprazak, line 127: I like this new name

pawelprazak:
Thank you for fixing this so fast! :)

PR #759, opened 2022-07-20
String based type reification mechanism

### LINKED ISSUE #629: Serialization error when using helm to install a chart with Java
### What happened?

Error when using helm to install a chart with Java.

### Steps to reproduce

```java
return new Release("nginx",
                ReleaseArgs.builder()
                        .chart("nginx-ingress")
                        .version("1.24.4")
                        .repositoryOpts(RepositoryOptsArgs.builder().repo("https://charts.helm.sh/stable").build())
                        .valueYamlFiles(new FileAsset("./values.yaml"))
                        .build(),
                CustomResourceOptions.builder()
                        .provider(provider)
                        .build());
```

### Expected Behavior

Working code.

### Actual Behavior

```
java.lang.UnsupportedOperationException: Convert [com.pulumi.kubernetes.helm.sh_v3.Release.resourceNames]: Error converting 'com.google.common.collect.RegularImmutableMap' to 'TypeShape{type=interface java.util.Map, parameters=[TypeShape{type=class java.lang.String, parameters=[]}, TypeShape{type=interface java.util.List, parameters=[]}]}'. Expected a parameter type for the List, got none
```

### Versions used

_No response_

### Additional context

_No response_

### Contributing

pawelprazak:
# Description

Fixes #629

## Checklist


sdk/java/pulumi/src/test/java/com/pulumi/core/internal/ImportExportMetadataTest.java
```diff
@@ -20,22 +20,22 @@ class ImportExportMetadataTest {
     @SuppressWarnings("unused")
     public static class Tester {
 
-        @Export(name = "complex1", type = Either.class, parameters = {Integer.class, String.class})
+        @Export(name = "complex1", typeString = "com.pulumi.core.Either<java.lang.Integer,java.lang.String>")
         public final Output<Either<Integer, String>> complex1 = Output.of(Either.ofLeft(1));
 
-        @Export(name = "complex2", type = Either.class, parameters = {Integer.class, String.class})
+        @Export(name = "complex2", typeString = "com.pulumi.core.Either<java.lang.Integer,java.lang.String>")
```
t0yv0, line 26: Hm. 

This looks workable, but it's worth thinking this through since this goes into providers and will be a long-term compatibility commitment. 

How about we ditch string representation and instead use `String.class` style references as before?

Using JSON notation so we do not need our own tree parser:

```
     Either<Map<String, List<String>>, Map<String, List<String>>>
     @Export(type="[0,[1,2,[3,2]],[1,2,[3,2]]]", refs=[Either.class, Map.class, String.class, List.class]

     Either<Integer, String>
     @Export(type="[0,1,2]", refs=[Either.class, String.class, Integer.class])
```

If prefer a bit more readable representation, go ahead and use our own tree parser:

```
     Either<Integer, String>
     @Export(type="0<1,2>", refs=[Either.class, String.class, Integer.class])
```

The more important aspect to me is if we could try and get to write `Either.class` instead of "com.pulumi.core.Either". This reuses JVM metadata handling and we don't need to invent class cache etc.

Whether we have a parser for our own format or no is less important - just need to review+test the parser thoroughy :)

pawelprazak: OK, this proposal looks more optimized and also more complex.

I'd like to kindly remind you @t0yv0 the string based representation was your idea.

pawelprazak: IMHO this boils down to where we move the complexity:
- string based moves the complexity to the SDK
- JSON+.class moves the complexity to the codegen

I like the idea of using .class, because we'd retain compile time check vs runtime lookup of classes based on strings.
On the other hand I like that with string based the modifications to the codegen are minimal.

t0yv0: Decided on a sync to go ahead and move complexity to SDK 🙏 ping me again when ready to review (no rush).

pawelprazak:
reworked, will add more test cases later

sdk/java/pulumi/src/test/java/com/pulumi/core/TypeShapeTest.java
```diff
@@ -0,0 +1,58 @@
...
+                        "com.pulumi.core.Either<java.lang.Integer,java.lang.String>"
+                ),
+                arguments(
+                        "[0,[0,[0,1]]]", new Class<?>[]{List.class, String.class},
+                        "java.util.List<java.util.List<java.util.List<java.lang.String>>>"
+                ),
+                arguments(
+                        "[0,[1,2,[3,2]],[1,[3,2],2]]",
+                        new Class<?>[]{Either.class, Map.class, String.class, List.class},
+                        "com.pulumi.core.Either<java.util.Map<java.lang.String,java.util.List<java.lang.String>>,java.util.Map<java.util.List<java.lang.String>,java.lang.String>>"
```
t0yv0, line 47: Neat!

pkg/codegen/java/typeshape_test.go
```diff
@@ -0,0 +1,105 @@
...
+					}},
+					{Type: names.Map, Parameters: []TypeShape{
+						{Type: names.List, Parameters: []TypeShape{{Type: names.String}}},
+						{Type: names.String},
+					}},
+				},
+			},
+			refs:    "Either.class,Map.class,String.class,List.class",
+			tree:    "[0,[1,2,[3,2]],[1,[3,2],2]]",
+			imports: "import com.pulumi.core.Either;\nimport java.lang.String;\nimport java.util.List;\nimport java.util.Map;",
```
t0yv0, line 90: Neat!

t0yv0 (approved):
Superb. Let's ship this. Apologies for very long review delay.

PR #764, opened 2022-07-26
Add rpc support for About and GetProgramDependencies

Frassle:
Update to the latest pulumi/pulumi code and copying over the current java logic from cmd/about.go to here to live in the language plugin.

### DIFF
--- a/pkg/cmd/pulumi-language-java/main.go
+++ b/pkg/cmd/pulumi-language-java/main.go
@@ -394,7 +394,8 @@ func (host *javaLanguageHost) InstallDependencies(req *pulumirpc.InstallDependen
 	return nil
 }
 
-func (host *javaLanguageHost) GetProgramDependencies(ctx context.Context, req *pulumirpc.GetProgramDependenciesRequest) (*pulumirpc.GetProgramDependenciesResponse, error) {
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -391,3 +393,59 @@ func (host *javaLanguageHost) InstallDependencies(req *pulumirpc.InstallDependen
 	logging.V(5).Infof("InstallDependencies(Directory=%s): done", req.Directory)
 	return nil
 }
+
+func (host *javaLanguageHost) GetProgramDependencies(
+	ctx context.Context, req *pulumirpc.GetProgramDependenciesRequest) (*pulumirpc.GetProgramDependenciesResponse, error) {
+	// TODO: Implement dependency fetcher for Java
```
pawelprazak, line 399: note: might be worth to create a ticket for this (if not present)

I think implementation in `GetRequiredPlugins` could be reused here.

pawelprazak (approved):
LGTM, thank you!

pkg/cmd/pulumi-language-java/main.go
Frassle: https://github.com/pulumi/pulumi-java/issues/770

PR #766, opened 2022-07-27
Cleanup Resource getters

### LINKED ISSUE #704: Change getId() to id()
## Hello!

## Issue details

We renamed all property getters to remove `get` from the name, but forgot to do the same for the `getId()` method: https://github.com/pulumi/pulumi-java/blob/5cd9786feb4ce575431baa81e92e55d0bd9b58f7/sdk/java/pulumi/src/main/java/com/pulumi/resources/CustomResource.java#L94-L96

We should go ahead and do that. It's a breaking change but we now is the best time we can do it yet.

pawelprazak:
# Description

Rename getters on Resource

- remove get prefix
- update codegen
- mark old getters as deprecated, do not delete

Fixes #704 

On top of #789

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReference.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReference.java
@@ -22,23 +22,14 @@ import java.util.concurrent.CompletableFuture;
  */
 public class StackReference extends CustomResource {
 
-    /**
-     * The name of the referenced stack.
-     */
     @Export(name = "name", type = String.class)
     private Output<String> name;
 
-    /**
-     * The outputs of the referenced stack.
-     */
     @Export(name = "outputs", type = Map.class, parameters = {String.class, Object.class})
-    public Output<Map<String, Object>> outputs;
+    private Output<Map<String, Object>> outputs;
 
-    /**
-     * The names of any stack outputs which contain secrets.
-     */
     @Export(name = "secretOutputNames", type = List.class, parameters = String.class)
-    public Output<List<String>> secretOutputNames;
+    private Output<List<String>> secretOutputNames;
 
     /**
      * Create a {@link StackReference} resource with the given unique name.
@@ -90,15 +81,63 @@ public class StackReference extends CustomResource {
         return args == null ? Output.of(name) : args.name().orElse(Output.of(name));
     }
 
+    /**
+     * The name of the referenced stack.
+     *
+     * @return the stack name
+     * @deprecated use {@link #name()}
+     */
+    @Deprecated
     public Output<String> getName() {
+        return name();
+    }
+
+    /**
+     * The name of the referenced stack.
+     *
+     * @return the stack name
+     */
+    public Output<String> name() {
         return name;
     }
 
+    /**
+     * The outputs of the referenced stack.
+     *
+     * @return the stack outputs
+     * @deprecated use {@link #outputs()}
+     */
+    @Deprecated
     public Output<Map<String, Object>> getOutputs() {
+        return outputs();
+    }
+
+    /**
+     * The outputs of the referenced stack.
+     *
+     * @return the stack outputs
+     */
+    public Output<Map<String, Object>> outputs() {
         return outputs.applyValue(ImmutableMap::copyOf);
     }
 
+    /**
+     * The secret output names of the referenced stack.
+     *
+     * @return the names of the secret outputs
+     * @deprecated use {@link #secretOutputNames()}
+     */
+    @Deprecated
     public Output<List<String>> getSecretOutputNames() {
+        return secretOutputNames();
+    }
+
+    /**
+     * The secret output names of the referenced stack.
+     *
+     * @return the names of the secret outputs
+     */
+    public Output<List<String>> secretOutputNames() {
         return secretOutputNames.applyValue(ImmutableList::copyOf);
     }
 
@@ -108,9 +147,11 @@ public class StackReference extends CustomResource {
      *
      * @param name The name of the stack output to fetch.
      * @return An {@link Output} containing the requested value.
+     * @deprecated use {@link #output(String)}
      */
+    @Deprecated
     public Output<?> getOutput(String name) {
-        return this.getOutput(Output.of(name));
+        return output(name);
     }
 
     /**
@@ -120,7 +161,31 @@ public class StackReference extends CustomResource {
      * @param name The name of the stack output to fetch.
      * @return An {@link Output} containing the requested value.
      */
+    public Output<?> output(String name) {
+        return output(Output.of(name));
+    }
+
+    /**
+     * Fetches the value of the named stack output, or {@code null} if the stack output was not found.
+     * <p>
+     *
+     * @param name The name of the stack output to fetch.
+     * @return An {@link Output} containing the requested value.
+     * @deprecated use {@link #output(Output)}
+     */
+    @Deprecated
     public Output<?> getOutput(Output<String> name) {
+        return output(name);
+    }
+
+    /**
+     * Fetches the value of the named stack output, or {@code null} if the stack output was not found.
+     * <p>
+     *
+     * @param name The name of the stack output to fetch.
+     * @return An {@link Output} containing the requested value.
+     */
+    public Output<?> output(Output<String> name) {
         // Note that this is subtly different from "apply" here. A default "apply" will set the secret bit if any
         // of the inputs are a secret, and this.outputs is always a secret if it contains any secrets.
         // We do this dance, so we can ensure that the Output we return is not needlessly tainted as a secret.
@@ -165,6 +230,7 @@ public class StackReference extends CustomResource {
      *
      * @param name The name of the stack output to fetch.
      * @return The value of the referenced stack output.
+     * @throws UnsupportedOperationException if the output is secret
      */
     public CompletableFuture<?> getValueAsync(String name) {
         return this.getValueAsync(Output.of(name));
@@ -179,9 +245,10 @@ public class StackReference extends CustomResource {
      *
      * @param name The name of the stack output to fetch.
      * @return The value of the referenced stack output.
+     * @throws UnsupportedOperationException if the output is secret
      */
     public CompletableFuture<?> getValueAsync(Output<String> name) {
-        return Internal.of(this.getOutput(name)).getDataAsync()
+        return Internal.of(this.output(name)).getDataAsync()
                 .thenApply(data -> {
                     if (data.isSecret()) {
                         throw new UnsupportedOperationException(
--- a/sdk/java/pulumi/src/test/java/com/pulumi/resources/internal/StackTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/resources/internal/StackTest.java
@@ -126,7 +126,7 @@ class StackTest {
         });
 
         var ref = extractValue(result.output("ref", StackReference.class));
-        assertThat(extractValue(ref.getOutput("ref"))).isNotNull();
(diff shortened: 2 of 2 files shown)


pkg/codegen/testing/test/testdata/mini-azurenative/java/src/main/java/com/pulumi/azurenative/alertsmanagement/outputs/GetActionRuleByNameResult.java
```diff
@@ -51,7 +51,7 @@ private GetActionRuleByNameResult() {}
      * @return Azure resource Id
      * 
      */
-    public String id() {
+    public String id_() {
```
t0yv0, line 54: Uhm this seems like a breaking change that's really undesirable? Can this work be accomplished without renaming id to id_ in providers in case some examples/ code out there already is linking to id()?

t0yv0: GetActionRuleByNameResult is not a resource. It seems like isReservedResourceMethod is perhaps mis-applied.

t0yv0:
The implementation is good I just worry about the projection of breaking changes via codegen to providers.



```
ls ~/schemas
aws-native-v0.12.0.json                 pulumi-azure-native-v1.56.0-full.json   pulumi-azuread-v5.19.0.json
awsx.json                               pulumi-azure-native-v1.56.0.json        pulumi-eks-v0.37.1.json
pulumi-aws-v4.37.3.json                 pulumi-azure-native-v1.63.0.json        pulumi-kubernetes-v3.15.1.json
```

I have this mini-corpus of schemata. Now it's possible to count how many times say resourceType is used:

```
jq -c '.resources[]|.properties.resourceType' *.json |grep -v null | wc -l                                                                   ~/schemas
      47
```

Thankfully the count for "id" is 0 which I was most worried about.

t0yv0:
```

Resource properties found across Pulumi schemas:
   
 | id             |  0 |
 | urn            |  2 |
 | resourceType   | 37 |
 | resourceName   | 16 |
 | childResources |  0 |
 
```

t0yv0:
Generally, the way we inherit (Resource -> CustomResource -> MyCloudResource), makes for collisions. Any method on superclass may collide with a property from the schema causing a rename. This particular change is relatively contained (a few crazy URN collisions and resourceType, resourceName collisions that will silently start retuning the wrong data after the change), so maybe we can take it as is.

I'm almost wondering if we should be using some prefix in Resource instead to minimize collision probability, like this:

```
pulumiId()
pulumiURN()
pulumiResourceType()
```

Most of the time the user wants to extract the resource properties, not these pulumi specific properties right?

But this is strictly worse than `getId()` etc.

Or dotted accessors like this?

```
foo.pulumi().id(), foo.pulumi().urn() ?  
```

Any other tricks  we should consider?

Copying @mikhailshilkov for ideas as you've requested the original ticket.

mikhailshilkov:
@t0yv0 We must have exactly the same problem in all other languages, right? Or it sounds like `resourceType` and `resourceName` may be unique?

Overall, I don't think we need to do anything special for Java compared to other languages.

t0yv0:
There's some slight differences - inheritance rules differ across languages, as well as naming choices. I recall in C# we had something similar with a deep bug around URN property being overridden by an "urn" property from the schema.

mikhailshilkov:
I think `id` is somewhat unique here: I expect it to be used much more often than the other four properties you have above, and also it's less Pulumi-specific and more cloud-related (users link resources with it). Also, C# has exactly the same problem with `Id` name clashes, which may be the cause for zero collisions in your data set.

If we don't think it's too weird, we can change only `id` and keep getters for other properties like `getResourceType`, at least for now.

pawelprazak:
IMO the easiest solution is to use `id_` and `urn_` in the SDK

pawelprazak:
> If we don't think it's too weird, we can change only `id` and keep getters for other properties like `getResourceType`, at least for now.

IMHO it is quite weird, I'd rather strive for consistency.

pkg/codegen/testing/test/testdata/mini-azurenative/java/src/main/java/com/pulumi/azurenative/alertsmanagement/outputs/GetActionRuleByNameResult.java
pawelprazak: > GetActionRuleByNameResult is not a resource. It seems like isReservedResourceMethod is perhaps mis-applied.

or miss-named, I'm not sure

pawelprazak:
I'm also ok with prefixing getters with `pulumi`, eg `pulumiId`.

The `.pulumi().id()` would probably be cleaner, but less discoverable from code completion and not consistent with other SDKs.

pawelprazak:
I've created new issue for the resource vs non-resource problem https://github.com/pulumi/pulumi-java/issues/787

pkg/codegen/testing/test/testdata/mini-azurenative/java/src/main/java/com/pulumi/azurenative/alertsmanagement/outputs/GetActionRuleByNameResult.java
pawelprazak: rebased on fix for this

t0yv0:
One final round on this, apologies for meandering around, but we will need to live with these decisions for a long time.

We've extended the quick impact analysis across major provider schemas and cross-compared to how C# does it.

```

"urn" renamed to "urn_":

kubernetes:helm.sh/v2:Chart
kubernetes:helm.sh/v3:Chart


"resourceName" renamed to "resourceName_":

aws:servicecatalog/tagOptionResourceAssociation:TagOptionResourceAssociation
azure-native:aadiam/v20200301:privateLinkForAzureAd
azure-native:aadiam/v20200301preview:privateLinkForAzureAd
azure-native:aadiam:privateLinkForAzureAd
azure-native:databoxedge/v20200901:ArcAddon
azure-native:databoxedge/v20200901preview:ArcAddon
azure-native:databoxedge/v20201201:ArcAddon
azure-native:databoxedge/v20210201:ArcAddon
azure-native:databoxedge/v20210201preview:ArcAddon
azure-native:databoxedge/v20210601:ArcAddon
azure-native:databoxedge/v20210601preview:ArcAddon
azure-native:databoxedge/v20220301:ArcAddon
azure-native:databoxedge/v20220401preview:ArcAddon
azure-native:databoxedge:ArcAddon
google-native:cloudbuild/v1:Trigger


"resourceType" renamed to "resourceType_":

aws-native:ec2:FlowLog
aws-native:ec2:TransitGatewayMulticastDomainAssociation
aws-native:ec2:TransitGatewayMulticastGroupMember
aws-native:ec2:TransitGatewayMulticastGroupSource
aws-native:fms:Policy
aws-native:lightsail:Disk
aws-native:lightsail:Instance
aws-native:servicecatalogappregistry:ResourceAssociation
aws:cfg/remediationConfiguration:RemediationConfiguration
aws:ec2/vpcIpamPoolCidrAllocation:VpcIpamPoolCidrAllocation
aws:ec2transitgateway/routeTableAssociation:RouteTableAssociation
aws:ec2transitgateway/routeTablePropagation:RouteTablePropagation
aws:fms/policy:Policy
aws:lightsail/containerService:ContainerService
aws:shield/protectionGroup:ProtectionGroup
aws:ssm/maintenanceWindowTarget:MaintenanceWindowTarget
azure-native:web/v20150801:SiteRelayServiceConnection
azure-native:web/v20150801:SiteRelayServiceConnectionSlot
azure-native:web/v20160801:WebAppRelayServiceConnection
azure-native:web/v20160801:WebAppRelayServiceConnectionSlot
azure-native:web/v20180201:WebAppRelayServiceConnection
azure-native:web/v20180201:WebAppRelayServiceConnectionSlot
azure-native:web/v20181101:WebAppRelayServiceConnection
azure-native:web/v20181101:WebAppRelayServiceConnectionSlot
azure-native:web/v20190801:WebAppRelayServiceConnection
azure-native:web/v20190801:WebAppRelayServiceConnectionSlot
azure-native:web/v20200601:WebAppRelayServiceConnection
azure-native:web/v20200601:WebAppRelayServiceConnectionSlot
azure-native:web/v20200901:WebAppRelayServiceConnection
azure-native:web/v20200901:WebAppRelayServiceConnectionSlot
azure-native:web/v20201001:WebAppRelayServiceConnection
azure-native:web/v20201001:WebAppRelayServiceConnectionSlot
azure-native:web/v20201201:WebAppRelayServiceConnection
azure-native:web/v20201201:WebAppRelayServiceConnectionSlot
azure-native:web/v20210101:WebAppRelayServiceConnection
azure-native:web/v20210101:WebAppRelayServiceConnectionSlot
azure-native:web/v20210115:WebAppRelayServiceConnection
azure-native:web/v20210115:WebAppRelayServiceConnectionSlot
azure-native:web/v20210201:WebAppRelayServiceConnection
azure-native:web/v20210201:WebAppRelayServiceConnectionSlot
azure-native:web/v20210301:WebAppRelayServiceConnection
azure-native:web/v20210301:WebAppRelayServiceConnectionSlot
azure-native:web/v20220301:WebAppRelayServiceConnection
azure-native:web/v20220301:WebAppRelayServiceConnectionSlot
azure-native:web:WebAppRelayServiceConnection
azure-native:web:WebAppRelayServiceConnectionSlot
azure:securitycenter/subscriptionPricing:SubscriptionPricing
google-native:apigee/v1:Reference
```

Helm Charts are not yet in the Java pkg, so we are not to worry about it.

Examples of what C# does:

https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Resources/Resource.cs#L82
https://github.com/pulumi/pulumi-google-native/blob/master/sdk/dotnet/Apigee/V1/Reference.cs#L46

Based on this, I'd recommended way forward is as follows. New Resource/CustomResource methods:

```
id() # because there are no conflicts in practice over the ID prop, and C# does it also
childResources() # no conflicts, non-breaking, C# does it also
urn() # currently not-breaking; when Charts make it into Java they will have urn_; conflicts rare
```

Leave as is, adding a doc comment that "get" prefix is retained to avoid name conflicts with custom resource properties (similar comment as in C#):

```
getType()
getName()
```

t0yv0:
Noting another design round after meeting with @pawelprazak. We decided to:

1. Keep "natural" forms for "id()" ad "urn()" to match C# and also understanding that those will be likely used by users

2. For the rest, use pulumi prefix, so "pulumiChildResources()", "pulumiResourceType()", and "pulumiResourceName()"; we understand these don't get used by users that often, they need some form of disambiguation to avoid conflicts, and using `get` prefix is undesirable as it's a stylistic problem with the rest of the code.

pawelprazak:
addresseed feedback, rebased, updated changelog

sdk/java/pulumi/src/main/java/com/pulumi/resources/StackReference.java
```diff
@@ -22,23 +22,14 @@
...
-     */
     @Export(name = "name", type = String.class)
     private Output<String> name;
 
-    /**
-     * The outputs of the referenced stack.
-     */
     @Export(name = "outputs", type = Map.class, parameters = {String.class, Object.class})
-    public Output<Map<String, Object>> outputs;
+    private Output<Map<String, Object>> outputs;
```
t0yv0, line 29: Interesting, this may be a minor breaking change but hopefully we can get away with this for now.

t0yv0 (approved):

Issue #702, opened 2022-06-28
`pulumi convert` doesn't fully work when targeting Java


mikhailshilkov:
### What happened?

`pulumi convert --language java --out java` doesn't quite work: it generates a `java` file but it doesn't set up the project file(s). The command tries building the generated code but fails with 

```
error: Failed to configure executor, tried: jar, maven, gradle, jbang

error: failed to load language plugin java: could not read plugin [/usr/local/bin/pulumi-language-java] stdout: EOF
```

which is fairly confusing.

### Steps to reproduce

Create a `Pulumi.yaml` program like

```yaml
name: yamyam
runtime: yaml
resources:
  echo:
    type: command:local:Command
    properties:
      create: echo ''
```

then run `pulumi convert --language java --out java`.

### Expected Behavior

A Java project is created and built successfully.

### Actual Behavior

```
error: Failed to configure executor, tried: jar, maven, gradle, jbang

error: failed to load language plugin java: could not read plugin [/usr/local/bin/pulumi-language-java] stdout: EOF
```

### Versions used

```
CLI          
Version      3.35.1
Go Version   go1.18.3
Go Compiler  gc

Plugins
NAME  VERSION
yaml  unknown

Host     
OS       darwin
Version  12.3.1
Arch     x86_64

This project is written in yaml

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/mikhailshilkov
User           mikhailshilkov
Organizations  mikhailshilkov, pulumi

Pulumi locates its logs in /var/folders/3_/sqqfv27j1h98c18tllqlbtyh0000gn/T/ by default
warning: Failed to get information about the current stack: No current stack
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #771, opened 2022-07-28
Fix import path for provider resources on `pulumi convert`

aq17:
# Description

Fixes `pu/pu` [#283](https://github.com/pulumi/pulumi-yaml/issues/283) for Java

## Checklist

### DIFF
--- a/pkg/codegen/testing/test/testdata/kubernetes-template-pp/java/kubernetes-template.java
+++ b/pkg/codegen/testing/test/testdata/kubernetes-template-pp/java/kubernetes-template.java
@@ -27,7 +27,7 @@ public class App {
         var provider = new Provider("provider", ProviderArgs.builder()        
             .enableServerSideApply(true)
             .build());
-			
+
         var argocd_serverDeployment = new Deployment("argocd_serverDeployment", DeploymentArgs.builder()        
             .apiVersion("apps/v1")
             .kind("Deployment")


aq17:
@Zaid-Ajaj @t0yv0 I'm having trouble testing this ([Levi's comment with current invalid output](https://github.com/pulumi/pulumi/pull/10150)) due to [#702](https://github.com/pulumi/pulumi-java/issues/702) still open

t0yv0:
We should fix that. Can mark current one as blocked for now, pointing to
the blocker.

myhau:
@t0yv0 @aq17 I've opened a draft PR with the fix for #702 here: https://github.com/pulumi/pulumi-java/pull/773. I need some early feedback before finishing up the work (see questions and todo sections in the PR description).

t0yv0:
This should be unblocked now @aq17 🙂 

Could we add a quick test?

AaronFriel (commented):
It might be simpler to remove the special casing in this PR, and use the `pcl.MapProvidersAsResources` trick, which simplifies codegen by removing a number of special cases. It remaps the token `pulumi:providers:foo` to `foo::Provider`, which enables existing codegen to work without special casing providers:

https://github.com/pulumi/pulumi/blob/4b6d3521407764305ca46dc18d06707930e318e1/pkg/codegen/pcl/utilities.go#L139-L144

In Python, C#, TS, and Go codegen, we call that at the beginning of GenerateProgram like so:

https://github.com/pulumi/pulumi/blob/4b6d3521407764305ca46dc18d06707930e318e1/pkg/codegen/go/gen_program.go#L56-L59

AaronFriel:
Much smaller diff! 🎉

Zaid-Ajaj (approved):
Looks great!

t0yv0:
Merging, thanks team! :shipit:

PR #773, opened 2022-08-02
Fix `pulumi convert` when targeting Java (by generating maven project structure)

### LINKED ISSUE #702: `pulumi convert` doesn't fully work when targeting Java

myhau:
# Description

(In progress, but needs some early feedback)

Fixes #702

This fixes `pulumi convert --language java`. This is done by creating a basic Maven project configuration:

- a basic `pom.xml` file with dependencies
- a directory structure for the generated Java file

I've tested this locally by building `pulumi` and `pulumi-java` (with bin tools) and converting some basic project.

**It's not perfect**, but fixes the issue at hand. Some other ideas I had (require more work and refactoring for other languages):

- detect or ask for the type of the executor - "Do you want to use gradle or maven?"
- use templates (e.g. [`java` from `pulumi/templates`](https://github.com/pulumi/templates/tree/master/java)) to generate the basic project configuration (instead of hardcoded `pom.xml`)

## Questions / problems

- Is maven a good default?
- Do you think that it is ok to hardcode `pom.xml` contents here? (I'm especially concerned with maven plugins versions, like `maven-assembly-plugin`,  but they do not seem to be updated on a regular basis in `pulumi/templates` repo either)
- I could not find any integration tests for `pulumi convert`. Are there any? Should I write some?

## To do

- [ ] try to use `p.ImportLanguages`
- [ ] do something when the package version is `nil`

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,5 +2,7 @@
 
 ### Bug Fixes
 
-- [codegen] #702: Fixes `pulumi convert` when targeting Java. 
-  This was done by generating maven configuration (`pom.xml`) and moving the generated Java files to a proper package. 
+- [codegen] [#702](https://github.com/pulumi/pulumi-java/issues/702): 
+  Fixes `pulumi convert` (when targeting Java). Now a maven 
+  configuration (`pom.xml` with dependencies) is created and
+  the generated Java files are moved to a proper package. 


t0yv0:
This is excellent stuff! Thank you.

I took some time to build a bit on top of this to continue https://github.com/pulumi/pulumi-java/pull/786

- maven is a good default
- it is right for GenerateProject to generate maven, I checked that nodejs GenerateProject does it
- hardcoding is ok for now, great start
- I've added an integration test, skipped until we can bump deps from `pulumi/pulumi` to `pulumi/pulumi-java` and ship that; once that's done we can unskip and it keeps working

The ideas to use templates etc are good but happy to defer for much later.

t0yv0:
Merged via https://github.com/pulumi/pulumi-java/pull/786 🙇

PR #677, opened 2022-06-07
Add stack transformations

pawelprazak:
# Description

Fixes #478 

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

Issue #478, opened 2022-04-26
Support stack transformations


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

This is a feature supported in other language SDKs but missing from the Java SDK.

https://www.pulumi.com/docs/intro/concepts/resources/options/transformations/#stack-transformations

pawelprazak:
Unfortunately, setting transformations on the context is not possible with the current implementation of `Resource`.

Since `Stack` inherits `Resource` initialization logic, the transformations are registered in the constructor.
Also the `Stack` needs to exist **before** any user code is executed (including setting the transformations on the context).

Alternative APIs:

dotnet inspired (very easy to implement)
```
Pulumi.run(ctx -> { ... }, StackOptions.builder().transformations(...).build())
```
or a "fluent" approach (new public interface will be required for the `withStackTransformations` to have something to return
```
Pulumi
    .withStackTransformations(...)
    .run(ctx -> {...}
```

I personally don't have strong preference.

PR #775, opened 2022-08-04
Add `stack-transformation` to integration test matrix (pr.yml)

myhau:
# Description

I've found this by accident and have a hunch that this change was supposed to be merged along with #677.

Related to #478.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (comment: Not required)

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change (comment: Not required?)

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version (comment: No)


t0yv0:
This is now on main (empty diff) ✔️

PR #778, opened 2022-08-09
Add integration tests for aliases + fix a discovered bug 

### LINKED ISSUE #712: Port alias explosion test
## Hello!

## Issue details

From  https://github.com/pulumi/pulumi/pull/10005 - we need to port the test to the Java SDK.

myhau:
# Description

Fixes #712 (also fixes a bug that these tests have revealed — see pending changelog)

This PR:
- adds integration tests for aliases ([ported existing tests](https://github.com/pulumi/pulumi/tree/6f55e1a82a0d123e295be6cf82798441a208525d/tests/integration/aliases) + I also copied and adapted the code-comments)
- fixes a bug that was discovered by these tests (we were trying to add an entry to an immutable list when copying parent aliases)
- cleans up code related to `PULUMI_REPO_ROOT` env variable (removed — it's not used anywhere)

## Checklist

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version (Comment: no?)


sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
```diff
@@ -191,7 +191,7 @@ protected Resource(
             options.protect = options.protect || options.parent.protect; // TODO: is this logic good?
 
             // Make a copy of the aliases array, and add to it any implicit aliases inherited from its parent
-            options.aliases = options.aliases == null ? new ArrayList<>() : copyNullableList(options.aliases);
+            options.aliases = options.aliases == null ? new ArrayList<>() : new ArrayList<>(options.aliases);
```
myhau, line 194: ➡️ bugfix

tests/integration/aliases/adopt-into-component/step1/src/main/java/com/pulumi/example/Main.java
```diff
@@ -0,0 +1,104 @@
...
+        // The resource creation was moved from top level to inside the component.
+        var adoptedResource = new Resource(
+                name + "-child",
+                ComponentResourceOptions.builder()
+                    // With a new parent
+                    .parent(this)
+                    // But with an alias provided based on knowing where the resource existing before - in this case at top
+                    // level.  We use an absolute URN instead of a relative `Alias` because we are referencing a fixed resource
+                    // that was in some arbitrary other location in the hierarchy prior to being adopted into this component.
+                    .aliases(Alias.withUrn(String.format("urn:pulumi:%s::%s::%s::%s", stackName, projectName, adoptedResourceType, adoptedResourceName)))
```
myhau, line 37: Is there a better way of doing this in Java? I mean something like `Pulumi.Urn.Create` in C#:

```c#
new ComponentResourceOptions
{
    Parent = this,
    Aliases = { Pulumi.Urn.Create("res2", "my:module:Resource").Apply(urn => new Alias { Urn = urn }) },
});
```

pawelprazak: `com.pulumi.core.internal.Urn` should be useful, might require small usability changes

tests/integration/aliases/rename-component/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+                credentials {
+                    username = gprUser
+                    password = gprToken
+                }
+            }
+        }
+    }
+}
+
+var pulumiJavaSdkVersion = System.getenv("PULUMI_JAVA_SDK_VERSION") ?: "0.5.2"
```
pawelprazak: this should probably be:
`var pulumiJavaSdkVersion = System.getenv("PULUMI_JAVA_SDK_VERSION") ?: "0.0.1"`

we currently have a hack that uses `0.0.1` as a special version used in dev

pawelprazak (commented):
thank you for your contribution, have couple of questions

also, could we skip at least `gradle-wrapper.jar`, please?

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
pawelprazak: I'm not sure about this fix, will this handle `null` aliases?

myhau:
> also, could we skip at least gradle-wrapper.jar, please?

I know, it's quite verbose and there is a lot of duplication.

I could convert these projects to Maven and keep just a single `pom.xml` file for each test. What do you think?

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
myhau: Yes, notice that `options.aliases == null` check is there already. 

I've decided to refactor it anyway (last commit). "Aliases work" is encapsulated in a single method now. Let me know if this is better + if this fits the codebase style (it's a bit more functional now).

tests/integration/aliases/rename-component/build.gradle
myhau: You might have been looking at the old files. This was reverted back to `0.01`.

tests/integration/aliases/adopt-into-component/step1/src/main/java/com/pulumi/example/Main.java
myhau: 👍 I saw that, but it is annotated with `@InternalUse`. 
Can we leave the test as is and refine it when some more convenient approach is there (e.g. expose `Urn`). I can create an issue if there aren't any.

pawelprazak:
> > also, could we skip at least gradle-wrapper.jar, please?
> 
> I know, it's quite verbose and there is a lot of duplication.
> 
> I could convert these projects to Maven and keep just a single `pom.xml` file for each test. What do you think?

I was mostly concerned about adding a `.jar` file to git, but if you feel like moving to maven, go for it :)

t0yv0:
Sorry looks like I'm sliding into next week for reviewing this 🙏 

Could you please merge `main` back in the meanwhile. There are a bunch of fixes that will help with CI running the tests.

myhau:
> Sorry looks like I'm sliding into next week for reviewing this 🙏
> 
> Could you please merge `main` back in the meanwhile. There are a bunch of fixes that will help with CI running the tests.

@t0yv0 

Done, btw is rebase + force push ok or should I do merges?


> > > also, could we skip at least gradle-wrapper.jar, please?
> > 
> > 
> > I know, it's quite verbose and there is a lot of duplication.
> > I could convert these projects to Maven and keep just a single `pom.xml` file for each test. What do you think?
> 
> I was mostly concerned about adding a `.jar` file to git, but if you feel like moving to maven, go for it :)

@pawelprazak

Done. I've decided to remove all the files related to gradle wrapper.

t0yv0:
Rebase is fine! Having a look.

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
```diff
@@ -311,6 +300,25 @@ private static ImmutableMap<String, ProviderResource> convertToProvidersMap(@Nul
         return result.build();
     }
 
+    private static List<Output<String>> computeAliases(String name, String type, ResourceOptions options) {
```
t0yv0, line 303: This "looks" right but I'm not super sure; it would be absolutely lovely if we could port unit tests for the alias computation in addition to the behavior integration tests. I found an example here:

https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/test/langhost/aliases/__main__.py#L17

That would also be a great hint to maintainers on what this function is supposed to do.

tests/examples/java_examples_test.go
```diff
@@ -274,7 +274,6 @@ func getJavaBaseNew(
 	}
 	opts := integration.ProgramTestOptions{
 		Dir: filepath.Join(repoRoot, "examples", dir),
-		Env: []string{fmt.Sprintf("PULUMI_REPO_ROOT=%s", repoRoot)},
```
t0yv0, line 277: Great cleanup this indeed seems unused!

tests/integration/util.go
```diff
@@ -0,0 +1,52 @@
...
+		Config: map[string]string{
+			"org": "pulumi-bot",
+		},
+		PrepareProject: func(*engine.Projinfo) error {
+			return nil // needed because defaultPrepareProject does not know about java
+		},
+	}
+	opts = opts.With(testSpecificOptions)
+
+	// local environment, to run locally offline, make sure you set:
```
t0yv0, line 31: This is nice! TBH I would support just switching to file-state backend completely in CI also. Java is a new addition to the Pulumi ecosystem and we do not need to test the cloud backend per se, and we are pretty sure the kinds of bugs we're trying to find will be found using filestate backend anyway. I don't suggest we make edits here I'll just make a ticket. https://github.com/pulumi/pulumi-java/issues/811

tests/integration/aliases/rename-component-and-child/build.gradle
```diff
@@ -0,0 +1,41 @@
+plugins {
+    id 'application'
+}
+
+group 'com.pulumi.example'
+version '0.0.1-SNAPSHOT'
+
+var gprUser = project.findProperty("gpr.user")  ?: System.getenv("GPR_USER")
```
t0yv0, line 8: We've removed everything connected with GPR_* this is not needed anymore and can be deleted! The build files on CI will find the Java SDK dependency via mavenLocal.

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resources.java
```diff
@@ -45,4 +45,8 @@ protected static <T extends Copyable<T>> T copyNullable(@Nullable T original) {
         }
         return original.copy();
     }
+
+    protected static <T> List<T> orEmpty(@Nullable List<T> list) {
```
t0yv0, line 49: Nit: this is a bit weird to put it here (though protected); I'd perhaps just inline this.

t0yv0 (approved):
Thank you! This is a valuable contribution! I left a few suggestions, LMK if you have a moment to complete, but generally LGTM.

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resources.java
pawelprazak: I agree, esp. because it is protected, I'd just inline. A little repetition is better than a ton of dependency in my book ;)

pawelprazak (approved):
LGTM, got one nit

myhau:
Thanks for feedback! I'll work on improvements and let you know when it's ready.

tests/integration/util.go
myhau: This code was not added in this PR, I just moved these functions. I definitely agree with your point, though.

tests/integration/aliases/rename-component-and-child/src/main/java/com/pulumi/example/Main.java
```diff
@@ -0,0 +1,35 @@
+package com.pulumi.example;
```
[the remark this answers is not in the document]
myhau: These tests were ported. I've [added a README file](https://github.com/pulumi/pulumi-java/pull/778/commits/25b4e7cf9a3c4b5b431c66351bc3140be442b617). 

Let me know if this is fine. I can also make it more granular and create a README file for each test directory (and link to a specific test directory in the dotnet sdk codebase, e.g. [adopt-into-component](https://github.com/pulumi/pulumi/tree/master/tests/integration/aliases/dotnet/adopt_into_component)).

tests/integration/aliases/rename-component-and-child/build.gradle
myhau: done

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
myhau: Makes sense 👍.

Is this a must have now or can I create an issue for this and do the work in a separate PR?

myhau:
> Thanks for feedback! I'll work on improvements and let you know when it's ready.

@pawelprazak @t0yv0 It's ready for another round :)

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
```diff
@@ -188,16 +190,7 @@ protected Resource(
             // the 'childResources' is a Synchronized Collection, so this is safe operation
             parentResource.childResources.add(this);
 
-            options.protect = options.protect || options.parent.protect; // TODO: is this logic good?
-
-            // Make a copy of the aliases array, and add to it any implicit aliases inherited from its parent
-            options.aliases = options.aliases == null ? new ArrayList<>() : copyNullableList(options.aliases);
```
pawelprazak, line 194: I'm not familiar with how this logic should work, but I've noticed we no longer set `options.aliases` here, is this right?

pawelprazak (commented):
thank you, looks good, had one question

sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
t0yv0: Separate PR should be fine 🙇

PR #777, opened 2022-08-09
Fix converter regression

pawelprazak:
# Description

Unit tests in separate PR: https://github.com/pulumi/pulumi-java/pull/782 on top of this one

Fixes #774

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

Issue #774, opened 2022-08-04
Changes in packages for Java SDK


salaboy:
### What happened?

I've created a new `pulumi kubernetes-java` project and the packages used for the Kubernetes objects had been moved to a different package. 

Package change from: `import com.pulumi.kubernetes.apps_v1` to `import com.pulumi.kubernetes.apps.v1`

### Steps to reproduce

Create new `kubernetes-java` pulumi app

### Expected Behavior

Use the correct packages for the kube APIs

### Actual Behavior

wrong package is used

### Versions used

CLI          
Version      3.37.2
Go Version   go1.17.12
Go Compiler  gc

Plugins
NAME  VERSION
java  unknown

Host     
OS       darwin
Version  12.4
Arch     x86_64

This project is written in java: java='openjdk 17.0.2 2022-01-18' javac='17.0.2' maven='Apache Maven 3.8.5 (3599d3414f046de2324203b78ddcf9b5e4388aa0)'

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

salaboy:
Fixing the package import doesn't solve the problem as it seems to be a dependency chain issue. When manually updating to the new packages I am getting this issue: 

```

java.lang.UnsupportedOperationException: Convert [com.pulumi.kubernetes.apps.v1.Deployment.spec]: Error converting 'com.google.common.collect.RegularImmutableMap' to 'TypeShape{type=class com.pulumi.kubernetes.apps.v1.outputs.DeploymentSpec, parameters=[]}'. argument type mismatch

```

So I thought about sharing which maven dependencies are being used: 

<img width="449" alt="Screenshot 2022-08-04 at 10 34 43" src="https://user-images.githubusercontent.com/271966/182815001-ded70edd-be7a-44f8-9af3-1051c5920680.png">

I wonder if there is a way just to get a stable `com.pulumi:kubernetes` package.

stack72:
Hi @salaboy 

Thanks for opening this :) Would you try this package? https://mvnrepository.com/artifact/com.pulumi/kubernetes/3.20.2 I believe that this is ok?

Paul

salaboy:
Yeah.. let me give that a try

salaboy:
@stack72 do you know which version of com.pulumi:pulumi version goes with 3.20.2 ?

salaboy:
[...]

Issue #750, opened 2022-07-19
Remove deprecated codegen for custom types before GA


pawelprazak:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Follow up for #390 
- [x] wait for providers release with the new codegen
- [x] port unit tests
- [ ] remove the old mechanism from SDK

### Affected area/feature

- codegen - `@Export` annotation

PR #782, opened 2022-08-10
Port custom types tests from old to new

pawelprazak:
# Description

On top of #777 

Part of #750 
Relates to #774

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,11 +8,11 @@
 - [sdk] [#774](https://github.com/pulumi/pulumi-java/issues/774) 
   fixes deserialization regression in new codegen specific part of SDK, 
   complex @CustomType's are deserializable now
-- [sdk] [#774](https://github.com/pulumi/pulumi-java/issues/774) 
-  fixes deserialization regression in new codegen specific part of SDK, 
-  complex @CustomType's are deserializable now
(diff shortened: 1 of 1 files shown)


CHANGELOG_PENDING.md
```diff
@@ -4,3 +4,10 @@
  [#776](https://github.com/pulumi/pulumi-java/pull/776)
 
 ### Bug Fixes
+
+- [sdk] [#774](https://github.com/pulumi/pulumi-java/issues/774) 
+  fixes deserialization regression in new codegen specific part of SDK, 
+  complex @CustomType's are deserializable now
```
t0yv0, line 10: Seems duplicative changelog but I can tidy it up later before releasing

t0yv0 (approved):
This fixes the following scenario for me:


```
pulumi new kubernetes-java
# edit pom.xml to ref pulumi 0.0.1 built from source
# edit pom.xml to ref latest kubernetes
mvn clean
pulumi preview
```

This works with pre-release but not released version of Pulumi. Good job.

CHANGELOG_PENDING.md
pawelprazak: yes, there were two regressions, the second one was just a debug message

pawelprazak: sorry, I missed it

pawelprazak: should be ok now

PR #784, opened 2022-08-10
Check SDK changes against latest pulumi/examples

t0yv0:
# Description

Final fix to make builds green in pulumi-java.

Similar to recent change in templates/ test, this checks latest SDK from the PR (or main) against select versions of examples in "pulumi/examples". This is no longer a submodule-pinned reference but a floating reference to "master". With the recent changes, the examples now compile and work.

Note that these tests do not longer check that rebuilding resource providers with changes present in codegen would actually result in providers that still pass these example tests. 

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -178,6 +178,7 @@ jobs:
         - random
     steps:
       - uses: actions/checkout@v3
+      - run: make clone_examples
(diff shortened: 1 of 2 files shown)


dixler (approved):

pawelprazak:
thank you!

PR #789, opened 2022-08-11
Do not escape methods in non-resource generated provider classes

### LINKED ISSUE #787: Codegen should escape resource methods only on resource classes
## Hello!

## Issue details

Currently every getter and setter is escaped to avoid collision with methods inherited from Resource class family, but should only apply to the Resource child classes, not all classes.

This is the current entry point to the escape logic:
image

Found during this PR: #766 

Ideas for a clean implementation that would differentiate between resource and non-resource classes very much welcome :)

pawelprazak:
# Description

Fixes #787 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,3 +17,6 @@
   dependency to fix the issue that manifested as:
 
   ```UnsupportedOperationException: Convert [...]: Error```
+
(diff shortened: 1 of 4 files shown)


t0yv0 (approved):
Perfect, ty!

pawelprazak:
rebased

PR #793, opened 2022-08-11
Try buliding without Github Package Repo

t0yv0:
# Description

At this point in development GitHub Package Repository is mostly getting in the way. We use it to:

1. exchange JARs between build phases - now that providers are out, this is just the SDK jar
2. maintain a pre-release build repository

Some of the issues with it were worked around in the build/ tool, but there's more issues. Third-party PRs fail to authenticate to GPR and therefore fail to run tests. I'm not sure I know if anyone is using (2) GPR as a pre-release repository. 

I think it's a win for us to totally remove it - which is what this PR is about, at least for now. For pre-release builds we will have to build from source.

Fixes # (issue)

## Checklist


pawelprazak (approved):
Less complexity sounds good.

.github/workflows/pr.yml
```diff
@@ -85,16 +83,15 @@ jobs:
...
-        # See Note 1.
-        if: |
-          github.event_name == 'repository_dispatch'
-          || (
-                github.event.pull_request.head.repo.full_name == github.repository
-                && ! startsWith(github.ref_name, 'dependabot/')
-             )
-          || github.ref_name == 'main'
+      - name: Publish Pulumi Java SDK to Maven Local
+        uses: gradle/gradle-build-action@9b814496b50909128c6a52622b416c5ffa04db49
```
sideeffffect, line 87: Why this commit? https://github.com/gradle/gradle-build-action/commit/9b814496b50909128c6a52622b416c5ffa04db49

Would just v2.2.2 work?

t0yv0: Eh, copied from elsewhere thoughtlessly. It should work yes. We should probably clean up and use a consistent version across all the build files (in a separate PR).

PR #797, opened 2022-08-12
Codegen schema override not affecting version.txt path

pawelprazak:
# Description

Fix a bug discovered during https://github.com/pulumi/pulumi-azure-native/pull/1855.

- schema package overrides will no longer affect the version.txt path in Utilities
- do not use Java package name, because Bootstrap logic expects the Pulumi package in the path

## Checklist

### DIFF
--- a/pkg/codegen/java/gen.go
+++ b/pkg/codegen/java/gen.go
@@ -1754,7 +1754,7 @@ func (mod *modContext) gen(fs fs) error {
 			}
 			return javaUtilitiesTemplate.Execute(ctx.writer, javaUtilitiesTemplateContext{
 				Name:        pkgName.String(),
-				PackagePath: strings.ReplaceAll(mod.packageName, ".", "/"),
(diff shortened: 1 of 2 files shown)


pkg/codegen/java/gen.go
```diff
@@ -1754,7 +1754,7 @@ func (mod *modContext) gen(fs fs) error {
 			}
 			return javaUtilitiesTemplate.Execute(ctx.writer, javaUtilitiesTemplateContext{
 				Name:        pkgName.String(),
-				PackagePath: strings.ReplaceAll(mod.packageName, ".", "/"),
```
pawelprazak, line 1757: do not use Java package name, because Bootstrap logic expects the Pulumi package in the path

t0yv0 (approved):
It would be great to add a test into TestGeneratePackage with an example schema that has this, and a unit test on top that asserts `version` is read correctly; but we can backlog it, I understand you need the patch quick.

PR #805, opened 2022-08-12
TF 0.12 module support

T-Vova:
Fully implementing Terraform 0.12 Module conversion for all languages so that support for Terraform 0.13 and beyond can be added (https://github.com/pulumi/tf2pulumi/issues/186).


t0yv0:
Mind adding a test? There seems to be a lint problem also. Thank you.

iwahbe:
@t0yv0 Do you know if this PR is still worth preserving?

Zaid-Ajaj:
Since we now support converting terraform using [pulumi-converter-terraform](https://github.com/pulumi/pulumi-converter-terraform) including support for modules, it makes this PR redundant. We do however need to support generating correct Java code from terraform modules at some point in future in the form of the resource components like we do for C# now.

PR #813, opened 2022-08-18
Add type mismatches tests

### LINKED ISSUE #643: Do not panic when the type of a property in resource outputs doesn't match
## Hello!

## Issue details

The motivation behind it can be found in:
https://github.com/pulumi/pulumi/issues/7329

There is also a PR implementing it for C#.

https://github.com/pulumi/pulumi/pull/8286

The fix proceed with a null / default value and a warning in place of an error. This makes some programs succeed.

Consider implementing the same here.

pawelprazak:
# Description

Fixes #643 

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/Converter.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/Converter.java
@@ -138,43 +138,23 @@ public class Converter {
         if (value == null) {
             if (targetIsOptional) {
                 // A null value coerces to a Optional.empty.
+                var empty = Optional.empty();
+                //noinspection unchecked,rawtypes
                 return tryEnsureType(
                         String.format("%s %s", context, targetType.getTypeName()),
-                        Optional.empty(),
-                        targetType
+                        empty, (TypeShape<Optional>) targetType, empty
                 );
             }
 
-            // We're null and we're NOT converting to an Optional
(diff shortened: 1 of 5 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/Converter.java
```diff
@@ -530,52 +553,68 @@ private void tryWriteJson(String context, JsonWriter jsonWriter, @Nullable Objec
...
-            return Either.ofLeft(
-                    tryConvertObjectInner(
-                            String.format("%s.left", context),
-                            value,
-                            leftType
-                    )
-            );
-        } catch (Exception leftException) {
-            try {
+            if (canBeCasted(value, leftType)) {
```
t0yv0, line 587: https://grammarist.com/usage/casted/#:~:text=The%20verb%20cast%20is%20conventionally,to%20cast%20in%20modern%20English. perhaps `canBeCast`.

t0yv0 (approved):

sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/Converter.java
pawelprazak: thank you

pawelprazak:
rebased, addressed feedback

Issue #817, opened 2022-08-24
pulumi java-gradle fails with cryptic message


nagkumar:
### What happened?

when build.gradle main class script changed from 

```
 mainClass = project.hasProperty("mainClass")
            ? project.getProperty("mainClass")
            : 'com.tejasoft.devops.iac.pulumi.gradle.GradlePulumiApp'
```

to 
` mainClass =  'com.tejasoft.devops.iac.pulumi.gradle.GradlePulumiApp'`

then `pulumi about`

fails with exception

```
G:\DWork\osource\cloudc\devops\pulumi\java\java-gradle>pulumi about
Exception in thread "main" java.lang.IllegalArgumentException: Program run without the Pulumi engine available; re-run using the `pulumi` CLI

        at com.pulumi.deployment.internal.DeploymentImpl$DeploymentState.lambda$fromEnvironment$2(DeploymentImpl.java:1627)

        at com.pulumi.core.Left.orThrow(Left.java:57)

        at com.pulumi.deployment.internal.DeploymentImpl$DeploymentState.fromEnvironment(DeploymentImpl.java:1632)

        at com.pulumi.deployment.internal.DeploymentImpl.fromEnvironment(DeploymentImpl.java:158)

        at com.pulumi.internal.PulumiInternal.fromEnvironment(PulumiInternal.java:39)

        at com.pulumi.Pulumi.runAsync(Pulumi.java:39)

        at com.pulumi.Pulumi.run(Pulumi.java:28)

        at com.tejasoft.devops.iac.pulumi.gradle.GradlePulumiApp.main(GradlePulumiApp.java:10)

Caused by: java.lang.IllegalArgumentException: expected environment variable: 'PULUMI_MONITOR', got: [DCA_BU_ROOT..]

        at com.pulumi.core.internal.Environment.getEnvironmentVariable(Environment.java:42)

        ... 6 more

FAILURE: Build failed with an exception.

* What went wrong:

Execution failed for task ':app:run'.

> Process 'command 'D:\Apps\Dev\langs\java\jdk\17.x\bin\java.exe'' finished with non-zero exit value 1

* Try:

> Run with --stacktrace option to get the stack trace.

> Run with --info or --debug option to get more log output.

> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s

CLI
Version      3.38.0
Go Version   go1.19
Go Compiler  gc

Plugins
NAME  VERSION
java  unknown

Host
[...]

Issue #785, opened 2022-08-10
[sdk/java] improve exception output


dixler:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Was reviewing https://github.com/pulumi/examples/pull/1260

The error message could be better.
```
  pulumi:pulumi:Stack (gcp-java-gke-hello-world-dev):
    [INFO] Scanning for projects...
    [INFO]
    [INFO] ----------------< com.pulumi:gcp-java-gke-hello-world >-----------------
    [INFO] Building gcp-java-gke-hello-world 1.0-SNAPSHOT
    [INFO] --------------------------------[ jar ]---------------------------------
    [INFO]
    [INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ gcp-java-gke-hello-world ---
    [INFO] Using 'UTF-8' encoding to copy filtered resources.
    [INFO] skip non existing resourceDirectory /home/kdixler/Documents/examples/gcp-java-gke-hello-world/src/main/resources
    [INFO]
    [INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ gcp-java-gke-hello-world ---
    [INFO] Nothing to compile - all classes are up to date
    [INFO]
    [INFO] --- exec-maven-plugin:3.0.0:java (default-cli) @ gcp-java-gke-hello-world ---
 
    Aug 10, 2022 8:16:53 AM com.pulumi.deployment.internal.DeploymentImpl$DefaultEngineLogger errorAsync
    SEVERE: Running program [PID: 217938](/usr/lib/jvm/java-11-openjdk/bin/java -classpath /opt/maven/boot/plexus-classworlds-2.6.0.jar -Dclassworlds.conf=/opt/maven/bin/m2.conf -Dmaven.home=/opt/maven -Dlibrary.jansi.path=/opt/maven/lib/jansi-native -Dmaven.multiModuleProjectDirectory=/home/kdixler/Documents/examples/gcp-java-gke-hello-world org.codehaus.plexus.classworlds.launcher.Launcher --no-transfer-progress compile exec:java) failed with an unhandled exception:
    java.util.NoSuchElementException: No value present
    	at java.base/java.util.Optional.orElseThrow(Optional.java:382)
    	at gcpgke.App.stack(App.java:94)
[...]

PR #820, opened 2022-08-26
Hide plugin discovery errors

pawelprazak:
# Description

The Pulumi CLI will no longer display, what look like a duplicate stack traces.
The language host was modified to hide error messages from an optional process
of plugin discovery through classpath introspection.

Also `plugin about` will no longer display errors from plugin discovery.
To display errors from plugin discovery, use config option `runtime.options.v`
with a value of 1 or greater.

Part of #817 and #785

## Checklist


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -195,7 +195,7 @@ func (host *javaLanguageHost) determinePulumiPackages(
 	// Run our classpath introspection from the SDK and parse the resulting JSON
 	cmd := exec.Cmd
 	args := exec.PluginArgs
-	output, err := host.runJavaCommand(ctx, exec.Dir, cmd, args)
+	output, err := host.runJavaCommand(ctx, exec.Dir, cmd, args, true)
```
t0yv0, line 198: LGTM in principle but have a nit:

Ok this is reasonable but note this:

- this is already printing the error into logging.V(3) on L201 so I think perhaps this should also send stderr/stdout via `logging.V(3)` in the case of error

- yes to runJavaCommand not printing stderr/stdout in the case of plugins; we want that to be much more quiet

- but printing to stdout if logging.V(1) is set is very weird - why not just send the data via the logging framework instead of printing to stdout

sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/Converter.java
```diff
@@ -437,7 +437,8 @@ private Object tryConvertObjectInner(
                             argument == null ? "null" : argument.getClass()
                     ), e);
                 } catch (IllegalAccessException | InvocationTargetException e) {
-                    throw new IllegalStateException(String.format("Unexpected exception: %s", e.getMessage()), e);
+                    var exMsg = e.getMessage() == null ? e.getClass().getSimpleName() : e.getMessage();
```
t0yv0, line 440: This went in through another PR I think so the diff will disappear after we rebase on main.

t0yv0:
Implemented by own nits in https://github.com/pulumi/pulumi-java/pull/825

t0yv0 (approved):

t0yv0:
Should be good to merge when tests pass.  Lots of merge commits here but final diff should be smaller.

PR #825, opened 2022-08-29
Hide bootstrap errors

t0yv0:
# Description

I have confirmed that the fix removes duplicate reporting of compilation errors, we very much want that included CC @dixler 

Now this version implements my nits, I slightly prefer this version since multiwriter sends data to stdout as it comes instead of waiting (better for slow cmds) and this version doesn't mixup logging flags with stdout verbosity which can be unexpected.

Fixes # (issue)

## Checklist

### DIFF
--- a/tests/integration/aliases/adopt-into-component/build.gradle
+++ b/tests/integration/aliases/adopt-into-component/build.gradle
@@ -11,18 +11,6 @@ repositories {
         url("https://maven-central.storage-download.googleapis.com/maven2/")
     }
     mavenCentral()
-
-    if (gprUser) {
-        repositories {
-            maven {
-                url = uri("https://maven.pkg.github.com/pulumi/pulumi-java")
-                credentials {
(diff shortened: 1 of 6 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java
```diff
@@ -376,6 +362,26 @@ private static ImmutableMap<String, ProviderResource> convertToProvidersMap(@Nul
         return result.build();
     }
 
+    private static List<Output<String>> computeAliases(String name, String type, ResourceOptions options) {
```
pawelprazak, line 365: if this is intentionally in this PR then all good, I just looks like a fragment of another PR ;)

pawelprazak (commented):
looks good, thanks! the milti-writer was cool

one thing: the commit history looks like something went wrong while rebasing/merging

PR #834, opened 2022-09-07
Remove null-check from codegen

pawelprazak:
# Description

Fixes #821 

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #837, opened 2022-09-12
Fix URN parsing

t0yv0:
# Description

Fixes # (issue)

As @pawelprazak points out in https://github.com/pulumi/pulumi-java/pull/834 we have a problem for type identifiers without a module. I've checked what happens in Node codegen to double-check. Codegen emits "package::type" form but URN fragments must be of the "package:type" form in this case. This mostly is not a problem for the Java SDK since URNs are allocated by the engine that does the right thing, but in case of mock-testing we hit the Type.parse function here and have a problem. SO the fix makes Type.parse more tolerant so that mock tests succeed.

Further I discovered a few gems in Java stdlib that needed fixing, namely `"t:m:".split(":").length == 2` - that was utterly surprising.

## Checklist


sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -99,7 +105,7 @@ public static Urn parse(String urn) {
                 () -> format("expected urn to start with '%s', got: '%s'", Prefix, urn)
         );
 
-        var urnParts = urn.split(PartsSeparator);
+        var urnParts = urn.split(PartsSeparatorRegex, -1); // -1 avoids dropping trailing empty strings
```
pawelprazak, line 108: wow, thanks for spotting that, that's surprising! not a good thing from an API ;)

pawelprazak (approved):
looks good thank you!

PR #838, opened 2022-09-13
Restore running junit on top of generated code

t0yv0:
# Description

Lifting the fix for running junit based tests on top of generated code examples that @pawelprazak discovered in https://github.com/pulumi/pulumi-java/pull/834

A couple of small amendments:

- the generation of junit code is gated by a flag in PackageInfo so as to minimize the diff when rolling this out to providers (they do not use this functionality)

- the tests are re-arranged so the dependencies are specific to the test, closing an old TODO and speeding up running the codegen tests quite a bit

- mini-awsclassic that now gets re-enabled detects another regression that slipped through - we stopped tolerating the situation where the provider does not return a required output although the schema marks it required; this is duplicative of another issue so the test is skipped

- jumbo-resources adopts @pawelprazak's fix to mocks but will need https://github.com/pulumi/pulumi-java/pull/837 checked in first

- output-funcs-edgeorder was duplicated (inherited from pulumi/pulumi as well as listed as java-specific) - this is now corrected

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/java/gen_test.go
+++ b/pkg/codegen/java/gen_test.go
@@ -3,6 +3,10 @@
 package java
 
 import (
+	"bytes"
+	"encoding/json"
+	"os/exec"
+	"path/filepath"
 	"testing"
 
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
@@ -19,7 +23,7 @@ type generatePackageTestConfig struct {
 
 // Configures tests that are maintained in the pulumi-java repository
 // specifially for Java codegen testing.
-func javaSpecificTests() []generatePackageTestConfig {
+func javaSpecificTests(keyDeps map[string]string) []generatePackageTestConfig {
 	return []generatePackageTestConfig{
 		newGeneratePackageTestConfig(&test.SDKTest{
 			Directory:   "mini-azurenative",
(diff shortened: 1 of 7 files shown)


pkg/codegen/java/gen_test.go
```diff
@@ -10,46 +10,100 @@ import (
...
+			Directory:   "mini-awsnative",
+			Description: "Regression tests extracted from trying to codegen aws-native",
+		}),
+		newGeneratePackageTestConfigWithExtras(&test.SDKTest{
+			Directory:   "mini-awsclassic",
+			Description: "Regression tests extracted from trying to codegen aws",
+			// TODO[pulumi/pulumi-java#821]
+			Skip: codegen.NewStringSet("java/test"),
+		}, &PackageInfo{
+			Dependencies: map[string]string{
```
pawelprazak, line 38: just nit, the deps used here are not in line with what we use in sdk

pawelprazak (approved):
looks good, thank you!

pkg/codegen/java/gen_test.go
t0yv0: Good point. I made a change to sync them to Pulumi's build.gradle file, so that Dependabot PRs upgrade all those together.

PR #913, opened 2022-12-13
Release windows as .tar.gz

Frassle:
Every other single pulumi plugin packs as a .tar.gz for all systems. We have loads of code that assumes plugins will always be .tar.gz. There is no reason for java to be a special snowflake and use .zip here.


dirien (approved):
LGTM!

PR #929, opened 2022-12-21
Demo provider record-replay

t0yv0:
# Description

Demonstrates how to use https://github.com/pulumi/pulumi/pull/11710 record-and-replay feature. 

Note that you need to install Pulumi CLI from 11710 for this to work.

Running the following example is 13s with provider replay mocks vs 200s against the actual cloud. The PR includes the pre-recorded provider mocks.

```
make test_example.aws-java-webserver
```

Fixes # (issue)

## Checklist


tests/examples/java_examples_test.go
```diff
@@ -90,6 +90,13 @@ func TestExamples(t *testing.T) {
 					assert.True(t, strings.Contains(publicHostName, "."))
 				},
 			})
+
+		test.StackName = "awsjavawebserverstack"
```
t0yv0, line 94: StackName must be set because a randomly chosen stack name currently confuses the record-replay framework and it fails to replay. Unfortunate.

PR #936, opened 2023-01-06
Fix typo in log message

thomas11:


Zaid-Ajaj (approved):

t0yv0 (approved):
Thank you! Kind of you to notice and send a fix right away. (and I probably should install a spell checker in my Emacs Go mode).

PR #966, opened 2023-02-01
[java/sdk] Delegate alias computation to the engine

### LINKED ISSUE #864: [sdks/java] Delegate alias computation to the engine
## Hello!

## Issue details

Currently, we pre-compute all possible alias combinations in the java sdk [here](https://github.com/pulumi/pulumi-java/blob/86f33978a6c8c79f0ef776e9a4c289389818e50d/sdk/java/pulumi/src/main/java/com/pulumi/resources/Resource.java#L365). When each resource in an inheritance chain has many aliases, this quickly grows multiplicatively. Instead, we should simply compute the aliases for the resource we're registering, and allow the engine to do this computation for us. Additionally, we should ensure we send the structured version of aliases across rather than just raw URNs.

### Affected area/feature

SDK

Zaid-Ajaj:
Fixes #864
 - Sends full alias specification to the Pulumi engine. 
 - Removes `computeAliases` entirely that causes "alias explosion"
 - Doesn't account for older engines as discussed 

## Checklist

### DIFF
--- a/tests/examples/minimal/pom.xml
+++ b/tests/examples/minimal/pom.xml
@@ -29,7 +29,7 @@
           <dependency>
             <groupId>io.grpc</groupId>
             <artifactId>grpc-okhttp</artifactId>
-            <version>1.51.1</version>
+            <version>1.52.1</version>
             <exclusions>
               <exclusion>
                 <groupId>com.google.guava</groupId>
@@ -65,7 +65,7 @@
       <dependency>
         <groupId>io.grpc</groupId>
         <artifactId>grpc-okhttp</artifactId>
-        <version>1.51.1</version>
+        <version>1.52.1</version>
         <exclusions>
           <exclusion>
             <groupId>com.google.guava</groupId>


t0yv0:
I have some suggested fixes in https://github.com/pulumi/pulumi-java/pull/967

- DeploymentImpl class is stylistically tortured, move the new code into a separate class
- Use Output combinators as far as possible, drop to CompletableFuture at the end
- Do not join() the CompletableFuture as this should not be necessarry
- Do case analysis explicitly, if I understand this right there's three cases:

```

message Alias {
    message Spec {
        string name = 1;        // The previous name of the resource.  If none is provided, we will use the current name.
        string type = 2;        // The previous type of the resource. If none is provided, we will use the current resoource type.
        string stack = 3;       // The previous stack of the resource. If not set, the current stack of the resource is used.
        string project = 4;     // The previous project of the resource. If not set, the current project of the resource is used.

        // The previous parent of the resource. If not set, the current parent of the resource is used.
        oneof parent {
            string parentUrn = 5;   // The urn of the previous parent.
            bool noParent = 6;      // Used to indicate the resource previously had no parent. If false this property is ignored.
        }
    }
    oneof alias {
        string urn = 1; // The previous urn to alias to.
        Spec spec = 2;  // An alias specification.
    }
}

```

Essentially alias:


```
data Alias = NoParent | ParentUrnAlias URN | SpecAlias Name Type Stack Project ParentURN
```

t0yv0:
Some kind of test would be splendid also, or some indication that this is tested by existing suite and passing. Thank you!

It looks like there's an integration test `runAliasTest` so perhaps can we verify it passed? Sometimes CI bit-rots.

Otherwise LGTM!

t0yv0:
Oh perfect, the test suite is active! In fact it flagged I got something borked in #967.  So it does verify aliases in this PR at integration level. Looking closely.

t0yv0 (approved):

Zaid-Ajaj:
> Essentially alias:

> `data Alias = NoParent | ParentUrnAlias URN | SpecAlias Name Type Stack Project ParentURN`

It is not very clear from the java API but it would be something like this:

<details>
  <summary>Show type definition</summary>

```fs
type AliasParent =  
 | NoParent of bool
 | ParentUrn of urn:string

type AliasSpec = {
  name: string
  type: string
  stack: string
  project: string
  parent: AliasParent
}

type Alias = 
   | Urn of urn:string
   | Spec of AliasSpec
```
</details>

> Oh perfect, the test suite is active!

Yeah the aliases integration tests are working as expected. Locally as well

> Do not join() the CompletableFuture as this should not be necessarry

Unfortunately, aliases are an odd case of being Output-y for no reason and needing to be resolved/awaited using `.join()` at the SDK level, not the engine. Every SDK had this problem with aliases and it's redundant

t0yv0:
We also should almost put this F# definition of alias of yours into every comment on the Alias type.. That would save so much trouble.

PR #981, opened 2023-02-12
Lint depending on pkg except for codegen

iwahbe:
To avoid including all pkg deps in tf-bridge, we want to move codegen into its own module. Since codegen depends on pulumi-java, we need to make sure that pulumi-java doesn't pull in the rest of pulumi/pkg. We can enforce this with a lint rule.

The rest of this PR is getting the linter happy.

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -291,7 +291,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}
-          working-directory: ${{ pkg }}
+          working-directory: pkg
           args: --config ../.golangci.yml ./...
       - name: Ensure no pulumi/pkg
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -294,8 +291,10 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}
-          working-directory: ${{ matrix.directory }}
+          working-directory: ${{ pkg }}
```
t0yv0, line 294: Interpolation does not look right. Is pkg a variable?

t0yv0 (approved):

.github/workflows/pr.yml
iwahbe: No, no it is not 😞.

PR #987, opened 2023-02-13
sdk: Explicate type parameters in a couple cases

abhinav:
There are a couple complex cases that some JDKs--or at least,
whatever VSCode's language server is using to interpret this code
has trouble with, with regards to type inference.

Add explicit type parameters for those cases
so that the code actually shows as compiling.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/core/internal/OutputData.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/core/internal/OutputData.java
@@ -283,18 +283,17 @@ public final class OutputData<T> implements Copyable<OutputData<T>> {
             CompletableFuture<OutputData<T1>> data1, CompletableFuture<OutputData<T2>> data2,
             CompletableFuture<OutputData<T3>> data3, CompletableFuture<OutputData<T4>> data4,
             CompletableFuture<OutputData<T5>> data5, CompletableFuture<OutputData<T6>> data6,
-            CompletableFuture<OutputData<T7>> data7, CompletableFuture<OutputData<T8>> data8
-    ) {
+            CompletableFuture<OutputData<T7>> data7, CompletableFuture<OutputData<T8>> data8) {
         return CompletableFuture.allOf(data1, data2, data3, data4, data5, data6, data7, data8)
                 .thenApply(ignore -> builder(Tuple0.Empty)
-                        .transform(data1.join(), Tuple0::append)
-                        .transform(data2.join(), Tuple1::append)
-                        .transform(data3.join(), Tuple2::append)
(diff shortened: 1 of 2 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/core/internal/OutputData.java
```diff
@@ -283,18 +283,17 @@ private static <T1, T2, T3, T4, T5, T6, T7, T8> CompletableFuture<OutputData<Tup
...
-                        .transform(data7.join(), Tuple6::append)
-                        .transform(data8.join(), Tuple7::append)
+                        .<T1, Tuple1<T1>>transform(data1.join(), Tuple0::append)
+                        .<T2, Tuple2<T1, T2>>transform(data2.join(), Tuple1::append)
+                        .<T3, Tuple3<T1, T2, T3>>transform(data3.join(), Tuple2::append)
+                        .<T4, Tuple4<T1, T2, T3, T4>>transform(data4.join(), Tuple3::append)
+                        .<T5, Tuple5<T1, T2, T3, T4, T5>>transform(data5.join(), Tuple4::append)
+                        .<T6, Tuple6<T1, T2, T3, T4, T5, T6>>transform(data6.join(), Tuple5::append)
+                        .<T7, Tuple7<T1, T2, T3, T4, T5, T6, T7>>transform(data7.join(), Tuple6::append)
+                        .<T8, Tuple8<T1, T2, T3, T4, T5, T6, T7, T8>>transform(data8.join(), Tuple7::append)
```
dixler, line 296: Is there a way to convey the desired type information without turning it into a staircase?

abhinav: I wasn't able to find any—I tried. This was trying to use type inference before, but it wasn't working reliably.
I don't know whether there's something wrong with the version of JRE that this is configured to use, or if it's just a limitation of the Java LSP (which is based on Eclipse, AFAIK), but it decides that this whole block is `Tuple*<Object, *>`s if we try to make it infer it.
This block is trying to be clever, and there doesn't appear to be a clean way of doing that transpose of `Tuple*<OutputData<T1>, OutputData<T2>, ...>` into an `OutputData<Tuple<T1, T2, ...>` — at least nothing obvious.

justinvp: I dislike the staircase too, but pragmatically, I'm fine with this change to improve the dev experience. It's not like we're making changes like this all over -- it's localized to just this function.

justinvp (approved):

PR #1002, opened 2023-02-27
codegen: Generate Gradle 8.0 compatible files

### LINKED ISSUE #993: Generated Gradle build files do not work on Version 8
## Hello!

## Issue details

Recently GitHub auto-upgraded Gradle to 8.0, causing providers build failure. We are trying to pin to 7.6. 

Symptom is :

Could not set unknown property 'classifier' for task ':sourcesJar' of type org.gradle.api.tasks.bundling.Jar.

For the moment we use the default Gradle template from pulumi-java-gen in most places, and if that could be fixed up it'd be ideal to roll out and unpin Gradle.

See https://github.com/pulumi/ci-mgmt/issues/351

abhinav:
GitHub now uses Gradle 8.0,
which has removed the `classifier` property
in favor of `archiveClassifier`.
https://docs.gradle.org/8.0/userguide/upgrading_version_7.html#abstractarchivetask_api_cleanup

This updates the code generator to generate build.gradle files
that use this property instead.
Change copied over
from https://github.com/pulumi/pulumi-azure-native/issues/2262#issuecomment-1442181180.

Resolves #993

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,4 @@
   - [java/sdk] Delegate alias computation to the engine. [#966](https://github.com/pulumi/pulumi-java/pull/966)
  
 ### Bug Fixes
+  - [codegen] Generate build.gradle files compatible with Gradle 8.0.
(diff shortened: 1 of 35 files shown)


dixler (approved):
LGTM. Just wait on tests to pass first :pray:

PR #1056, opened 2023-04-10
Fix broken URN parsing on nested parents.

### LINKED ISSUE #1045: Issue with creating AWSX resources (Vpc) with the Java SDK
### What happened?

When creating a Vpc resource (from Crosswalk, under `com.pulumi.awsx.ec2.*`) pulumi reports an error related to the resource URN generation. Somewhere in the chain of validating the resource URN there is a failure within one of the Java library methods. The stacktrace is as follows:

```
java.lang.IllegalArgumentException: expected qualified type to have 1 or 2 parts, split by '\$', got 'awsx:ec2:Vpc, aws:ec2/vpc:Vpc, aws:ec2/internetGateway:InternetGateway'
        at com.pulumi.core.internal.Objects.require(Objects.java:34)
        at com.pulumi.core.internal.Objects.require(Objects.java:48)
        at com.pulumi.core.internal.Urn$QualifiedTypeName.parse(Urn.java:178)
        at com.pulumi.core.internal.Urn.parse(Urn.java:131)
        at com.pulumi.serialization.internal.Deserializer.tryDeserializeResource(Deserializer.java:365)
        at com.pulumi.serialization.internal.Deserializer.deserializeCore(Deserializer.java:103)
        at com.pulumi.serialization.internal.Deserializer.deserialize(Deserializer.java:59)
        at com.pulumi.serialization.internal.Converter.convertValue(Converter.java:81)
        at com.pulumi.core.internal.OutputCompletionSource.setValue(OutputCompletionSource.java:95)
        at com.pulumi.deployment.internal.DeploymentImpl$ReadOrRegisterResourceInternal.lambda$completeResourceAsync$0(DeploymentImpl.java:1193)
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646)
        at java.base/java.util.concurrent.CompletableFuture$Completion.exec(CompletableFuture.java:483)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)
        at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
        at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
        at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
        at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)
```

### Expected Behavior

The VPC should be created as expected with the correct subnet specs.

### Steps to reproduce

1. Create a new pulumi project using `mkdir awsx-pulumi-issue && cd awsx-pulumi-issue && pulumi new java-gradle`. Answer the questionnaire with any defaults you prefer.
2. In the generated code directory, add the following line in `app/build.gradle` under the `dependencies` section: `   implementation 'com.pulumi:awsx:1.0.2'` 
3. In the generated code, replace the contents of `app/src/main/java/myproject/App.java` with

```
package myproject;

import com.pulumi.Pulumi;
import com.pulumi.awsx.ec2.Vpc;
import com.pulumi.awsx.ec2.VpcArgs;
import com.pulumi.awsx.ec2.enums.NatGatewayStrategy;
import com.pulumi.awsx.ec2.enums.SubnetType;
import com.pulumi.awsx.ec2.inputs.NatGatewayConfigurationArgs;
import com.pulumi.awsx.ec2.inputs.SubnetSpecArgs;

public class App {
        Pulumi.run(ctx -> {
            Vpc vpc = new Vpc(
                    "example-private-vpc",
                    VpcArgs.builder()
                            .subnetSpecs(
                                    SubnetSpecArgs.builder()
                                            .type(SubnetType.Private)
                                            .cidrMask(22)
                                            .build()
                            )
                            .natGateways(
                                    NatGatewayConfigurationArgs.builder()
                                            .strategy(NatGatewayStrategy.None)
                                            .build()
                            )
                            .build()
            );

            ctx.export("privateVpcId", vpc.vpcId());
        });
    }
}
```

The above might even be a little more elaborate than required to reproduce this issue. Even the following fails with the exact same error:

```
Pulumi.run(ctx -> {
            Vpc vpc = new Vpc(
                    "example-private-vpc"
            );

            ctx.export("privateVpcId", vpc.vpcId());
        });
```

4. In the pulumi code directory, run `pulumi up`

### Output of `pulumi about`

```
CLI          
Version      3.46.1
Go Version   go1.19.3
Go Compiler  gc

Plugins
NAME  VERSION
aws   5.16.2
awsx  1.0.2
java  unknown

Host     
OS       arch
Version  "rolling"
Arch     x86_64

This project is written in java: executable='/usr/bin/java' version='openjdk 17.0.6 2023-01-17
OpenJDK Runtime Environment (build 17.0.6+10)
OpenJDK 64-Bit Server VM (build 17.0.6+10, mixed mode)' java='/usr/bin/java' javac='17.0.6' maven='Apache Maven 3.8.7 (b89d5959fcde851dcb1c8946a785a163f14e1e29)' gradle='8.0.2'

Current Stack: awsx-pulumi-issue

Found no resources associated with awsx-pulumi-issue

Found no pending operations associated with awsx-pulumi-issue

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/rohanprabhu
User           rohanprabhu
Organizations  rohanprabhu, commandk

No dependencies found

Pulumi locates its logs in /tmp by default
```

### Additional context

_No response_

### Contributing

dixler:
# Description
Added failing test case for nested parenting in resource URN. Changed Parent from Optional<Type> to Optional<String>.

Fixes #1045

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
@@ -4,6 +4,8 @@ import com.pulumi.core.Output;
 import com.pulumi.core.internal.annotations.InternalUse;
 
 import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.StringJoiner;
@@ -148,11 +150,12 @@ public final class Urn {
     @ParametersAreNonnullByDefault
     @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
     public static final class QualifiedTypeName {
-        public final Optional<String> parent;
+        // Contains the entire parent string, if exists. (i.e. parent1Type$parent2Type$...)
+        public final Optional<String> parents;
         public final Type type;
 
-        public QualifiedTypeName(Optional<String> parent, Type type) {
-            this.parent = requireNonNull(parent);
+        public QualifiedTypeName(Optional<String> parents, Type type) {
+            this.parents = requireNonNull(parents);
             this.type = requireNonNull(type);
         }
 
@@ -161,18 +164,18 @@ public final class Urn {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
             QualifiedTypeName that = (QualifiedTypeName) o;
-            return parent.equals(that.parent) && type.equals(that.type);
+            return parents.equals(that.parents) && type.equals(that.type);
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(parent, type);
+            return Objects.hash(parents, type);
         }
 
         @Override
         public String toString() {
             return new StringJoiner(", ", QualifiedTypeName.class.getSimpleName() + "[", "]")
-                    .add("parent=" + parent)
+                    .add("parent=" + parents)
                     .add("type=" + type)
                     .toString();
         }
@@ -182,7 +185,7 @@ public final class Urn {
         public static QualifiedTypeName parse(String qualifiedType) {
             var qualifiedTypeParts = qualifiedType.split(ParentSeparatorRegex, -1); // -1 avoids dropping trailing empty strings
             require(qp -> qp.length > 0, qualifiedTypeParts,
-                    () -> format("expected qualified type to have 1 or 2 parts, split by '%s', got '%s'",
+                    () -> format("expected qualified type to not be empty, split by '%s', got '%s'",
                             ParentSeparatorRegex, String.join(", ", qualifiedTypeParts))
             );
             var type = qualifiedTypeParts[qualifiedTypeParts.length - 1];
@@ -193,11 +196,8 @@ public final class Urn {
 
             // Handle parents. URNs can have nested parents which should be joined with a "$".
             if (qualifiedTypeParts.length >= 2) {
-                // Remove the last qualifiedTypeParts element as it's the type.
-                String[] parentList = new String[qualifiedTypeParts.length - 1];
-
                 // Get only the parent parts of the URN type.
-                System.arraycopy(qualifiedTypeParts, 0, parentList, 0, parentList.length);
+                List<String> parentList = Arrays.asList(qualifiedTypeParts).subList(0, qualifiedTypeParts.length - 1);
 
                 var pt = String.join(ParentSeparator, parentList);
                 require(p -> isNonEmptyOrNull(p), pt,
@@ -211,7 +211,7 @@ public final class Urn {
         }
 
         public String asString() {
-            return this.parent.map(p -> p
+            return this.parents.map(p -> p
                     + ParentSeparator).orElse("")
                     + this.type.asString();
         }


sdk/java/pulumi/src/test/java/com/pulumi/core/internal/UrnTest.java
```diff
@@ -127,7 +136,7 @@ private static Stream<Arguments> testCreateUrnString() {
                         "stack", "project",
                         Optional.of("urn:pulumi:stack::project::parentPackage:parentType$package:type::name"),
                         "package2:type2", "name2",
-                        "urn:pulumi:stack::project::package:type$package2:type2::name2"
+                        "urn:pulumi:stack::project::parentPackage:parentType$package:type$package2:type2::name2"
```
dixler, line 139: I believe the original test case was incorrect because there should be 2 parents in the type since the parent resource itself has a parent.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -148,10 +150,10 @@ public String asString() {
     @ParametersAreNonnullByDefault
     @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
     public static final class QualifiedTypeName {
-        public final Optional<Type> parent;
+        public final Optional<String> parent;
```
dixler, line 153: This is an internal package so I believe it should be safe to change this.

dixler:
I have verified that the change fixes the bug when switching from an old version to the new version tested locally on my machine.

recording: https://asciinema.org/a/ejFDCEUUFxikr3vPhaqIpKdEJ

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -181,7 +183,7 @@ public String toString() {
         @SuppressWarnings("ConstantConditions") // IntelliJ can't understand the custom validators
         public static QualifiedTypeName parse(String qualifiedType) {
             var qualifiedTypeParts = qualifiedType.split(ParentSeparatorRegex, -1); // -1 avoids dropping trailing empty strings
-            require(qp -> qp.length == 1 || qp.length == 2, qualifiedTypeParts,
+            require(qp -> qp.length > 0, qualifiedTypeParts,
                     () -> format("expected qualified type to have 1 or 2 parts, split by '%s', got '%s'",
```
Frassle, line 187:
```suggestion
                    () -> format("expected qualified type to not be empty, split by '%s', got '%s'",
```

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -190,20 +192,29 @@ public static QualifiedTypeName parse(String qualifiedType) {
...
             );
             final Optional<String> parent;
-            if (qualifiedTypeParts.length == 2) {
-                var pt = qualifiedTypeParts[0];
+
+            // Handle parents. URNs can have nested parents which should be joined with a "$".
+            if (qualifiedTypeParts.length >= 2) {
+                // Remove the last qualifiedTypeParts element as it's the type.
+                String[] parentList = new String[qualifiedTypeParts.length - 1];
+                var pt = String.join(ParentSeparator, parentList);
```
Frassle, line 200: parentList is empty at this point?

tests/examples/minimal/pom.xml
```diff
@@ -75,6 +75,14 @@
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>com.google.errorprone</groupId>
+            <artifactId>error_prone_annotations</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>io.grpc</groupId>
```
t0yv0, line 83: Ideas why this is needed?

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
dixler: correct.

tests/examples/minimal/pom.xml
dixler: Thanks for the review. There's a git commit message detailing this, but tl;dr it fixes the minimal example test.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
Frassle: So pt = ""

dixler: It turned out that it was a list of length - 1 nulls so the join was nonNull and nonEmpty so it was going past the require, but the overall value was overwritten so the rest of the logic worked.

Fixed it.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -181,7 +181,7 @@ public String toString() {
         @SuppressWarnings("ConstantConditions") // IntelliJ can't understand the custom validators
         public static QualifiedTypeName parse(String qualifiedType) {
             var qualifiedTypeParts = qualifiedType.split(ParentSeparatorRegex, -1); // -1 avoids dropping trailing empty strings
-            require(qp -> qp.length == 1 || qp.length == 2, qualifiedTypeParts,
+            require(qp -> qp.length > 0, qualifiedTypeParts,
                     () -> format("expected qualified type to have 1 or 2 parts, split by '%s', got '%s'",
```
abhinav, line 185: The require failure message is now outdated. Please update.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
```diff
@@ -190,20 +190,28 @@ public static QualifiedTypeName parse(String qualifiedType) {
...
-            if (qualifiedTypeParts.length == 2) {
-                var pt = qualifiedTypeParts[0];
+
+            // Handle parents. URNs can have nested parents which should be joined with a "$".
+            if (qualifiedTypeParts.length >= 2) {
+                // Remove the last qualifiedTypeParts element as it's the type.
+                String[] parentList = new String[qualifiedTypeParts.length - 1];
+
+                // Get only the parent parts of the URN type.
+                System.arraycopy(qualifiedTypeParts, 0, parentList, 0, parentList.length);
```
abhinav, line 200: I think you can shorten this with [Arrays.copyOf](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#copyOf-T:A-int-).

```java
var parentList = Arrays.copyOf(qualifiedTypeParts, parentList.length-1)
```

Although seeing that this is just fed into String.join, which accepts an Iterable, you don't need to copy at all. You can create a List view of the array (without copying), and then get a sublist of that:

```java
List<String> parentList = Arrays.asList(qualifiedTypeParts).subList(0, qualifiedTypeParts.length - 1);
var pt = String.join(ParentSeparator, parentList)
```

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Urn.java
abhinav: Consider clarifying with a comment here that parent is the full "all parents" string, and not just the last parent.
We should consider renaming this to parents if that won't break any callers. (This doesn't look like it's exposed so it should be fine?)

tests/examples/minimal/pom.xml
t0yv0: In that case this might belong in a separate PR? Is it failing on master or why did it start failing with the changes? The minimal example test is doing double-duty as a canary for misaligned dependencies, and I'm worried this change papers over that. Misaligned dependencies sometimes result in funny breakage for users depending on their maven import order of pulumi dips.

 I can fill you in on the back-story here.

t0yv0 (changes requested):
"Fix example test minimal" is incorrect as it papers over a dependency convergence error. 

Can we get grpc-stub 1.54.0 and grpc-core 1.52.0 to line up please instead so that dependency convergence checking continues to cover any discrepancies we might have.

Also it does not belong in this PR it seems like.

tests/examples/minimal/pom.xml
dixler: I'll paper-over the commit, but leave it in the history before squashing so the details are still in this PR.

t0yv0 (approved):

Issue #1058, opened 2023-04-12
test broken on main branch: example(minimal)


dixler:
### What happened?

A test is broken on main.

This was noticed in https://github.com/pulumi/pulumi-java/pull/1056, but was not appropriate to address.

### Expected Behavior

Not broken in main.

### Steps to reproduce

Run tests

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1064, opened 2023-04-14
Fix minimal example

Frassle:
# Description

Grpc version was bumped in https://github.com/pulumi/pulumi-java/commit/796aa1eda3c26b1c7cf7fbd44f3b4708169897f1, but the minimal example wasn't updated.

Fixes https://github.com/pulumi/pulumi-java/issues/1058.

## Checklist


abhinav (approved):

t0yv0 (approved):
Thanks a ton!

PR #1068, opened 2023-04-18
Add zip64 option to javadoc jar task due to size of pulumi-aws

thomas11:
This fixes [this failure to publish the Java SDK of pulumi-aws](https://github.com/pulumi/pulumi-aws/actions/runs/4726597414/jobs/8388469282).

Not compatible with Java versions earlier than 7 which should be fine. [Docs are here](https://docs.gradle.org/7.4.1/dsl/org.gradle.jvm.tasks.Jar.html#org.gradle.jvm.tasks.Jar:zip64).


justinvp (approved):
LGTM

t0yv0 (approved):

PR #1090, opened 2023-05-04
Increase memory & zip file count limits

### LINKED ISSUE #1087: :jar fails for large SDKs
When attempting to package the Java SDK for azure-native v2 the build fails with the following message:

```
> Task :jar FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':jar'.
> archive contains more than 65535 entries.
```

Is there any way to get around this problem by reducing the number of files or increasing the maximum number of files in a package?

danielrbradley:
# Description

1. Increase main compile step max memory to 16g. This fails on anything less for azure-native.
2. Enable zip64 for the :jar task

Fixes #1087 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,5 @@
 
 ### Bug Fixes
 
+  - [sdk] Enable Zip64 jar files for jar task
+  - [sdk] Increase maximum memory for compile step to 16g


pkg/codegen/java/build.gradle.template
```diff
@@ -151,6 +151,10 @@ javadoc {
     options.jFlags("-Xmx8g", "-Xms512m")
 }
 
+jar {
+    zip64 = true
```
t0yv0, line 155: This probably benefits all the providers right so that's great!

pkg/codegen/java/build.gradle.template
```diff
@@ -38,7 +38,7 @@ java {
 
 compileJava {
     options.fork = true
-    options.forkOptions.jvmArgs.addAll(["-Xmx4g"])
+    options.forkOptions.jvmArgs.addAll(["-Xmx16g"])
```
t0yv0, line 41: Tested on GHA CI? I wonder if for smaller providers this causes OOM somewhere. Could use an env var. But can also merge and then edit later as appropriate.

RobbieMcKinstry (approved):

RobbieMcKinstry:
Looks like you need to regenerate the coegen changes to get CI to pass. I expect that's `PULUMI_ACCEPT=true` and rerunning tests.

danielrbradley:
> Looks like you need to regenerate the coegen changes to get CI to pass. I expect that's PULUMI_ACCEPT=true and rerunning tests.

Couldn't see anything in the contributing docs on which targets are required to run before CI so just ran `make`. Will take a read of the makefile and CI and see if I can figure it out if in doubt.

justinvp:
`make codegen_tests_update` should do it.

PR #1141, opened 2023-06-26
Refactor config

pawelprazak:
# Description

A small step towards Automation API implementation, that reduces the complexity and tight coupling.

Public API changes:
- add missing `getSecretDouble`, `requireDouble` and `requireSecretDouble` methods for completeness since we already provide `getDouble`
- `com.pulumi.Config` changed from a `class` to an `interface` without breaking compile time backward compatibility 

Internal changes:
- Merge two internal config implementations into one:
   - split `com.pulumi.Config` class into `Config` interface and `ConfigInternal` implementation
   - move `DeploymentImpl.Config` code into `ConfigInternal` class
- remove unused `DeploymentImpl.Config#setConfig` and `DeploymentImpl.Config#setAllConfig`
- add `Codegen#config` to be used in generated code instead of `Config.of` (changes to the codgen are out of scope for this PR)
- simplify the dependency between Config and Deployment

Fixes small part of #626

## Checklist

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceInternal.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceInternal.java
@@ -20,17 +20,21 @@ import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * Metadata of the deployment that is currently running. Accessible via @see {@link com.pulumi.deployment.Deployment#getInstance()}.
  */
 @InternalUse
 public final class DeploymentInstanceInternal implements DeploymentInstance {
 
+    private final ConfigInternal config;
     private final DeploymentInternal deployment;
 
     @InternalUse
-    public DeploymentInstanceInternal(DeploymentInternal deployment) {
-        this.deployment = deployment;
+    public DeploymentInstanceInternal(ConfigInternal config, DeploymentInternal deployment) {
+        this.config = requireNonNull(config);
+        this.deployment = requireNonNull(deployment);
     }
 
     @InternalUse
@@ -57,7 +61,7 @@ public final class DeploymentInstanceInternal implements DeploymentInstance {
 
     @Override
     public ConfigInternal getConfig() {
-        return deployment.getConfig();
+        return config;
     }
 
     @Override
--- a/sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
@@ -36,7 +36,6 @@ import javax.annotation.Nullable;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -231,7 +230,7 @@ public class PulumiTestInternal extends PulumiInternal implements PulumiTest {
         @Nullable
         private DeploymentImpl.DeploymentState state;
         @Nullable
-        private BiFunction<ConfigInternal, DeploymentImpl.DeploymentState, DeploymentInternal> deploymentFactory;
+        private Function<DeploymentImpl.DeploymentState, DeploymentInternal> deploymentFactory;
         @Nullable
         private Function<MockMonitor, MockMonitor> monitorDecorator;
 
@@ -301,7 +300,7 @@ public class PulumiTestInternal extends PulumiInternal implements PulumiTest {
          * @return this Pulumi test {@link Builder}
          */
         public Builder deploymentFactory(
-                BiFunction<ConfigInternal, DeploymentImpl.DeploymentState, DeploymentInternal> deploymentFactory
+                Function<DeploymentImpl.DeploymentState, DeploymentInternal> deploymentFactory
         ) {
             this.deploymentFactory = requireNonNull(deploymentFactory);
             return this;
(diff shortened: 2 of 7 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -424,19 +424,4 @@ public static Log mockLog(Logger logger) {
     public static Log mockLog(Logger logger, Supplier<Engine> engine) {
         return new Log(new DeploymentImpl.DefaultEngineLogger(logger, MockRunner::new, engine));
     }
-
-    @InternalUse
-    public static DeploymentImpl.Config config(ImmutableMap<String, String> allConfig, ImmutableSet<String> configSecretKeys) {
```
pawelprazak, line 429: those methods are no longer needed thanks to simplified Config implementation

sdk/java/pulumi/src/main/java/com/pulumi/test/internal/PulumiTestInternal.java
```diff
@@ -271,8 +272,8 @@ public PulumiTestInternal.Builder standardLogger(Logger standardLogger) {
          * @return this {@link PulumiTestInternal.Builder}
          */
         public PulumiTestInternal.Builder config(Map<String, String> allConfig, Set<String> configSecretKeys) {
-            return internalConfig(new DeploymentImpl.Config(
-                    ImmutableMap.copyOf(allConfig), ImmutableSet.copyOf(configSecretKeys)
+            return internalConfig(name -> new ConfigInternal(
```
pawelprazak, line 275: I needed to introduce the `configFactory` because, the project name is available only after it is initialized in `build()` - it is an internal API change

sdk/java/pulumi/src/main/java/com/pulumi/Config.java
```diff
@@ -69,230 +42,143 @@ public static Config of(String name) {
...
 
     /**
      * Loads an optional configuration value, as a number, by its key, marking it as a secret
      * or null if it doesn't exist.
      * If the configuration value isn't a legal number, this function will throw an error.
      */
-    public Output<Optional<Integer>> getSecretInteger(String key) {
-        return Output.ofSecret(getInteger(key));
-    }
+    Output<Optional<Double>> getSecretDouble(String key);
```
pawelprazak, line 95: add missing `getSecretDouble` method since we already provide `getDouble`

sdk/java/pulumi/src/main/java/com/pulumi/Config.java
```diff
@@ -69,230 +42,143 @@ public static Config of(String name) {
...
-    public Output<Integer> requireSecretInteger(String key) {
-        return Output.ofSecret(requireInteger(key));
-    }
+    Output<Integer> requireSecretInteger(String key);
+
+    /**
+     * Loads a configuration value, as a number, by its given key. If it doesn't exist, or the
+     * configuration value is not a legal number, an error is thrown.
+     */
+    double requireDouble(String key);
```
pawelprazak, line 162: add missing `requireDouble ` method since we already provide `getDouble`

sdk/java/pulumi/src/main/java/com/pulumi/Config.java
```diff
@@ -69,230 +42,143 @@ public static Config of(String name) {
...
+
+    /**
+     * Loads a configuration value, as a number, by its given key, marking it as a secret.
+     * If it doesn't exist, or the configuration value is not a legal number, an error is thrown.
+     */
+    Output<Double> requireSecretDouble(String key);
```
pawelprazak, line 168: add missing `requireSecretDouble ` method since we already provide `getDouble`

sdk/java/pulumi/src/main/java/com/pulumi/context/internal/ConfigContextInternal.java
```diff
@@ -3,31 +3,29 @@
...
+import com.pulumi.internal.ConfigInternal;
 
 import javax.annotation.ParametersAreNonnullByDefault;
-import java.util.function.Function;
 
 import static java.util.Objects.requireNonNull;
 
 @InternalUse
 @ParametersAreNonnullByDefault
 public class ConfigContextInternal implements ConfigContext {
```
pawelprazak, line 14: we can simplify this class now, the original implementation was a workaround in the anticipation of the changes done in this PR

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/Codegen.java
```diff
@@ -53,6 +54,10 @@ public static <T, O extends Output<T>> Output<T> ofNullable(@Nullable O value) {
         return value == null ? Codegen.empty() : value;
     }
 
+    public static Config config(String name) {
```
pawelprazak, line 57: to make sure we cleanly separate the internal codegen API, the config factory method should be moved here IMHO

changes to the codegen are out of scope for this PR

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceInternal.java
```diff
@@ -19,17 +20,21 @@
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 
+import static java.util.Objects.requireNonNull;
+
 /**
  * Metadata of the deployment that is currently running. Accessible via @see {@link com.pulumi.deployment.Deployment#getInstance()}.
  */
 @InternalUse
 public final class DeploymentInstanceInternal implements DeploymentInstance {
```
pawelprazak, line 29: moving `ConfigInternal` here gives us less tight coupling and makes future development easier and the implementations is made simpler

IMHO the `DeploymentInstanceInternal` should be renamed `GlobalContext` and separated from the `Deployment` in future refactoring to clearly state its role in the architecture, but this is just a remark and remains out of scope for this PR.

PR #1228, opened 2023-09-27
[sdk-gen] Use specialized missing required property exception

### LINKED ISSUE #1199: Java SDK panics when missing required property
### What happened?

See https://github.com/pulumi/pulumi-rke/issues/129 for the user facing issue. Basically, the user didn't provide a required field, and they got a `java.lang.NullPointerException: expected parameter 'user' to be non-null`. This exception sounds like we have dereferenced a null pointer (which I assume we have).

### Expected Behavior

We don't throw `java.lang.NullPointerException`s. If the user failed to supply a required input, we should throw a Pulumi exception: `com.pulumi.MissingRequiredProperty`[^1] that clearly indicates that this is a user problem and how they should address it. 

Missing arguments should never result in a null pointer exception.

[^1]: I'm not saying it should be that name in particular, but it should definitely be from the Pulumi namespace, not `java.lang`.

### Steps to reproduce

See https://github.com/pulumi/pulumi-rke/issues/129

### Output of `pulumi about`

```
CLI          
Version      3.78.1
Go Version   go1.20.7
Go Compiler  gc

Plugins
NAME  VERSION
java  unknown

Host
OS       Microsoft Windows 11 Enterprise
Version  10.0.22000 Build 22000
Arch     x86_64

This project is written in java: executable='' version='openjdk 19 2022-09-20
OpenJDK Runtime Environment (build 19+36-2238)
OpenJDK 64-Bit Server VM (build 19+36-2238, mixed mode, sharing)' java='' javac='19'

Current Stack: organization/infrastructure/dev

TYPE                                               URN
pulumi:pulumi:Stack                                urn:pulumi:dev::infrastructure::pulumi:pulumi:Stack::infrastructure-dev
pulumi:providers:docker                            urn:pulumi:dev::infrastructure::pulumi:providers:docker::default_4_3_1
docker:index/remoteImage:RemoteImage               urn:pulumi:dev::infrastructure::docker:index/remoteImage:RemoteImage::ubuntuRemoteImage
com.myproject.infrastructure:core:cluster:Rancher  urn:pulumi:dev::infrastructure::com.myproject.infrastructure:core:cluster:Rancher::RancherDeployment
pulumi:providers:rke                               urn:pulumi:dev::infrastructure::pulumi:providers:rke::default_3_3_0
rke:index/cluster:Cluster                          urn:pulumi:dev::infrastructure::com.myproject.infrastructure:core:cluster:Rancher$rke:index/cluster:Cluster::rancher-cluster
```

### Additional context

_No response_

### Contributing

Zaid-Ajaj:
# Description

This PR adds a new custom exception `MissingRequiredPropertyException` to the main Java SDK. Then SDK-gen uses this type and throws it as an error when an input property is required but isn't specified. This fixes #1199 where previously the users would receive a `NullPointerException` instead.

Due to using Go templates and conditionals as follows:
```
{{ if $setter.IsRequired }} 
...
{{ end }}
```
The generated code emits an empty line when `IsRequired` is `false` which is why you will see the new line diffs in the commits of this PR

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
 
 - Generated SDKs use specialized `MissingRequiredPropertyException` rather than throwing `NullPointerException` when a required input property is missing.
+
+### Bug Fixes
--- a/sdk/java/pulumi/src/main/java/com/pulumi/test/internal/MockMonitor.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/test/internal/MockMonitor.java
@@ -55,7 +55,7 @@ public class MockMonitor implements Monitor {
 
     @Override
     public CompletableFuture<SupportsFeatureResponse> supportsFeatureAsync(SupportsFeatureRequest request) {
-        var hasSupport = "secrets".equals(request.getId()) || "resourceReferences".equals(request.getId());
+        var hasSupport = !"outputValues".equals(request.getId());
         return CompletableFuture.completedFuture(
                 SupportsFeatureResponse.newBuilder().setHasSupport(hasSupport).build()
         );


pkg/codegen/java/gen.go
```diff
@@ -599,6 +608,15 @@ func (pt *plainType) genInputType(ctx *classFileContext) error {
 	for propIndex, prop := range pt.properties {
 		propType := propTypes[propIndex]
 		fieldName := names.Ident(pt.mod.propertyName(prop)).AsProperty().Field()
+
+		if prop.IsRequired() && prop.DefaultValue == nil && prop.ConstValue == nil {
+			fprintf(w, "            if ($.%s == null) {\n", fieldName)
+			fprintf(w, "                throw new %s(\"%s\", \"%s\");\n",
+				ctx.ref(names.PulumiMissingRequiredPropertyException), pt.name, fieldName)
```
justinvp, line 615: Are we going to need to first ship the exception in the core Java SDK, and then somehow have the codegen indicate that version of the core Java SDK is now required (since it depends on this new exception being there)? I'm not familiar enough with Java on how to indicate that requirement.

Frassle: Yes we do.

Zaid-Ajaj: I thought it should be fine because if it is merged and tag a release, a new sdk is published that includes the new exception type, sometime later providers team will pick up the newer sdk-gen from pulumi-java and start using that new type. This is assuming that sdk-gen references latest pulumi-java sdk

Frassle: sdk-gen dictates the version of the core library to use. So we need to release the type first, then make sure sdk-gen is setting that version as the required version when emiting the provider sdks.

Frassle: i.e. "This is assuming that sdk-gen references latest pulumi-java sdk" is the assumption that doesn't hold. e.g. [this line](https://github.com/pulumi/pulumi-aws/blob/master/sdk/java/build.gradle#L46) which will be set from in https://github.com/pulumi/pulumi-java/blob/main/pkg/codegen/java/templates_gradle.go#L87 somehow.

Zaid-Ajaj: > i.e. "This is assuming that sdk-gen references latest pulumi-java sdk" is the assumption that doesn't hold.

The `pulumi-java-gen` CLI used to generate java SDKs takes the version of the pulumi-java main sdk as input but it defaults to the version of the `pulumi-java-gen` itself, see [here](https://github.com/pulumi/pulumi-java/blob/main/pkg/cmd/pulumi-java-gen/command.go#L91-L92):
```go
cmd.Flags().StringVar(&javaSdkVersionArg, "sdk", version.Version,
	"com.pulumi:pulumi version to depend on in the generated code")
```
Doesn't that mean if we tag 0.9.x which publishes both SDK and generator, that the generated SDKs will be using 0.9.x as well? 🤔

Frassle: What about from the Pulumi cli where we call the gen methods directly?

Zaid-Ajaj: Tested what Pulumi CLI does when you run:
```
pulumi package gen-sdk schema.json --language java --out sdk
```
and it generates just java source files without a `gradle` project file 😱 so basically gen-sdk isn't usable with java projects

Frassle: 🤦‍♂️ alright, well we'll need to look into that at some point but means nothing to worry about for this PR

Frassle (approved):

PR #1231, opened 2023-10-03
Add build.gradle.kts to Gradle executor detection

### LINKED ISSUE #846: Gradle executor is not detected if only build.gradle.kts file is present
https://github.com/pulumi/pulumi-java/blob/af0e43990042b9649a8f04a4aed7902bce8e8212/pkg/internal/executors/executor_gradle.go#L88-L92

The code above is used to detect an executor, and if I have only `build.gradle.kts` file in my project, then it falls through to `sbt` executor

Also, this function might be worth looking at as well:
https://github.com/pulumi/pulumi-java/blob/af0e43990042b9649a8f04a4aed7902bce8e8212/pkg/internal/executors/executor_gradle.go#L58-L62

josephglanville:
# Description

Most modern Gradle projects use the Gradle Kotlin DSL.
This change is necessary to detect projects as it's not possible
to add one of the existing filetypes without disrupting the build
configuration, especially in multi-project Gradle builds.

Fixes #846 

## Checklist


josephglanville:
@Zaid-Ajaj could you help me get this reviewed?

justinvp (approved):
LGTM. Thanks for the improvement and for your patience!

PR #1233, opened 2023-10-05
[sdk-gen] Fix local variable name collisions in builder() implementation and character encoding option

### LINKED ISSUE #1232: Generated code does not compile
### What happened?

The Akamai upgrade (https://github.com/pulumi/pulumi-akamai/pull/286) generated code that does not compile.

### Example

NA

### Output of `pulumi about`

This is on `pulumi-java-gen v0.9.5`.

### Additional context

_No response_

### Contributing

Zaid-Ajaj:
### Description

Fixes #1232

The issue with compiling the generated akamai code was two fold:
 - Some generated comments included characters that were not ut8-encoded so I added `options.encoding="UTF-8"` to the `compileJava` block to fix that
 - The implementation of `builder.build()` uses a local variable `o` of the result type, then fills each field of that value. However, it fails to compile when the result type has a field named `o` so they collide. Fixed this by changing the name of the local variable from `o` to `_resultValue`

### Checklist


Frassle (approved):
> +554,915 −993

Welp LGTM 😆 

More seriously the changes to pkg/codegen/java look sane.

PR #1308, opened 2024-01-11
Don't JSON-encode provider config

blampe:
See https://github.com/pulumi/pulumi/pull/15032 for context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+### Improvements
+
+- Provider configuration is no longer serialized to JSON. This improves secret management for providers and does not
+  impact public APIs. [#1308](https://github.com/pulumi/pulumi-java/pull/1308)
+
 ## 0.9.9 (2023-12-27)
 
 ### Improvements


pkg/codegen/testing/test/testdata/akamai/java/src/main/java/com/pulumi/akamai/ProviderArgs.java
```diff
@@ -18,14 +18,14 @@ public final class ProviderArgs extends com.pulumi.resources.ResourceArgs {
 
     public static final ProviderArgs Empty = new ProviderArgs();
 
-    @Import(name="cacheEnabled", json=true)
```
t0yv0, line 21: Reminder here to remove `json` from the ImportAnnotation or at least mark deprecated.

EronWright:
Should this code in kubernetes provider be changed? It synthesizes the `@import` annotation for custom resource properties.

https://github.com/pulumi/pulumi-kubernetes/blob/e49daf0ebf8b9a30de8853c363781166e687d3e7/sdk/java/src/main/java/com/pulumi/kubernetes/apiextensions/Util.java#L46-L50

t0yv0:
@EronWright I think the CustomResource feature introduced in:

https://github.com/pulumi/pulumi-kubernetes/pull/3020/files#diff-842f1f409f5a9ae544bb27542564a1e90ccab80a818faa702731ac2048395e11

Is not actually affected, afaik, correct me if I am wrong it only affects custom resources that are not providers? The JSON encoding is only affecting explicit providers (and, as we found out, default providers), but I'm hoping it has no interaction with #3020. 

Happy to discuss further.

t0yv0:
We will need to add a flag here for a flagged rollout and resurrect this change.

PR #1342, opened 2024-04-26
Escape javadoc special characters

### LINKED ISSUE #1271: Ignore or escape unintended javadoc comments
## Hello!

## Issue details

Pulumi-GCP v7.1.0 introduced what we believe is a spurious extra string to the description of a few resources. This string started with `@pattern: ` and was populated to the documentation sections in the Pulumi schema and picked up by all related resources
see [Csharp example](https://github.com/pulumi/pulumi-gcp/blob/v7.1.0/sdk/dotnet/Compute/ForwardingRule.cs#L370) in all SDKs.

All SDKs build successfully and passed integration tests, so the upgrade was merged and released.

However, during the SDK publish step, the [Java publish failed](https://github.com/pulumi/pulumi-gcp/actions/runs/6881931301/job/18728079843#step:3:2872) because Javadoc was interpreting this spurious string as a doc tag.

It would be fantastic if we could allow only a subset of expected Javadoc tags, and escape anything not on this allowlist.

### Affected area/feature

SDK publication

iwahbe:
# Description

Fixes #1271 

Correctly scope comment escapes to outside of code blocks. The resulting code is both more idiomatic javadoc and safe to embed.

I have two motivating examples:

1. Correctness: Upstream comments that start lines with `@pattern` will now be escaped to `{@literal @}pattern`. This has prevented us from publishing in the past (https://github.com/pulumi/pulumi-gcp/issues/1950, https://github.com/pulumi/pulumi-auth0/issues/516, https://github.com/pulumi/pulumi-azure/issues/1979)

2. Idiomaticity: Generated code blocks will no longer be HTML escaped (since they don't need to be). Instead, they will render correctly.

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/pkg/codegen/java/utilities.go
+++ b/pkg/codegen/java/utilities.go
@@ -11,23 +11,94 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
 )
 
-func formatBlockComment(comment string, indent string) string {
-	prefix := fmt.Sprintf("%s * ", indent)
-	comment = strings.ReplaceAll(comment, "*/", "*{@literal /}")
+// formatForeignBlockComment formats comment as a java block comment for use in the
+// generated SDK. formatForeignBlockComment is responsible for ensuring that comment is
+// safe to embed in the generated SDK. That means that comment cannot contain any codes
+// that escape javadoc or the embedding comment.
+func formatForeignBlockComment(comment, indent string) string {
+	return formatForeignBlockCommentFrom(comment, 0, indent)
+}
+
+// formatForeignBlockCommentFrom is like [formatForeignBlockComment], except that it
+// unconditionally accepts the first idx bytes of comment.
+func formatForeignBlockCommentFrom(comment string, idx int, indent string) string {
 	comment = codegen.FilterExamples(comment, "java")
 
-	// TODO it seems that inputs sometimes have at least `@(`
-	// inside java code examples. This quick fix targets it
-	// directly to escape `@` for Javadoc. Some refactoring is
-	// needed around javadoc emission: would be better to work on
-	// AST as inside codegen.FilterExamples so that the escaping
-	// is complete and applies only to comments, while not
-	// touching intentional Javadoc directives emitted such as
-	// `@return`.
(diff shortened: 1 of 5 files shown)


pkg/codegen/java/utilities.go
```diff
@@ -11,23 +11,94 @@ import (
...
-func formatBlockComment(comment string, indent string) string {
+// formatForeignBlockComment formats comment as a java block comment for use in the
+// generated SDK. formatForeignBlockComment is responsible for ensuring that comment is
+// safe to embed in the generated SDK. That means that comment cannot contain any codes
+// that escape javadoc or the embedding comment.
+func formatForeignBlockComment(comment, indent string) string {
+	return formatForeignBlockCommentFrom(comment, 0, indent)
+}
+
+// formatForeignBlockCommentFrom is like [formatForeignBlockComment], except that it
```
t0yv0, line 22: Style nit for helper functions fHelper that are only called from f and are not supposed to be used anywhere else I personally forgo the doc comments

pkg/codegen/java/utilities.go
```diff
@@ -11,23 +11,94 @@ import (
...
+
+		// Write all non-copied text proceeding the code block
+		if codeStart != 0 {
+			dst.WriteString(nonCode(comment[:codeStart]))
+		}
+
+		// Then write the code block itself
+		dst.WriteString(code(comment[codeStart:codeEndOffset]))
+		// Then adjust copiedTo to start after the code block
+		comment = comment[codeEndOffset:]
```
t0yv0, line 81: Old-fashioned scanner FTW.

t0yv0 (approved):
Thank you for fixing this!

pkg/codegen/java/utilities.go
iwahbe: It's not a helper function. It's called directly from the rest of the module. It's used for generating comments like:
```go
fprintf(w, formatForeignBlockCommentFrom("@return "+invoke.Comment, 1))
```

That way any `@` in the `invoke.Comment` string is escaped but the first `@` of the `@return` is left alone.

PR #1344, opened 2024-04-30
Fix emitting `PANIC` in generated programs when we don't have access to the schema of resource properties

### LINKED ISSUE #1341: `PANIC` generated into public-facing documentation
### What happened?

Currently as spotted by @VenelinMartinov RandomInteger docs generate a PANIC into Java code, where other languages work fine. This is using `pulumi convert` to perform the translation. 

### Example

```hcl

# The following example shows how to generate a random priority
# between 1 and 50000 for a aws_alb_listener_rule resource:

resource "random_integer" "priority" {
  min = 1
  max = 50000
  keepers = {
    # Generate a new integer each time we switch to a new listener ARN
    listener_arn = var.listener_arn
  }
}

resource "aws_alb_listener_rule" "main" {
  listener_arn = random_integer.priority.keepers.listener_arn
  priority     = random_integer.priority.result

  action {
    type             = "forward"
    target_group_arn = var.target_group_arn
  }
  # ... (other aws_alb_listener_rule arguments) ...
}

```

Getting this rendered:

```java

package generated_program;

import com.pulumi.Context;
import com.pulumi.Pulumi;
import com.pulumi.random.RandomInteger;
import com.pulumi.random.RandomIntegerArgs;
import com.pulumi.aws.albListenerRule;
import com.pulumi.aws.AlbListenerRuleArgs;
import java.util.List;
import java.util.ArrayList;
import java.util.Map;
import java.io.File;
import java.nio.file.Files;
import java.nio.file.Paths;

public class App {
        Pulumi.run(App::stack);
    }

        // The following example shows how to generate a random priority
        // between 1 and 50000 for a aws_alb_listener_rule resource:
        var priority = new RandomInteger("priority", RandomIntegerArgs.builder()        
            .min(1)
            .max(50000)
            .keepers(Map.of("listener_arn", listenerArn))
            .build());

        var main = new AlbListenerRule("main", AlbListenerRuleArgs.builder()        
            .listenerArn(priority.keepers().applyValue(keepers -> keepers.listenerArn()))
            .priority(priority.result())
            .action(%!v(PANIC=Format method: runtime error: invalid memory address or nil pointer dereference))
            .build());

    }
}

```

Python translation:

```python
import pulumi
import pulumi_aws as aws
import pulumi_random as random

# The following example shows how to generate a random priority
# between 1 and 50000 for a aws_alb_listener_rule resource:
priority = random.RandomInteger("priority",
    min=1,
    max=50000,
    keepers={
        "listener_arn": listener_arn,
    })
main = aws.index.AlbListenerRule("main",
    listener_arn=priority.keepers.listener_arn,
    priority=priority.result,
    action=[{
        type: forward,
        targetGroupArn: target_group_arn,
    }])

```

### Output of `pulumi about`

N/A

### Additional context

N/A

### Contributing

Zaid-Ajaj:
# Description

Fixes #1341

The problem here is when we don't have access to a pulumi schema during program-gen, then `currentResourcePropertyType` from the `generator` type is `nil`. Then at some point we ask for `typeName(g.currentResourcePropertyType)` and call `.String()` on the schema type which is `nil` causing a `PANIC` to be generated. This PR simply makes sure that `currentResourcePropertyType` is initialized to a non-nil value to avoid the panic. 

That is only half of the problem of #1341 though, because we _should_ have access to the schema when generating examples in the docs. That is however not something we can fix from the point of view of pulumi-java

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,4 @@
 
 
 ### Bug Fixes
+ - Fix emitting `PANIC` in generated programs when we don't have access to the schema of resource properties
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -184,8 +184,9 @@ func GenerateProgram(program *pcl.Program) (map[string][]byte, hcl.Diagnostics, 
 	}
 
 	g := &generator{
-		program:         program,
-		functionInvokes: map[string]*schema.Function{},
+		program:                     program,
+		functionInvokes:             map[string]*schema.Function{},
+		currentResourcePropertyType: schema.AnyType,
 	}
 
 	g.Formatter = format.NewFormatter(g)


Frassle (approved):
Looks reasonable. Can we add a conformance tests in pu/pu for this? I know Java isn't running conformance tests yet but this sounds like a valid use case that we should make sure we're checking for all the languages.

t0yv0:
This looks reasonable. I added a lot more tickets today around these PANICS they do not all have the same root cause it appears. 

For the root cause that "I don't know the schema here" the provider build use case would actually be lovely to error out with exactly that error message, so we know that we missed a reference to a provider in our build environment.

In addition I seem to hit cases like the one recently fixed in pulumi-random where the AWS provider is referenced in the build environment but at the wrong version (v5 vs v6), so some of the schemas for resources do not load. That'd be nice to know too..

Having actionable warnings is a nice to have of course, for now I just would love to find a way to close out all the actual AWS PANIC issues.

lunaris:
After digging into the issues that Anton raised, I think that the correct fix for several of them is to error when the schema lookup fails, rather than trying to continue and avoid the panic. Specifically, for:

* #1346
* #1347 
* #1348
* #1350
* #1351 
* #1352 
* #1353
* #1355 

the examples are incorrect/contain mis-named attributes, which cause the schema lookup failures. This appears to be the approach that other languages take (and indeed, looking at the other language conversions for the examples Anton gave, many fail due to the examples being incorrect). As a result I'm going to raise another PR and recommend we close this one.

PR #1356, opened 2024-05-01
"Escape javadoc special characters" V2

iwahbe:
Reverts pulumi/pulumi-java#1345

Fixes https://github.com/pulumi/pulumi-java/issues/1271


iwahbe:
I have tested and confirmed that this works for pulumi-azure.

thomas11 (approved):
Not easy to review but looks reasonable.

t0yv0 (approved):
Thank you for a quick turnaround!

PR #1379, opened 2024-06-24
Update pulumi/pulumi to v3.121.0

julienp:
Update to pulumi/pulumi to 3.121 and add a noop `RuntimeOptionsPrompts`.

### DIFF
--- a/pkg/codegen/java/gen_test.go
+++ b/pkg/codegen/java/gen_test.go
@@ -180,6 +180,10 @@ func adaptTest(t *test.SDKTest, keyDeps map[string]string) generatePackageTestCo
 		t.Skip = codegen.NewStringSet("java/any") // python-only
 	case "array-of-enum-map":
 		t.Skip = codegen.NewStringSet("java/any") // go-only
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -441,6 +441,12 @@ func (host *javaLanguageHost) GetProgramDependencies(
 	return &pulumirpc.GetProgramDependenciesResponse{}, nil
 }
 
+func (host *javaLanguageHost) RuntimeOptionsPrompts(ctx context.Context,
```
lunaris, line 444: If neither are used we can just do `(context.Context, *pulumirpc.RuntimeOptionsRequest)`

lunaris (approved):

pkg/cmd/pulumi-language-java/main.go
julienp: Formatter ended up complaining at me when I had the named params and told me to use `_`. Since the rest of the file also uses `_` I kept that instead of going to no name at all.

PR #1384, opened 2024-07-04
Exclude Non-Loadable Packages from ClassLoader. Fixes #issue 1385

### LINKED ISSUE #1385: ClassLoader loads all packages presented in pom
### What happened?

We used a packages wihich contains other not used packages, so these packages is not loaded by maven and therefore  the class loader has a problem to load it. This will some out as a error on packages problem:
    SEVERE: Running program [PID: 13236](unknown) failed with an unhandled exception:
    java.lang.IllegalStateException: Failed to load class 'org.apache.commons.compress.harmony.pack200.NewAttribute$ErrorAttribute' (package: 'org.apache.commons.compress.harmony.pack200') from class path: org/objectweb/asm/Attribute

### Example

It is pom.xml excluding a package. We would like to only de it sometimes, because we have multiple class providers etc.   
 <dependency>
      <groupId>com.github.docker-java</groupId>
      <artifactId>docker-java-core</artifactId>
      <exclusions>
        <exclusion>
          <groupId>org.apache.commons</groupId>
          <artifactId>commons-compress</artifactId>
        </exclusion>
      </exclusions>
    </dependency>

### Output of `pulumi about`

CLI          
Version      3.112.0
Go Version   go1.22.1
Go Compiler  gc

Plugins
NAME        VERSION
aws         6.40.0
command     0.10.0
eks         2.7.1
java        unknown
kubernetes  4.13.1

Host
OS       Microsoft Windows 11 Pro
Version  10.0.22631 Build 22631
Arch     x86_64

### Additional context

_No response_

### Contributing

andreaspartisia:
# Description

We encountered an issue with the package loader that loads every package indiscriminately, instead of using the Maven class loader's smart loading. This is a known problem, and while some packages have already been excluded, the process requires a merge request for each exclusion. Therefore, we identified the need for a faster and more dynamic exclusion mechanism.

This fix introduces a system property string containing a comma-separated list of packages to exclude. By setting this property, the specified packages will be excluded from the loaded packages.

Fixes #1385

## Checklist

### DIFF
--- a/sdk/java/pulumi/build.gradle
+++ b/sdk/java/pulumi/build.gradle
@@ -121,8 +121,17 @@ sourceSets {
     }
 }
 
-test {
+task isolatedTest(type: Test) {
     useJUnitPlatform()
+
+    // Include only the specific test class you want to run in isolation
+    include '**/ResourcePackagesIsolatedTest.class'
+
+    // Fork a new JVM process for this specific test
+    forkEvery = 1
+    maxParallelForks = 1
+
+    // Set up test logging if needed
     testLogging {
         showStandardStreams = true
         exceptionFormat = 'full'
@@ -141,6 +150,36 @@ test {
 
     systemProperty 'java.util.logging.config.file', "${projectDir}/src/test/resources/logging.properties"
 }
+
+test {
+    useJUnitPlatform()
+
+    // Run general tests, excluding the isolated test
+    filter {
+        excludeTestsMatching "com.pulumi.serialization.internal.ResourcePackagesIsolatedTest"
+    }
+
+    testLogging {
+        showStandardStreams = true
+        exceptionFormat = 'full'
+
+        events = ['failed']
+
+        info {
+            events = ['failed', 'skipped']
+        }
+
+        debug {
+            events = ['started', 'skipped', 'failed']
+        }
+    }
+
+    systemProperty 'java.util.logging.config.file', "${projectDir}/src/test/resources/logging.properties"
+
+    // Ensure isolatedTest runs after the main tests
+    finalizedBy isolatedTest
+}
+
 
 def ossrhUsername = project.findProperty("ossrhUsername") ?: System.getenv("OSSRH_USERNAME")
 def ossrhPassword = project.findProperty("ossrhPassword") ?: System.getenv("OSSRH_PASSWORD")
--- a/sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/ResourcePackagesIsolatedTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/serialization/internal/ResourcePackagesIsolatedTest.java
@@ -1,0 +1,46 @@
+package com.pulumi.serialization.internal;
+
+import com.pulumi.Log;
+import org.junit.jupiter.api.Test;
+import java.util.Optional;
+import com.pulumi.core.annotations.ResourceType;
+import com.pulumi.resources.ComponentResource;
+import com.pulumi.resources.ComponentResourceOptions;
+import com.pulumi.resources.Resource;
+import com.pulumi.resources.ResourceArgs;
+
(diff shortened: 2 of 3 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/ResourcePackages.java
```diff
@@ -158,7 +158,21 @@ private static boolean excludePackages(ClassInfo c) {
                 && !c.getPackageName().equals("net.bytebuddy")
                 && !c.getPackageName().startsWith("net.bytebuddy.")
                 && !c.getPackageName().startsWith(Resource.class.getPackageName())
-                && !c.getPackageName().startsWith(EngineGrpc.class.getPackageName());
+                && !c.getPackageName().startsWith(EngineGrpc.class.getPackageName())
+                && !excludePackagesFromProperties(c);
+    }
+
+    private static boolean excludePackagesFromProperties(ClassInfo c) {
+        String packages = System.getProperty("ADDITIONAL_EXCLUDES");
```
justinvp, line 166: Should we pick a better name than `ADDITIONAL_EXCLUDES` for this? Is there a standard idiomatic naming convention that most other Java libraries typically use for system properties? For example, wondering if this should be something like `pulumi.resourcepackages.excludes` or `pulumi.classloader.excludes` or something like that?

sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/ResourcePackages.java
```diff
@@ -158,7 +158,21 @@ private static boolean excludePackages(ClassInfo c) {
...
+        if (packages != null) {
+            String[] items = packages.split(",");
+            for (String item : items){
```
justinvp, line 169: Wondering if we should trim whitespace from `item`?

justinvp (changes requested):
Thanks for the contribution! I left some review comments. This also needs a test before we can merge it.

sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/ResourcePackages.java
andreaspartisia: Yes this is a very valid point, I think pulumi.resourcepackages.excludes fit the purpose very well.

andreaspartisia:
Thanks for the review @justinvp

I have corrected it now and made a test for a the exclusion of packages.

justinvp (approved):
Looks good to me. Thanks!

@andreaspartisia, looks like there's a conflict with the `pulumi` submodule. Can you revert that change?

andreaspartisia:
Yes @justinvp it has been updated now.

andreaspartisia:
@justinvp 
Do you have time to look at this again?

justinvp:
@andreaspartisia, thanks for your patience!

I opened a temp PR to run the tests in CI and the new test is failing:

https://github.com/pulumi/pulumi-java/actions/runs/10464711824/job/28978759154?pr=1412#step:12:376

```
ResourcePackagesTest > ExcludedResource() FAILED
    java.lang.AssertionError: Test resource was found
        at com.pulumi.serialization.internal.ResourcePackagesTest.ExcludedResource(ResourcePackagesTest.java:91)
```

andreaspartisia:
Thanks for the reminder @justinvp 
It was because all test used the same JVM in parallel testing, so they polluted each other, but I have made my test run in it's own JWM, so it's fine now.
I hope you have time to look at it again.

andreaspartisia:
@justinvp do you have time to look at it again or do one of your colleagues have time for it?

justinvp (approved):
LGTM! I'll get it merged after reviewing the test run.

Note: we were running into some unrelated lint issues in CI in this repo yesterday that we may need to fix separately, so it may be another day or so. Thanks for your patience and thanks again for addressing this!

PR #1394, opened 2024-07-23
Unable to rehydrate a resource that has required inputs

### LINKED ISSUE #1393: Unable to rehydrate a resource that has required inputs
The root cause of https://github.com/pulumi/pulumi-kubernetes/issues/3057 is that some generated resource classes crash when being rehydrated due to unnecessary input validation logic.  Here's the relevant stack trace:

```
    Caused by: java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
        at com.pulumi.serialization.internal.ResourcePackages.tryConstruct(ResourcePackages.java:231)
        ... 21 more
    Caused by: com.pulumi.exceptions.MissingRequiredPropertyException: Missing required property 'roleRef when constructing 'RoleBindingArgs'
        at com.pulumi.kubernetes.rbac.v1.RoleBindingArgs$Builder.build(RoleBindingArgs.java:246)
        at com.pulumi.kubernetes.rbac.v1.RoleBinding.makeArgs(RoleBinding.java:131)
        at com.pulumi.kubernetes.rbac.v1.RoleBinding.<init>(RoleBinding.java:119)
```

Upon receiving a resource reference from the Chart component, the internal `ResourcePackages` class tries to instantiate the corresponding resource class (see [code](https://github.com/pulumi/pulumi-java/blob/0f614b60a861e9e2faa2f798036dda6fddb4dba4/sdk/java/pulumi/src/main/java/com/pulumi/serialization/internal/ResourcePackages.java#L231)). The constructor then fails because it erroneously attempts to validate the args and fails when a required input is missing. The constructor should not validate or pass args down to the engine in this mode; see [the Go SDK](https://github.com/pulumi/pulumi/blob/4516fbce89af1188bbc2f049d976e3d6b81a2666/sdk/go/pulumi/context.go#L1330-L1337) for comparison.  The basis for detecting this situation is the presence of the `Urn` option, which signals that an existing resource is being read.

This issue would affect any generated resource that meets BOTH criteria:
1. Has a constant input (e.g. `apiVersion`, `kind`), necessitating a `makeArgs` helper.
2. Has at least one required top-level input (e.g. `roleRef` in `RoleBinding` resource).

Some, but not all, of the Kubernetes resource types meet the above criteria.

For reference, the Pulumi schema for RoleBinding:
```json
        "kubernetes:rbac.authorization.k8s.io/v1:RoleBinding": {
            "description": "RoleBinding references a role, but does not contain it.  It can reference a Role in the same namespace or a ClusterRole in the global namespace. It adds who information via Subjects and namespace information by which namespace it exists in.  RoleBindings in a given namespace only have effect in that namespace.",
            "properties": {
                "apiVersion": {
                    "type": "string",
                    "description": "APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources",
                    "const": "rbac.authorization.k8s.io/v1"
                },
                "kind": {
                    "type": "string",
                    "description": "Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds",
                    "const": "RoleBinding"
                },
                "metadata": {
                    "$ref": "#/types/kubernetes:meta/v1:ObjectMeta",
                    "description": "Standard object's metadata."
                },
                "roleRef": {
                    "$ref": "#/types/kubernetes:rbac.authorization.k8s.io/v1:RoleRef",
                    "description": "RoleRef can reference a Role in the current namespace or a ClusterRole in the global namespace. If the RoleRef cannot be resolved, the Authorizer must return an error. This field is immutable."
                },
                "subjects": {
                    "type": "array",
                    "items": {
                        "$ref": "#/types/kubernetes:rbac.authorization.k8s.io/v1:Subject"
                    },
                    "description": "Subjects holds references to the objects the role applies to."
                }
            },
            "type": "object",
            "required": [
                "roleRef"
            ],
            "language": {
                "nodejs": {
                    "requiredOutputs": [
                        "apiVersion",
                        "kind",
                        "metadata",
                        "roleRef",
                        "subjects"
                    ]
                }
            }
        }
```

EronWright:
# Description

This PR makes a change to the code generator, to avoid any attempt to validate the `args` in the "read" case (e.g. rehydration of a resource reference) vs "register" case. This case is distinguished by the presence of the URN option, as evidenced [here](https://github.com/pulumi/pulumi-java/blob/59a8569adf5c1338cd1f869065c82bad3740530d/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java#L1236-L1256). Note that the args aren't used in the "read" case.

https://github.com/pulumi/pulumi-java/blob/59a8569adf5c1338cd1f869065c82bad3740530d/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java#L1236-L1237

See also, the impact on pulumi-kubernetes: https://github.com/pulumi/pulumi-kubernetes/pull/3119

Fixes #1393 

## Examples
There's two variants to consider: resources with consts and resources without consts; the latter didn't need a `makeArgs` function (until now).

### Variant 1 (with consts)
Before:
```java
    public ConfigMap(String name, @Nullable ConfigMapArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes:core/v1:ConfigMap", name, makeArgs(args), makeResourceOptions(options, Codegen.empty()));
    }

    private ConfigMap(String name, Output<String> id, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes:core/v1:ConfigMap", name, null, makeResourceOptions(options, id));
    }

    private static ConfigMapArgs makeArgs(@Nullable ConfigMapArgs args) {
        var builder = args == null ? ConfigMapArgs.builder() : ConfigMapArgs.builder(args);
        return builder
            .apiVersion("v1")
            .kind("ConfigMap")
            .build();  // exception thrown here if the resource has any required inputs
    }
```

After:
```java
    public ConfigMap(String name, @Nullable ConfigMapArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes:core/v1:ConfigMap", name, makeArgs(args, options), makeResourceOptions(options, Codegen.empty()));
    }

    private ConfigMap(String name, Output<String> id, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes:core/v1:ConfigMap", name, null, makeResourceOptions(options, id));
    }

    private static ConfigMapArgs makeArgs(@Nullable ConfigMapArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        if (options != null && options.getUrn().isPresent()) {
            return null;
        }
        var builder = args == null ? ConfigMapArgs.builder() : ConfigMapArgs.builder(args);
        return builder
            .apiVersion("v1")
            .kind("ConfigMap")
            .build();
    }
```

### Variant 2 (without consts)

Before: 
```java
    public Provider(String name, @Nullable ProviderArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes", name, args == null ? ProviderArgs.Empty : args, makeResourceOptions(options, Codegen.empty()));
    }
```

After:
```java
    public Provider(String name, @Nullable ProviderArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        super("kubernetes", name, makeArgs(args, options), makeResourceOptions(options, Codegen.empty()));
    }

    private static ProviderArgs makeArgs(@Nullable ProviderArgs args, @Nullable com.pulumi.resources.CustomResourceOptions options) {
        if (options != null && options.getUrn().isPresent()) {
            return null;
        }
        return args == null ? ProviderArgs.Empty : args;
    }

```

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
 
 ### Bug Fixes
+
+- Unable to rehydrate a resource that has required inputs


pkg/codegen/java/gen.go
```diff
@@ -1129,8 +1131,11 @@ func (mod *modContext) genResource(ctx *classFileContext, r *schema.Resource, ar
 			}
 		}
 		fprintf(w, "            .build();\n")
-		fprintf(w, "    }\n")
+	} else {
+		fprintf(w,
+			"        return args == null ? %s.Empty : args;\n", ctx.ref(argsFQN))
```
EronWright, line 1136: Not really related to this fix, but `Empty` really shouldn't be used here. In my opinion we should use the builder, so to get 'required' property validation.

EronWright:
I tried to add some unit or integration tests to exercise rehydration but I don't see a way to. I think it would involve an "echo" component that returned a resource reference. I opted to contribute a simple example based on the kubernetes provider, specifically using `ConfigGroup` (v2).

pkg/codegen/java/gen.go
lunaris: I don't disagree but wonder if this would be a breaking change -- for sure it feels like the "right" thing to do but wonder if it would upset anything.

tests/examples/java_examples_test.go
```diff
@@ -133,6 +133,14 @@ func TestExamples(t *testing.T) {
 		integration.ProgramTest(t, &test)
 	})
 
+	t.Run("kubernetes", func(t *testing.T) {
+		t.SkipNow()
```
lunaris, line 137: Do we want to skip this? Or is this still WIP? I'm guessing this doesn't exercise the codepath right now, is that correct? As you say, we'd need some better testing infrastructure to make it possible?

lunaris (commented):
Thanks for this Eron! Looks great, I guess we just need to figure out how to test it?

tests/examples/java_examples_test.go
EronWright: Am skipping until the kubernetes provider has been updated with a new release of pulumi-java. It is a catch-22 that this PR is needed first.  Of course I would have preferred a self-contained example but I was unable to find a way to exercise the relevant code path without  the use of an MLC.  Maybe the 'echo' provider in the "pulumi" submodule could be used, but I don't have time to figure that out.

PR #1395, opened 2024-07-29
prepare for v0.13.0

EronWright:
# Description

Prepare for v0.13.0 release.

### Improvements

- Codegen: pluginDownloadURL is now supported through the automatically generated build.gradle
- Update pulumi/pulumi to v3.121.0

### Bug Fixes

- Unable to rehydrate a resource that has required inputs

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,5 @@
 ### Improvements
 
-- Codegen: pluginDownloadURL is now supported through the automatically generated build.gradle
 
 ### Bug Fixes
 
-- Unable to rehydrate a resource that has required inputs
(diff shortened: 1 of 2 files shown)


blampe (approved):

CHANGELOG_PENDING.md
```diff
@@ -1,7 +1,5 @@
 ### Improvements
 
-- Codegen: pluginDownloadURL is now supported through the automatically generated build.gradle
```
justinvp, line 3: I think we want to keep the changes in this file, as it'll be used by goreleaser when creating the GH release. But it shouldn't be a problem. We can edit the GH release's description after the fact.

EronWright: Sorry I thought it was a manual step.

PR #1399, opened 2024-08-05
Print to use gen-sdk instead of pulumi-java-gen

Frassle:
We should clean up all our uses of java-gen before merging this, but once we've done that make it clear that this command is deprecated and gen-sdk should be used instead.


lunaris (approved):

Frassle:
We want to close https://github.com/pulumi/pulumi-java/issues/1593 before merging this.

PR #1401, opened 2024-08-06
Updated docs on outputs

pierskarsenbarg:
The previous version was technically correct in that you shouldn't create resources within an apply function but was incorrect that it is not allowed

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
@@ -58,7 +58,7 @@ public interface Output<T> extends Copyable<Output<T>> {
      * The result remains an {@link Output}{@literal <T>} so that dependent resources
      * can be properly tracked.
      * <p>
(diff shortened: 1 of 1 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/core/Output.java
```diff
@@ -58,7 +58,7 @@ public interface Output<T> extends Copyable<Output<T>> {
      * The result remains an {@link Output}{@literal <T>} so that dependent resources
      * can be properly tracked.
      * <p>
-     * {@code func} is not allowed to make resources.
+     * {@code func} hould not be used to create resources unless necessary. Creating resources within 'func' can cause unknown behaviour.
```
tgummerer, line 61:
```suggestion
     * {@code func} should not be used to create resources unless necessary as `func` may not be run during some program executions.
```

Let's make it the same as in https://github.com/pulumi/pulumi/pull/16885.

tgummerer (approved):
Sorry for the delay in reviewing this.

PR #1407, opened 2024-08-08
Prepare for v0.14.0

lunaris:


tgummerer (approved):

CHANGELOG_PENDING.md
```diff
@@ -1,12 +1,4 @@
 ### Improvements
 
 
-### Bug Fixes
```
justinvp, line 4: Just a note (it's been done incorrectly the past couple of releases): I think we want to keep the `CHANGELOG_PENDING.md` as-is for the release, as this is what ends up going in the GitHub release's release notes. Then after the release is created, we should do a follow-up change to clear the pending changelog.

But it's not a big deal. We can edit the GitHub release's release notes to fill in the things from here.

Issue #1422, opened 2024-09-05
Debugger Support: Java


justinvp:
Add debugging support to the Java language host.

PR #1423, opened 2024-09-09
implement debugging support for Java

tgummerer:
For Java to run the program with debugger support, we currently only support maven, so we can use mvnDebug to run it. There might be other ways to start the debugger using
https://github.com/microsoft/java-debug, which could be more generic, but the documentation is somewhat lacking, and this works.  There's always a possibility of swapping this out for java-debug later if we find this method lacking.

Fixes https://github.com/pulumi/pulumi-java/issues/1422

### DIFF
--- a/pkg/internal/executors/executor.go
+++ b/pkg/internal/executors/executor.go
@@ -50,7 +50,8 @@ type javaExecutorFactory interface {
 	NewJavaExecutor(JavaExecutorOptions) (*JavaExecutor, error)
 }
 
-// Create a new java executor. Note that when a debugger is being attached the only executor that is supported is the maven executor.
+// Create a new java executor. Note that when a debugger is being attached the only
+// executor that is supported is the maven executor.
 func NewJavaExecutor(opts JavaExecutorOptions, attachDebugger bool) (*JavaExecutor, error) {
 	if attachDebugger {
 		opts.UseExecutor = "mvnDebug"


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -138,9 +142,9 @@ func newLanguageHost(execOptions executors.JavaExecutorOptions,
 	}
 }
 
-func (host *javaLanguageHost) Executor() (*executors.JavaExecutor, error) {
-	if host.currentExecutor == nil {
-		executor, err := executors.NewJavaExecutor(host.execOptions)
+func (host *javaLanguageHost) Executor(attachDebugger bool) (*executors.JavaExecutor, error) {
+	if host.currentExecutor == nil || attachDebugger {
+		executor, err := executors.NewJavaExecutor(host.execOptions, attachDebugger)
```
lunaris, line 147: Should this be part of `execOptions`, maybe?

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -447,6 +504,16 @@ func (host *javaLanguageHost) RuntimeOptionsPrompts(_ context.Context,
 	return &pulumirpc.RuntimeOptionsResponse{}, nil
 }
 
+func WaitForDebuggerReady(out io.Reader) error {
```
lunaris, line 507: A doc comment on this might be good. Is the string the same regardless of whether we use Maven, etc.?

pkg/internal/executors/executor.go
```diff
@@ -49,7 +50,22 @@ type javaExecutorFactory interface {
 	NewJavaExecutor(JavaExecutorOptions) (*JavaExecutor, error)
 }
 
-func NewJavaExecutor(opts JavaExecutorOptions) (*JavaExecutor, error) {
+func NewJavaExecutor(opts JavaExecutorOptions, attachDebugger bool) (*JavaExecutor, error) {
```
lunaris, line 53: I think if we can move it into `JavaExecutorOptions`, that'd be nice because we could then add a doc comment that explains that if you set this, we are going to pick Maven for you.

pkg/internal/executors/executor.go
```diff
@@ -49,7 +50,22 @@ type javaExecutorFactory interface {
...
+	if attachDebugger {
+		opts.UseExecutor = "mvnDebug"
+		e, err := combineJavaExecutorFactories(
+			&maven{},
+		).NewJavaExecutor(opts)
+		if e == nil {
+			return nil, errors.New(
+				"failed to configure executor.  For debugging only the maven executor is supported, " +
+					"and mvnDebug must be installed")
```
lunaris, line 62: When's the case that `e` is `nil` but `err` is also `nil`?

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -314,9 +338,42 @@ func (host *javaLanguageHost) Run(_ context.Context, req *pulumirpc.RunRequest)
 	var stdoutBuf bytes.Buffer
 	var stderrBuf bytes.Buffer
 
-	cmd.Stdout = &stdoutBuf
+	pr, pw := io.Pipe()
+
+	cmd.Stdout = pw
```
lunaris, line 343: A comment might be helpful here as it took me a minute to work out what's going on. As I understand it, we want to keep sending the command's output to stdout, but we also need to process it to see when the debugger is ready -- is that right?

pkg/cmd/pulumi-language-java/main.go
```diff
@@ -314,9 +338,42 @@ func (host *javaLanguageHost) Run(_ context.Context, req *pulumirpc.RunRequest)
...
+			contract.IgnoreError(cmd.Process.Kill())
+		}
+		_, err = engineClient.StartDebugging(ctx, &pulumirpc.StartDebuggingRequest{
+			Config:  debugConfig,
+			Message: "on port 8000",
+		})
+		if err != nil {
+			logging.Errorf("unable to start debugging: %v", err)
+			contract.IgnoreError(cmd.Process.Kill())
+		}
```
lunaris, line 375: Almost makes you want to use a `goto` :trollface:

lunaris (approved):
I think this is in good shape. A few cleanups but otherwise great stuff!

pkg/cmd/pulumi-language-java/main.go
tgummerer: I had it there initially, but it felt better to do it this way.  `execOptions` is a global thing, while we only want to attach the debugger when running the program (and only have access to the information whether a debugger is going to be attached when running the program).  So adding it to `execOptions` gets a bit messy.

pkg/internal/executors/executor.go
tgummerer: If the project is not a maven project, `e` is nil, but and `err` also, since it's not really an error from the point of view of the maven executor (it's just not that kind of project, there was no failure to set it up).

pkg/cmd/pulumi-language-java/main.go
tgummerer: Tempting ;)

pkg/internal/executors/executor.go
tgummerer: I added a comment to this function to specify that.

PR #1424, opened 2024-09-10
Prepare for v0.16.0 release

justinvp:


CHANGELOG_PENDING.md
```diff
@@ -1,8 +1,6 @@
 ### Improvements
 
-- Classloader: now supports dynamic exclusions of packages from system property.
+- Add support for attaching debuggers
 
 ### Bug Fixes
 
-- Fix HTML escaping after `{@literal ...}` blocks in Javadocs
-- Fix issue with direct dependencies causing IndexOutOfBoundsException
```
Frassle, line 8: Doesn't look like this line is in CHANGELOG.md

justinvp: Good catch. Fixed up CHANGELOG_PENDING.md and CHANGELOG.md.

Frassle (approved):

CHANGELOG_PENDING.md
Frassle: We should probably use changie to stop cases like this....

justinvp: > We should probably use changie to stop cases like this....

Yes

PR #1465, opened 2024-11-18
Don't cache Java executors

lunaris:
Presently, the Java language host caches a single executor, which abstracts e.g. the command-line executions needed to run Java code. This is problematic should a single language runtime ever be fielded with concurrent requests for program execution. This commit removes this caching for now to fix this.


tgummerer (approved):
Makes sense.  Creating these executors doesn't seem super expensive (We're looking for the paths for each of them on the filesystem in the worst case), so I don't think this is going to be a big detriment to performance.

Frassle (approved):

PR #1473, opened 2024-11-18
Implement `Pack`

lunaris:
The `LanguageRuntime.Pack` method allows a caller to package code (typically a generated SDK) into a language-specific artifact. Its primary use at present is in conformance testing, where we generate SDKs for various test cases and package them up for use in test programs to assert properties about a given language runtime and its code generation facilities. This commit implements `Pack` for Java, producing *Java archives* (`.jar`s) from a given source location using `gradle` (which is assumed to be the packaging system of choice since we presently use this for SDK generation).

### DIFF
--- a/pkg/cmd/pulumi-language-java/main.go
+++ b/pkg/cmd/pulumi-language-java/main.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"encoding/xml"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi-language-java/main.go
```diff
@@ -767,3 +769,134 @@ func (host *javaLanguageHost) GeneratePackage(
...
 }
+
+// Implements the `LanguageRuntime.Pack` RPC method, which packs a package (SDK) into a language-specific artifact. In
+// the case of Java, the artifacts in question are *Java Archives* (or JARs/`.jar` files).
+//
+// The incoming `PackRequest` specifies the directory to be packaged and where the packed artifact should be produced.
+// Since we expect that our packed artifacts will be consumed by other Java libraries and programs, this implementation
+// creates a Maven artifact repository at the specified location and publishes a `.jar` there using the
+// `publishToMavenLocal` Gradle target (generated SDKs, which is what we expect to be packing, currently use Gradle as
+// their build system). We then return in the `PackResponse` to the path to the `.jar` we published.
```
Frassle, line 780:
```suggestion
// their build system). We then return in the `PackResponse` the path to the `.jar` we published.
```

lunaris: Ach good catch -- fixed

Frassle (approved):

PR #1479, opened 2024-11-20
Bump default `gradle-nexus.publish-plugin` version

blampe:
[Release notes.](https://github.com/gradle-nexus/publish-plugin/releases/tag/v2.0.0)

> Backward incompatible changes
>
> closeAndReleaseStagingRepository has been renamed to closeAndReleaseStagingRepositories for consistency

We [don't use](https://github.com/search?q=org%3Apulumi%20closeAndReleaseStagingRepository&type=code) this option, so this shouldn't impact us.

Refs https://github.com/pulumi/home/issues/3762.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,6 @@
 CHANGELOG
 =========
 
-## Unreleased
-
-### Improvements
-
-- Update `io.github.gradle-nexus.publish-plugin` to 2.0.0 from 1.1.0.
-
 ## 0.17.0 (2024-11-05)
 
 ### Improvements
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,4 +10,6 @@
 
 - Enable language conformance tests for Java
 
+- Update `io.github.gradle-nexus.publish-plugin` to 2.0.0 from 1.1.0.
+
 ### Bug Fixes


CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+## Unreleased
+
+### Improvements
+
+- Update `io.github.gradle-nexus.publish-plugin` to 2.0.0 from 1.1.0.
+
```
lunaris, line 9: If you add this to `CHANGELOG_PENDING` it gets taken care of for you as part of the release process.

lunaris (approved):
Minor comment on the changelog stuff but LGTM otherwise

t0yv0 (approved):
I have some hesitation with respect to this dependency so I hope we tried running it at least once to see it produces the releases and does not require changes from our SDK gen setup (2.0 might be breaking after all) before rolling out here. But otherwise LGTM :)

blampe:
@t0yv0 of course -- https://github.com/pulumi/pulumi-command/actions/runs/11942352524/job/33292492882 seems there was a flake but it eventually succeeded.

CHANGELOG.md
blampe: My kingdom for a consistent changelog process :)

PR #1482, opened 2024-11-22
Allow specifying dependencies for output invokes

julienp:
Provider functions that take inputs as arguments, and return an output (aka output form invokes), now allow specifying a depends_on option. This allows programs to ensure that invokes are executed after things they depend on, similar to the depdends_on resource option.

Fixes https://github.com/pulumi/pulumi/issues/17754

### DIFF
--- a/pkg/cmd/pulumi-language-java/testdata/sdks/alpha-3.0.0-alpha.1.internal/src/main/java/com/pulumi/alpha/Utilities.java
+++ b/pkg/cmd/pulumi-language-java/testdata/sdks/alpha-3.0.0-alpha.1.internal/src/main/java/com/pulumi/alpha/Utilities.java
@@ -69,7 +69,7 @@ public class Utilities {
         );
     }
 
-    public static InvokeOutputOptions invokeOutputOptionsWithVersion(@Nullable InvokeOutputOptions options) {
+    public static InvokeOutputOptions withVersion(@Nullable InvokeOutputOptions options) {
         if (options != null && options.getVersion().isPresent()) {
             return options;
         }
--- a/pkg/cmd/pulumi-language-java/testdata/sdks/simple-2.0.0/src/main/java/com/pulumi/simple/Utilities.java
+++ b/pkg/cmd/pulumi-language-java/testdata/sdks/simple-2.0.0/src/main/java/com/pulumi/simple/Utilities.java
@@ -69,7 +69,7 @@ public class Utilities {
         );
     }
 
-    public static InvokeOutputOptions invokeOutputOptionsWithVersion(@Nullable InvokeOutputOptions options) {
+    public static InvokeOutputOptions withVersion(@Nullable InvokeOutputOptions options) {
         if (options != null && options.getVersion().isPresent()) {
             return options;
         }


sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOptions.java
```diff
@@ -17,7 +17,7 @@
  * Options to help control the behavior of @see {@link Deployment#invokeAsync(String, TypeShape, InvokeArgs, InvokeOptions)}.
  */
 @ParametersAreNonnullByDefault
-public final class InvokeOptions {
+public class InvokeOptions {
```
julienp, line 20: I want `InvokeOutputOptions` to inherit from this, do we care about this being `final`?

sdk/java/pulumi/src/test/java/com/pulumi/deployment/DeploymentInvokeDependsOnTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+public class DeploymentInvokeDependsOnTest {
+
+    @AfterEach
+    void cleanup() {
+        PulumiTestInternal.cleanup();
+    }
+
+    @Test
+    void testInvokesDependsOn() {
```
julienp, line 42: We have a unit test here, but the language conformance test `l2-invoke-options-depends-on` is not enabled yet. I spent some time trying to fix program gen to generate something usable, but haven't completed that work yet. The way we generate lambdas is broken, and requires work.

pkg/codegen/java/templates.go
```diff
@@ -77,16 +78,28 @@ public class {{ .ClassName }} {
...
+            return options;
+        }
+        return new InvokeOptions(
+            options == null ? null : options.getParent().orElse(null),
+            options == null ? null : options.getProvider().orElse(null),
+            getVersion()
+        );
+    }
+
+    public static InvokeOutputOptions invokeOutputOptionsWithVersion(@Nullable InvokeOutputOptions options) {
```
lunaris, line 92: Not sure if it's a good idea, but you can overload static methods, so we could call this `withVersion` too -- terrible/won't work?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOutputOptions.java
```diff
@@ -0,0 +1,67 @@
...
+
+import javax.annotation.Nullable;
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.Optional;
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Options to help control the behavior of @see
```
lunaris, line 18: I don't think you need/should have `@see` here -- `{@link ...}` should be sufficient

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOutputOptions.java
```diff
@@ -0,0 +1,67 @@
...
+    public InvokeOutputOptions(@Nullable Resource parent, @Nullable ProviderResource provider, @Nullable String version,
+            @Nullable List<Resource> dependsOn) {
+        super(parent, provider, version);
+        this.dependsOn = dependsOn;
+    }
+
+    /**
+     * Optional resources that this invoke depends on. The invoke will wait for
+     * these
+     * resources to be resolved before executing.
```
lunaris, line 41: Odd wrapping?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOutputOptions.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    public List<Resource> getDependsOn() {
+        return this.dependsOn == null ? List.of() : this.dependsOn;
```
lunaris, line 44: Wondering if we should downgrade this to `Iterable<Resource>`, since as it is you could `getDependsOn().add(...)` and mutate the `dependsOn`. Alternatively do `... ? List.of() : new ArrayList<>(this.dependsOn)`, for instance, so that callers get a copy that doesn't affect the underlying object. Maybe this is overkill.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -505,6 +523,37 @@ public <T> Output<T> invoke(String token, TypeShape<T> targetType, InvokeArgs ar
...
+            Objects.requireNonNull(args);
+            Objects.requireNonNull(options);
+
+            log.debug(String.format("Invoking function: token='%s' asynchronously", token));
+
+            final CompletableFuture<String> packageRefFuture = packageRef == null
+                ? CompletableFuture.completedFuture(null)
+                : packageRef;
+
+            // Find all the resource dependencies from dependsOn, we need to wait for this task to complete
```
lunaris, line 538:
```suggestion
            // Find all the resource dependencies from dependsOn. We need to wait for these futures to complete
```

lunaris (approved):
Super cool!

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOutputOptions.java
julienp: `ew ArrayList<>(this.dependsOn)` sounds good!

Issue #1500, opened 2024-12-11
Default dependencies not added when caled from `pulumi package gen-sdk`


danielrbradley:
When invoking the pulumi-java-gen program directly, it computes the build options as a combination of the schema language options and the command line flags in command.go:

https://github.com/pulumi/pulumi-java/blob/d629808b011e82ffb30b362e4e65c10d0f9a7af2/pkg/cmd/pulumi-java-gen/command.go#L120-L150

command.go then calls [`generateJava`](https://github.com/pulumi/pulumi-java/blob/d629808b011e82ffb30b362e4e65c10d0f9a7af2/pkg/cmd/pulumi-java-gen/generate.go#L47) which calls javagen.GeneratePackage .. which what is called from `pulumi package gen-sdk`. 

When invoking the pulumi-java-gen you therefore get `WithDefaultDependencies` [applied consistently](https://github.com/pulumi/pulumi-java/blob/d629808b011e82ffb30b362e4e65c10d0f9a7af2/pkg/cmd/pulumi-java-gen/command.go#L140-L141) but no default dependencies when running from `pulumi package gen-sdk`.

PR #1501, opened 2024-12-11
Apply default dependencies when called from Pulumi CLI

danielrbradley:
The `pulumi package gen-sdk` command calls `GeneratePackage` directly, so we need to apply the defaults at that layer rather than in the pulumi-java-gen specific code.

Fixes https://github.com/pulumi/pulumi-java/issues/1500

### DIFF
--- a/pkg/codegen/java/gen.go
+++ b/pkg/codegen/java/gen.go
@@ -22,7 +22,7 @@ import (
 )
 
 // This should be bumped as required at the point of release.
-var DEFAULT_SDK_VERSION = semver.Version{Major: 0, Minor: 18, Patch: 0}
+var DefaultSdkVersion = semver.Version{Major: 0, Minor: 18, Patch: 0}
 
 func packageName(packages map[string]string, name string) string {
 	if pkg, ok := packages[name]; ok {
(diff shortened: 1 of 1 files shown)


lunaris (approved):
Looks good -- thanks for raising!

Frassle (approved):

danielrbradley:
This fixes two of the three missing dependencies in https://github.com/pulumi/pulumi-azure-native/pull/3776

However, the third missing dependency is actually the Pulumi Java SDK itself. This is added when calling the pulumi-java-gen binary directly because the build process bakes in the version number. However, AFAIK there's no way for a Go package to fetch its own version automatically.

Question for @lunaris and @Frassle: do we just hard code the self-reference and manually bump it as required or should we do something clever like letting the pulumi/cmd set the verison?

Frassle:
> do we just hard code the self-reference and manually bump it as required or should we do something clever like letting the pulumi/cmd set the verison?

We add it as a linker flag when building. https://github.com/pulumi/pulumi-java/blob/main/.goreleaser.yml#L22

danielrbradley:
> We add it as a linker flag when building. https://github.com/pulumi/pulumi-java/blob/main/.goreleaser.yml#L22

That's fine when building the binary, but for `pulumi package gen-sdk` it's running this code as a go module reference, so doesn't get the version injected because there's no "build time".

blampe:
Also unblocks https://github.com/pulumi/pulumi-kubernetes/pull/3151

Frassle:
> That's fine when building the binary, but for pulumi package gen-sdk it's running this code as a go module reference, so doesn't get the version injected because there's no "build time". Unless I'm missing something about how we release the go package?

Oh right yeh. We plan on moving this code to run over grpc so it will actually be in the build pulumi-language-java binary. Probably easier to just wait for that change. I imagine we're close to switching over, I think the java binary does now support these rpc calls.

danielrbradley:
Ok, will stick with the hard coded default for now but can override that when we do the grpc plugin change.

PR #1509, opened 2024-12-12
Add release instructions to CONTRIBUTING.md

julienp:
I always have to think way too hard about how we do this, and look at old PRs and shell history. Let’s document this!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -169,5 +169,5 @@ To release a new version of `pulumi-java`, follow these steps:
   ./scripts/release.sh v0.20.0
   ```
 
-3. Create a new pull request that emptries the pending changelog in
-  [CHANGELOG_PENDING.md](./CHANGELOG_PENDING.md).
+3. Create a new pull request that empties the pending changelog in
+  [CHANGELOG_PENDING.md](./CHANGELOG_PENDING.md) and merge it.


CONTRIBUTING.md
```diff
@@ -149,3 +149,15 @@ If you want to talk with other folks in the Pulumi community
 (including members of the Pulumi team) come hang out in the
 `#contribute` channel on the
 [Pulumi Community Slack](https://slack.pulumi.com/).
+
+## Release Process
+
+To release a new version of `pulumi-java`, create a new pull request that
+updates [CHANGELOG.md](./CHANGELOG.md) and [CHANGLOG_PENDING.md](./CHANGLOG_PENDING.md).
```
lunaris, line 156: I think we should be more explicit here since there is a multi-step process:

1. Raise a PR that copies pending to changelog. Leave pending as is -- if you empty it at this point, I think there is some GH process (e.g. generated release notes) that fails -- Justin has details IIRC.
2. Merge that PR
3. _Make sure your local `main` is up-to-date and that you do not just push your local copy of that PR_ (I have done this several times :see_no_evil: )
4. When you are sure that `main` == `origin/main`, run the release script
5. Raise another PR that empties out pending changelog when the release has completed

lunaris (approved):
Y E S   :boom:

CONTRIBUTING.md
julienp: Aha, it does break the release notes here https://github.com/pulumi/pulumi-java/releases :)

julienp: (I manually updated them just now)

PR #1539, opened 2025-01-02
Implement `getOrganizationName`

### LINKED ISSUE #877: Add organizationName to Context
## Hello!

## Issue details

See https://github.com/pulumi/pulumi/issues/2800 for context.

We should add `organizationName` to `Context`.

justinvp:
This change implements `getOrganizationName` on `Deployment` and `Context`, as well as providing a way to specify it via `TestOptions`.

Note: I would like to have enabled `l1-builtin-info`, but that's still blocked on the `ctx.export` issue (#1540).

Fixes #877

---

Reference:
- https://github.com/pulumi/pulumi/pull/10564
- https://github.com/pulumi/pulumi/pull/10820

### DIFF
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -664,8 +664,8 @@ func (g *generator) collectFunctionCallImports(functionCall *model.FunctionCallE
 				imports = append(imports, collectObjectImports(argumentsExpr, argumentExprType)...)
 			}
 		}
-	case "stack", "projectName":
-		// stack() and projectName() functions are pulumi built-ins
+	case "stack", "project", "organization":
+		// stack(), project(), and organization() functions are pulumi built-ins
 		// they require the Deployment class
 		imports = append(imports, "com.pulumi.deployment.Deployment")
 	}
--- a/sdk/java/pulumi/src/main/java/com/pulumi/Context.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/Context.java
@@ -22,6 +22,13 @@ public interface Context extends OutputContext, ConfigContext {
     String stackName();
 
     /**
+     * @return the name of the current organization
+     */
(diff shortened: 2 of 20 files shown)


pkg/codegen/java/gen_program.go
```diff
@@ -654,8 +654,8 @@ func (g *generator) collectFunctionCallImports(functionCall *model.FunctionCallE
 				imports = append(imports, collectObjectImports(argumentsExpr, argumentExprType)...)
 			}
 		}
-	case "stack", "projectName":
-		// stack() and projectName() functions are pulumi built-ins
+	case "stack", "project", "organization":
```
justinvp, line 657: Using `projectName` was a bug, so I've gone ahead and fixed it here. The PCL name is `project`. I confirmed with the `pulumi-builtins.pp` programgen test.

sdk/java/pulumi/src/main/java/com/pulumi/Context.java
```diff
@@ -21,6 +21,13 @@ public interface Context extends OutputContext, ConfigContext {
      */
     String stackName();
 
+    /**
+     * @return the name of the current organization
+     */
+    default String organizationName() {
```
justinvp, line 27: I decided to provide a default implementation out of an abundance of caution (here and on `Deployment`). I could see an argument for not doing it, though, if we don't care about any external implementers of the interfaces. Curious if anyone feels strongly that we shouldn't provide a default implementation.

lunaris (approved):

sdk/java/pulumi/src/main/java/com/pulumi/Context.java
lunaris: I think it's OK for now. If we decide to remove it (e.g. ahead of 1.0), it'd be no more breaking than not adding it in the first place would have been?

PR #1547, opened 2025-01-03
Revert "build(deps): bump grpcVersion from 1.57.2 to 1.69.0 in /sdk/java"

tgummerer:
Reverts pulumi/pulumi-java#1528

This upgrade makes CI fail.  This dependabot update was previously reverted in https://github.com/pulumi/pulumi-java/pull/1523.

Fixes https://github.com/pulumi/pulumi-java/issues/1545


lunaris:
I cheekily already reverted this as part of another commit it made fail -- my bad, sorry. It seems that the `examples` tests, which detect this, didn't re-run as part of the Dependabot rebase :thinking: Or I screwed up somehow (likely).

tgummerer:
Ah I see, it's in https://github.com/pulumi/pulumi-java/commit/465ddfffb984ef1952e64eabc66f692859bd2e6c.

Gonna close this out.  

Awkwardly these tests don't run in dependabot PRs it seems because those don't have access to the credentials I assume (https://github.com/pulumi/pulumi-java/blob/main/.github/workflows/pr.yml#L192).  Not sure there's a good way to fix that with GitHub actions either :(

PR #1570, opened 2025-01-06
Link issues for skipped conformance tests

lunaris:
Currently, we skip a number of conformance tests which Java does not currently pass. This commit links the skipped tests to relevant GitHub issues so that we can track progress when it comes to fixing them.


justinvp (approved):

julienp (approved):
❤️

PR #1573, opened 2025-01-06
[program-gen] Fix emitted import statements in generated programs such that imports with the same symbol are fully qualified

### LINKED ISSUE #1558: [programgen] Duplicate identifiers aren't fully qualified
The `l2-failed-create-continue-on-error` conformance test must currently be skipped because it emits a program like:

```java
package generated_program;

import com.pulumi.Context;
import com.pulumi.Pulumi;
import com.pulumi.fail_on_create.Resource;
import com.pulumi.fail_on_create.ResourceArgs;
import com.pulumi.simple.Resource;
import com.pulumi.simple.ResourceArgs;
import com.pulumi.resources.CustomResourceOptions;
import java.util.List;
import java.util.ArrayList;
import java.util.Map;
import java.io.File;
import java.nio.file.Files;
import java.nio.file.Paths;

public class App {
        Pulumi.run(App::stack);
    }

        var failing = new Resource("failing", ResourceArgs.builder()
            .value(false)
            .build());

        var dependent = new Resource("dependent", ResourceArgs.builder()
            .value(true)
            .build(), CustomResourceOptions.builder()
                .dependsOn(failing)
                .build());

        var dependent_on_output = new Resource("dependent_on_output", ResourceArgs.builder()
            .value(failing.value())
            .build());

        var independent = new Resource("independent", ResourceArgs.builder()
            .value(true)
            .build());

        var double_dependency = new Resource("double_dependency", ResourceArgs.builder()
            .value(true)
            .build(), CustomResourceOptions.builder()
                .dependsOn(                
                    independent,
                    dependent_on_output)
                .build());

    }
}
```

Note these imports:

```java
import com.pulumi.fail_on_create.Resource;
import com.pulumi.fail_on_create.ResourceArgs;
import com.pulumi.simple.Resource;
import com.pulumi.simple.ResourceArgs;
```

Uses of `new Resource("...", ResoureArgs.builder().build())` doesn't work because it's ambiguous which `Resource` and `ResourceArgs` is intended:

```
reference to Resource is ambiguous
          both class com.pulumi.fail_on_create.Resource in com.pulumi.fail_on_create and class com.pulumi.simple.Resource in com.pulumi.simple match
reference to ResourceArgs is ambiguous
          both class com.pulumi.fail_on_create.ResourceArgs in com.pulumi.fail_on_create and class com.pulumi.simple.ResourceArgs in com.pulumi.simple match
```

Those imports must be elided and the uses must be fully qualified when there are duplicates like this, e.g.:
- `new com.pulumi.fail_on_create.Resource("...", com.pulumi.fail_on_create.ResourceArgs.builder().build())`
- `new com.pulumi.simple.Resource("...", com.pulumi.simple.ResourceArgs.builder().build())`

Zaid-Ajaj:
### Description

Fix emitted import statements in generated programs such that imports with the same symbol are fully qualified

if we have two imports such as the following:
- `com.pulumi.my_package.ResourceType`
- `com.pulumi.your_package.ResourceType`

then we shouldn't generate import statements for them instead, we use the fully qualified name at the usage site. 

Unskips `l2-failed-create-continue-on-error`

Fixes #1558

### DIFF
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -593,8 +593,13 @@ func (g *generator) collectResourceImports(resource *pcl.Resource) []string {
 	if len(resource.Inputs) > 0 || hasCustomResourceOptions(resource) {
 		// import args type name
 		argsTypeName := g.resourceArgsTypeName(resource)
-		resourceArgsImport := pulumiImport(pkg, module, argsTypeName)
-		imports = append(imports, resourceArgsImport)
+		alreadyFullyQualified := strings.Contains(argsTypeName, ".")
+		if !alreadyFullyQualified {
+			resourceArgsImport := pulumiImport(pkg, module, argsTypeName)
+			imports = append(imports, resourceArgsImport)
+		} else {
+			imports = append(imports, argsTypeName)
+		}
 		resourceProperties := typedResourceProperties(resource)
 		for _, inputProperty := range resource.Inputs {
 			inputType := resourceProperties[inputProperty.Name]


pkg/codegen/java/gen_program.go
```diff
@@ -745,17 +746,45 @@ func (g *generator) genPreamble(w io.Writer, nodes []pcl.Node) {
...
+	// instead, we use the fully qualified name at the usage site
+	importsWithDuplicateName := codegen.NewStringSet()
+	for _, importDefA := range importStatements {
+		for _, importDefB := range importStatements {
+			if importMember(importDefA) == importMember(importDefB) && importDefA != importDefB {
+				importsWithDuplicateName.Add(importDefA)
+				importsWithDuplicateName.Add(importDefB)
+			}
+		}
+	}
```
brandonpollack23, line 769: I'm kind of not a fan of this n^2 approach when we can use a set of importMembers.

what if it is a super large program with many files each with tons of imports? It's premature optimization but seems reasonable to turn into a set check

pkg/codegen/java/gen_program.go
```diff
@@ -745,17 +746,45 @@ func (g *generator) genPreamble(w io.Writer, nodes []pcl.Node) {
...
 
-		g.genImport(w, importDef)
+		if !importsWithDuplicateName.Has(importDef) {
+			// do not generate import statements for symbols with duplicate members
+			// because those will require full qualification at the usage site
+			g.genImport(w, importDef)
+		}
 	}
 
+	g.importStatements = importStatements
```
brandonpollack23, line 786: this makes me think the name is misleading.  They are not import statements they are dependencies because genImport is only called for non conflicting symbols

pkg/codegen/java/gen_program.go
```diff
@@ -817,20 +846,28 @@ func (g *generator) genPostamble(w io.Writer, _ []pcl.Node) {
...
 		member = "Provider"
 	}
 
+	for _, importStatement := range g.importStatements {
+		// if there is an import statement that ends with the same member name
+		// return the fully qualified resource name
+		if strings.HasSuffix(importStatement, "."+member) && pulumiImport(pkg, module, member) != importStatement {
+			return pulumiImport(pkg, module, member)
+		}
+	}
```
brandonpollack23, line 863: instead of saving all import statements wouldn't it also be better to just save a set of all types that need to be fully qualified and just check this isn't in that set?

brandonpollack23 (changes requested):

pkg/codegen/java/gen_program.go
Zaid-Ajaj: Renamed to `emittedImportTypeSymbols` ✅

pkg/codegen/java/gen_program.go
Zaid-Ajaj: Done ✅

pkg/codegen/java/gen_program.go
Zaid-Ajaj: I've optimized the lookup ✅ can you have another look?

brandonpollack23 (approved):

PR #1580, opened 2025-01-08
[program-gen] Fix generation of `double` literals in generated programs

### LINKED ISSUE #1565: [programgen] Better generation of numeric literals
When we generate numeric literals, we'd ideally have context of the Java type we are generating a literal for. In PCL, all literals are "numbers", which could be floating point or integers. When we spit these out into Java, we have some code that tries to omit decimal points etc. depending on the value, but sometimes this isn't enough. In the `l2-resource-primitive` conformance test, for instance, we need to produce a varargs list of `double`s, but we produce integer literals. Java can't convert "twice" -- the integer literals to doubles, then the vararg `...double` to `double[]`, so we get a compile error. "Just" switching all literals to have a double-indicating `d` suffix, breaks things that need integers, so we probably need to rework programgen to carry more information here.

`l2-resource-primitive` generated program (`numberArray` is the problematic part):

```java
package generated_program;

import com.pulumi.Context;
import com.pulumi.Pulumi;
import com.pulumi.primitive.Resource;
import com.pulumi.primitive.ResourceArgs;
import java.util.List;
import java.util.ArrayList;
import java.util.Map;
import java.io.File;
import java.nio.file.Files;
import java.nio.file.Paths;

public class App {
        Pulumi.run(App::stack);
    }

        var res = new Resource("res", ResourceArgs.builder()
            .boolean_(true)
            .float_(3.14)
            .integer(42)
            .string("hello")
            .numberArray(            
                -1,
                0,
                1)
            .booleanMap(Map.ofEntries(
                Map.entry("t", true),
                Map.entry("f", false)
            ))
            .build());

    }
}
```

Compilation error:

```
[ERROR] COMPILATION ERROR : 
[ERROR] /run/user/1000/TestLanguage1675730795/002/projects/l2-resource-primitives/src/main/java/generated_program/App.java:[26,13] method numberArray in class com.pulumi.primitive.ResourceArgs.Builder cannot be applied to given types;
  required: java.lang.Double[]
  found: int,int,int
  reason: varargs mismatch; int cannot be converted to java.lang.Double
```

Zaid-Ajaj:
### Description

This PR implements an initial pass for handling `__convert` calls in program-gen starting with `double` literals. Later on we will extend this to handle enums but we don't have a conformance test for these yet.

The return type of `__convert` calls is annotated with the target schema type so we detect when are going to emit an expression of type `model.NumberType` (in PCL space) in a location that expects `schema.NumberType`. When that pattern is detected we make sure to emit double literals that end with `.0` when the value of the number is exact. 

- Fixes #1565 
- Unskips `l2-resource-primitive` now that the test is passing
- ~~Removes `l2-map-keys` as there is no such test~~

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,5 +2,6 @@
 
  - Fix emitted functions for `assetArchive` and `remoteAsset` in generated programs
  - Fix generation of `double` literals in generated programs
+ - Avoid calling invokes with dependencies on unknown resources
 
 ### Bug Fixes
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
@@ -91,6 +91,7 @@ import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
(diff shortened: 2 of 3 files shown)


Zaid-Ajaj:
It looks like the program generation tests (not conformance tests) are failing due to missing type information of object expressions. I'll have a look

lunaris (approved):

Issue #1566, opened 2025-01-06
Fix `l2-resource-config` / plugin download URL code generation


lunaris:
The `l2-resource-config` conformance test currently fails due to some values not lining up. It looks like we are not respecting plugin download URLs (at least -- there may be other things we are missing).

```
=== NAME  TestLanguage/l2-resource-config
    language_test.go:146: l2_resource_config.go:56: 
                Error Trace:    /home/will/git/pulumi/pulumi-java-main/pulumi/cmd/pulumi-test-language/tests/l2_resource_config.go:56
                                                        /home/will/git/pulumi/pulumi-java-main/pulumi/cmd/pulumi-test-language/interface.go:1098
                                                        /home/will/git/pulumi/pulumi-java-main/pulumi/cmd/pulumi-test-language/tests/types.go:182
                                                        /nix/store/chzgk756zb2cqlzbjr86m0lfxi63cdfy-go-1.22.7/share/go/src/runtime/asm_amd64.s:1695
                Error:          Not equal: 
                                expected: resource.PropertyMap{"__internal":resource.PropertyValue{V:resource.PropertyMap{"pluginDownloadURL":resource.PropertyValue{V:"http://example.com"}}}, "name":resource.PropertyValue{V:"my config"}, "pluginDownloadURL":resource.PropertyValue{V:"not the same as the pulumi resource option"}, "version":resource.PropertyValue{V:"9.0.0"}}
                                actual  : resource.PropertyMap{"__internal":resource.PropertyValue{V:resource.PropertyMap{}}, "name":resource.PropertyValue{V:"my config"}, "pluginDownloadURL":resource.PropertyValue{V:"not the same as the pulumi resource option"}, "version":resource.PropertyValue{V:"9.0.0"}}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -2,6 +2,3 @@
                                  (resource.PropertyKey) (len=10) "__internal": (resource.PropertyValue) {
                                -  V: (resource.PropertyMap) (len=1) {
                                -   (resource.PropertyKey) (len=17) "pluginDownloadURL": (resource.PropertyValue) {
                                -    V: (string) (len=18) "http://example.com"
                                -   }
                                +  V: (resource.PropertyMap) {
                                   }
        
    language_test.go:148: stdout: Created stack 'test'
        
         +  pulumi:pulumi:Stack l2-resource-config-test creating (0s) 
         +  pulumi:providers:config prov creating (0s) 
[...]

PR #1585, opened 2025-01-10
[sdk] Implement InvokeOptionsBuilder and InvokeOutputOptionsBuilder

Zaid-Ajaj:
### Description

The PR implements the following builder classes
 - `InvokeOptionsBuilder` for `InvokeOptions`
 - `InvokeOutputOptionsBuilder` for `InvokeOutputOptions`

It adds `pluginDownloadURL` for both of them.

This is a required in order to invoke options for program-gen for #1566

### API

Example usage for `InvokeOptionsBuilder`
```java
InvokeOptions.builder() // returns InvokeOptionsBuilder
  .parent(...)
  .provider(...)
  .pluginDownloadURL(...)
  .version(...)
  .build() // returns InvokeOptions
```

Example usage for `InvokeOutputOptionsBuilder`

```java
(new InvokeOutputOptionsBuilder())
  .parent(...)
  .provider(...)
  .pluginDownloadURL(...)
  .version(...)
  .dependsOn(...) 
  .build() // returns InvokeOutputOptions
```

The reason `InvokeOutputOptionsBuilder` cannot be initialized via `InvokeOutputOptions.builder()` is because the `builder()` method would conflict with the `InvokeOptions.builder()` method. 

For the time being, we are omitting `.builder()` until we no longer have `InvokeOutputOptions` _extending_ `InvokeOptions.`


sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOptionsBuilder.java
```diff
@@ -0,0 +1,54 @@
+package com.pulumi.deployment;
+
+import com.pulumi.resources.ProviderResource;
+import com.pulumi.resources.Resource;
+import javax.annotation.Nullable;
+
+public final class InvokeOptionsBuilder {
```
lunaris, line 7: I think it's convention (at least in our SDK) to have the builder be a `static` inner class of the type it's constructing, named `Builder`. I reckon we should follow that here so we can be consistent?

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOptionsBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+     * An optional URL, corresponding to the url from which the provider plugin that should be
+     * used when operating on this invoke is downloaded from. This URL overrides the download URL
+     * inferred from the current package and should rarely be used.
+     */
+    public InvokeOptionsBuilder pluginDownloadURL(String pluginDownloadURL) {
+        this.pluginDownloadURL = pluginDownloadURL;
+        return this;
+    }
+
+
```
lunaris, line 50: Extraneous blank lines

sdk/java/pulumi/src/main/java/com/pulumi/deployment/InvokeOutputOptions.java
```diff
@@ -26,12 +26,25 @@ public class InvokeOutputOptions extends InvokeOptions {
...
+        super(parent, provider, version, pluginDownloadURL);
+        this.dependsOn = dependsOn;
+    }
+
+    public InvokeOutputOptions(
+        @Nullable Resource parent, 
+        @Nullable ProviderResource provider, 
+        @Nullable String version,
+        @Nullable List<Resource> dependsOn) {
+        super(parent, provider, version, null);
```
lunaris, line 47: As above, it'd be good if we could reduce one of these constructor overloads to a call to the other.

lunaris (approved):

PR #1586, opened 2025-01-10
[program-gen] Emit invoke options and invoke output options in generated programs

Zaid-Ajaj:
Built on top of https://github.com/pulumi/pulumi-java/pull/1585 in order to generate `InvokeOptions` for invoke calls. 

- Fixes #1566
- Unskips `l2-invoke-options`
- Implements optimized import handling when encountering function calls
- Implements lambda parameter shadowing when generating `apply(...)` 
    - `data.applyValue(data -> data.result())` becomes `data.applyValue(_data -> _data.result())`
- Fixes `apply(...)` calls for invokes when used in output variables

This PR prepares program-gen and sdk to handle `l2-invoke-options-depends-on` which currently generate correct programs but fails the actual conformance test where the invoke has no dependencies cc @julienp

PR #1591, opened 2025-01-14
[sdk/program-gen] Include explicit dependencies of invokes in their resulting output

Zaid-Ajaj:
Built on top of #1586

Fixes the SDK in invokes where explicit dependencies of output invokes should be included in the resulting Output instance

Unskips `l2-invoke-options-depends-on` conformance test that asserts the behaviour

### DIFF
--- a/pkg/cmd/pulumi-language-java/language_test.go
+++ b/pkg/cmd/pulumi-language-java/language_test.go
@@ -161,6 +161,7 @@ func TestLanguage(t *testing.T) {
 // skip them with an informative message until they are fixed.
 var expectedFailures = map[string]string{
 	"l1-output-string":                      "#1562 Large string literals are not generated correctly",
+	"l2-invoke-options-depends-on":          "#1563 Invoke argument and result handling",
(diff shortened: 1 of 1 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -568,7 +568,7 @@ public <T> Output<T> invoke(String token, TypeShape<T> targetType, InvokeArgs ar
                             } else {
                                 return packageRefFuture
                                     .thenCompose(packageRefString -> this.invokeRawAsync(token, serializedArgs, options, packageRefString))
-                                    .thenApply(result -> parseInvokeResponse(token, targetType, result));
```
Zaid-Ajaj, line 571: The actual fix as suggested by @julienp 🚀

julienp (approved):
Amazing work on the whole set of PRs to get program test for invokes working!

lunaris (approved):

PR #1594, opened 2025-01-15
Bump core SDK versions in generated code

### LINKED ISSUE #1589: Audit version bounds we generate in programgen/SDKgen
Now that we have bumped to 1.0.0, we should check and possibly change/fix the version bounds we generate for Maven POMs/Gradle builds etc. Historically for instance we have used e.g. `(,1.0]` and this may no longer be appropriate.

lunaris:
When we generate SDKs or programs, we want the generated code to reference the right version of the core SDK. Generally, this is the latest version, though as the Java language implementation matures we might want to make this a range/minimum version as we have in other languages.

Fixes #1589

### DIFF
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -305,7 +305,12 @@ func GenerateProject(
 	}
 
 	repositories := make(map[string]bool)
-	coreSDKVersion := "(,1.0]"
+
+	// If no version is specified for the pulumi package, use the default SDK version. In either case, presently we emit
+	// a "soft" dependency requirement, which essentially means that it will be used if no other preference is expressed
+	// in the dependency tree. See https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification for
+	// more information.
+	coreSDKVersion := DefaultSdkVersion.String()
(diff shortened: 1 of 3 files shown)


pkg/codegen/java/gen_program.go
```diff
@@ -305,7 +305,7 @@ func GenerateProject(
 	}
 
 	repositories := make(map[string]bool)
-	coreSDKVersion := "(,1.0]"
+	coreSDKVersion := DefaultSdkVersion.String()
```
julienp, line 308: Out of curiosity I looked at https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges and this will now become a "soft" requirement of "1.0.0"

> Default strategy: Of the overlapping ranges, the highest soft requirement is the version to be used.

Might be worth linking to the doc here.

lunaris: Good shout -- have added a link to the Maven docs along with a comment. FWIW I expect us to change this soon to a range, perhaps with 1.0 as the lower bound and HEAD as the upper, maybe.

Issue #1606, opened 2025-01-28
Implement RunPlugin to enable source-based plugins written in Java


mikhailshilkov:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

An example scenario:

- Implement a provider in Java
- Add a PulumiPlugin.yaml with `runtime: java` in it
- Point the CLI to the folder with the provider's source code
- Run a program that instantiates resources from that provider

For this to work, the Java language host has to implement the `RunPlugin` method that would use the right package manager (Maven, Gradle, etc) to compile and execute the Java application. A lot of that startup logic can be reused from the `Run` method.

PR #1608, opened 2025-01-29
Support RunPlugin for Maven and Gradle plugins

mikhailshilkov:
This PR implements RunPlugin for the Java provider host. It's implemented for both Maven and Gradle but not other executors. For now, I assume we don't want folks to author plugins with less popular package managers, but we can also add those later.

To test RunPlugin, I creates a very basic provider gRPC service that implements only GetSchema. My test provider returns a hard-coded schema, and then tests validate the it goes smoothly through the CLI. Later on we can enhance those tests to be real resource providers.

All SDK classes are in the `internal` module.

Resolve https://github.com/pulumi/pulumi-java/issues/1606

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/Provider.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/Provider.java
@@ -4,8 +4,11 @@ import java.util.concurrent.CompletableFuture;
 
 import com.pulumi.provider.internal.models.*;
 
-public abstract class Provider {
-    public CompletableFuture<GetSchemaResponse> getSchema(GetSchemaRequest request) {
-        throw new UnsupportedOperationException("Method 'getSchema' is not implemented");
-    }
+public interface Provider {
+    /**
+     * Returns the schema for this provider's package.
+     * @param request The schema request
+     * @return A future containing the schema response
+     */
+    CompletableFuture<GetSchemaResponse> getSchema(GetSchemaRequest request);
 }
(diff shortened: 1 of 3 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/Provider.java
```diff
@@ -0,0 +1,11 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+import com.pulumi.provider.internal.models.*;
+
+public abstract class Provider {
+    public CompletableFuture<GetSchemaResponse> getSchema(GetSchemaRequest request) {
+        throw new UnsupportedOperationException("Method 'getSchema' is not implemented");
+    }
+}
```
lunaris, line 11: Part of me thinks we should prefer interfaces with default implementations to abstract classes, but I don't know if there are important reasons to/to not to, so happy either way. I think `Provider` was an interface in TS and we found that a bit problematic at times, but TS interfaces don't support default implementations, so maybe that was it. Not sure.

lunaris: It's all internal for now in any case.

lunaris (approved):

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/Provider.java
mikhailshilkov: At the point of this PR, interface sounds like a great idea. I converted it for now, let's see how we want it to evolve later. It's all internal for now, indeed.

Issue #1593, opened 2025-01-15
Add schema `build` config


danielrbradley:
When calling the java-gen program, we used the option `--build gradle-nexus` to also generate the gradle build files. Internally, this set the configuration:

```json
  "buildFiles": "gradle",
  "gradleNexusPublishPluginVersion": "2.0.0",
```

When moving to running java SDK generation via `pulumi package gen-sdk`, we can't set CLI arguments and don't want to have to hard code plugin versions in our schemas.

Therefore, it would be nice to have a new language configuration key to mirror the old CLI argument behaviour so we can just add to our schemas:

```json
  "build": "gradle-nexus",
```

See also: https://github.com/pulumi/pulumi-azure-native/pull/3776

This would also avoid providers pinning to a specific version of the gradle plugin and would allow codegen to seamlessly upgrade this in the future.

PR #1612, opened 2025-01-30
Default to using gradle-nexus in gen-sdk

julienp:
When using `pulumi-java-gen`, the user can bass a `—build` flag to set gradle related settings in the PackageInfo:

| --build            | PackageInfo          | Behaviour                            |
|--------------------|----------------------|--------------------------------------|
|**`none`**          |                      | no gradle file, default              |
|`gradle `           | `buildFiles: gradle` | gradle file without nexus plugin     |
|`gradle-nexus`      | `buildFiles: gradle, gradleNexusPublishPluginVersion: 2.0.0` | gradle file with nexus plugin, using default version of the plugin |
|`gradle-nexus:$VER` | `buildFiles: gradle, gradleNexusPublishPluginVersion: $VER` | gradle file with nexus plugin, using specified version of the plugin |

Pulumi providers use `--build gradle-nexus`.

When using `gen-sdk` instead of `pulumi-java-gen`, you control the gradle generation with the `PackageInfo`. To streamline the usage of the most desired behaviour (`gradle-nexus`), this PR changes the behaviour of the `buildFile` option when using `gen-sdk`. The behaviour when using `pulumi-java-gen` is unchanged. 

This allows providers to not set any options in their schema and still get the best default behaviour.
 
| PackageInfo                    | Behaviour                            |
|--------------------------------|--------------------------------------|
| `buildFiles: none`             | no gradle file                       |
| **`buildFiles: gradle-nexus`** | gradle file with nexus plugin, using default version of the plugin, `2.0.0`, the default |
| `buildFiles: gradle-nexus, gradleNexusPublishPluginVersion: $VER` | gradle file with nexus plugin, using version $VER of the plugin |
| `buildFiles: gradle`           | gradle file without nexus plugin     |
| `buildFiles: gradle, gradleNexusPublishPluginVersion: $VER` | gradle file with nexus plugin, using version $VER of the plugin |

Fixes https://github.com/pulumi/pulumi-java/issues/1593


pkg/codegen/testing/test/testdata/kubernetes20/java/src/main/java/com/pulumi/kubernetes/core/v1/ConfigMap.java
```diff
@@ -1,15 +1,15 @@
 // *** WARNING: this file was generated by test. ***
 // *** Do not edit by hand unless you're certain you know what you are doing! ***
 
-package com.pulumi.kubernetes.core_v1;
+package com.pulumi.kubernetes.core.v1;
```
julienp, line 4: Previously the tests ignored any language specific settings. They now no longer do, and this change comes from the config

```json
    "java": {
      "packages": {
        "core/v1": "core.v1",
        "helm.sh/v3": "helm.v3",
        "meta/v1": "meta.v1"
      }
    },
```

pkg/codegen/java/gen.go
```diff
@@ -2308,38 +2301,46 @@ func GeneratePackage(
...
+	// If the legacyBuildFiles is set to true, we will only use gradle if the
+	// BuildFiles field is explicitly set to `gradle`. This is to support the
+	// legacy behavior of `pulumi-java-gen`. Once `pulumi-java-gen` is
+	// deprecated, we can remove the legacyBuildFiles flag, and always use
+	// gradle if `local` is false.
+	useGradle := !local && // Only use gradle if we are not generating a local package.
+		// legacy behavior requires explicit gradle setting
+		(legacyBuildFiles && info.BuildFiles == "gradle") ||
+		// new behavior uses gradle by default, unless explicitly disabled
+		(!legacyBuildFiles && info.BuildFiles != "none")
```
lunaris, line 2317: I feel like I'd make this an `if`, which although a bit more verbose might be clearer and obviate some of the comment. E.g.:

```go
var useGradle bool
if local {
  // ...
  useGradle = false
} else {
  if legacyBuildFiles {
    // The default for legacy invocations is "none", so we need to see an explicit "gradle" setting
    useGradle = info.BuildFiles == "gradle"
  } else {
    // The default for new invocations is gradle, unless "none" is specified explicitly
    useGradle = info.BuildFiles != "none"
  }
}
```

pkg/codegen/java/gen_test.go
```diff
@@ -224,10 +242,20 @@ func TestGeneratePackage(t *testing.T) {
...
+				if pkg.Language["java"] == nil {
+					pkg.Language["java"] = testCase.packageInfo
+				} else {
+					if err := pkg.ImportLanguages(map[string]schema.Language{"java": Importer}); err != nil {
+						panic(err)
+					}
+					info := pkg.Language["java"].(PackageInfo)
+					pkg.Language["java"] = info.With(testCase.packageInfo)
+				}
+				return GeneratePackage(tool, pkg, extraFiles, nil, false /*local*/, false /*legacyBuildFiles*/)
```
lunaris, line 258: I do wonder if it's possible to find a way to test this set to `true`, since that's the behaviour we want to preserve, but equally it's probably fine?

pkg/codegen/java/templates_gradle.go
```diff
@@ -126,7 +130,20 @@ func newGradleTemplateContext(
 		ctx.Version = pkg.Parameterization.BaseProvider.Version.String()
 	}
 
-	if packageInfo.GradleNexusPublishPluginVersion != "" {
+	if legacyBuildFiles {
```
lunaris, line 133: I wonder if it's worth adding some version of the amazing table in the PR description here; happy either way though.

 
| PackageInfo                    | Behaviour                            |
|--------------------------------|--------------------------------------|
| `buildFiles: none`             | no gradle file                       |
| **`buildFiles: gradle-nexus`** | gradle file with nexus plugin, using default version of the plugin, `2.0.0`, the default |
| `buildFiles: gradle-nexus, gradleNexusPublishPluginVersion: $VER` | gradle file with nexus plugin, using version $VER of the plugin |
| `buildFiles: gradle`           | gradle file without nexus plugin     |
| `buildFiles: gradle, gradleNexusPublishPluginVersion: $VER` | gradle file with nexus plugin, using version $VER of the plugin |

lunaris (approved):
Awesome stuff! And amazing PR description :purple_heart:

pkg/codegen/java/gen.go
julienp: Yeah that's nicer!

pkg/codegen/java/gen_test.go
julienp: I updated `TestGenGradleProject` to check the generated file for both legacy falls and true variants.

Issue #1335, opened 2024-03-26
Add ability to build multi-language components in Java


pierskarsenbarg:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

I couldn't see a specific issue for MLCs in Java. It is mentioned in [this issue](https://github.com/pulumi/pulumi-java/issues/11) but I don't want it to get lost.

mikhailshilkov:
Component Packages will be supported in the 1.6.0 release

PR #1616, opened 2025-01-31
Inline runner machinery

mikhailshilkov:
The use case is to be able to create resources without an arbitrary function that is created inside a provider (`Construct` in particular). All methods are marked internal and are intended to be consumed from our own SDK.

Based on our .NET implementation in https://github.com/pulumi/pulumi-dotnet/tree/main/sdk/Pulumi/Deployment

Towards https://github.com/pulumi/pulumi-java/issues/1335

No testing yet - but opening it up also for @justinvp to take a look in case there may be any overlap with Automation API work.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/InlineDeploymentSettings.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/InlineDeploymentSettings.java
@@ -6,9 +6,12 @@ import com.google.common.collect.ImmutableSet;
 import javax.annotation.Nullable;
 import java.util.Optional;
 
+import com.pulumi.core.internal.annotations.InternalUse;
+
 /**
  * Settings for creating an inline deployment.
  */
+@InternalUse
 public class InlineDeploymentSettings {
     private final String monitorAddr;
     private final String engineAddr;
--- a/sdk/java/pulumi/src/main/java/com/pulumi/Pulumi.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/Pulumi.java
@@ -1,14 +1,10 @@
 package com.pulumi;
 
-import com.pulumi.core.internal.annotations.InternalUse;
-import com.pulumi.deployment.internal.InlineDeploymentSettings;
-import com.pulumi.deployment.internal.Runner.Result;
 import com.pulumi.internal.PulumiInternal;
 import com.pulumi.resources.StackOptions;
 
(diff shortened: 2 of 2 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentImpl.java
```diff
@@ -164,6 +165,14 @@ public static DeploymentImpl fromEnvironment() {
         return impl;
     }
 
+    @InternalUse
+    public static DeploymentImpl fromInline(InlineDeploymentSettings settings) {
+        var state = DeploymentState.fromInline(settings);
+        var impl = new DeploymentImpl(state);
+        DeploymentInstanceHolder.setInstance(new DeploymentInstanceInternal(impl));
```
justinvp, line 172: Have you tried multiple calls to `Construct` within the same process (e.g. creating two separate instances of an MLC from the Pulumi program)?

I am hitting the following when I have multiple operations that involve the inline program (e.g. a preview and up) in the same process:

https://github.com/pulumi/pulumi-java/blob/85e505ddcadbd6c674b36c61af1dc08ad5e49b69/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceHolder.java#L39

I haven't dug in to see if/how .NET avoids it.

mikhailshilkov: I confirm I can see the same, thank you. .NET seems to work fine - I'll dig further.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/InlineDeploymentSettings.java
```diff
@@ -0,0 +1,128 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+
+/**
+ * Settings for creating an inline deployment.
+ */
+public class InlineDeploymentSettings {
```
justinvp, line 12:
```suggestion
@InternalUse
public class InlineDeploymentSettings {
```

justinvp (approved):
A couple comments. I need this for Auto API, so I'm going to address my two comments in a commit and go ahead and merge this. But I'm happy to discuss, if you think we should revert any of my tweaks.

PR #1618, opened 2025-02-03
auto: Initial interfaces and supporting types

justinvp:
This adds the abstract `Workspace` class and associated supporting classes for the Java Automation API. Currently, these are going under `com.pulumi.experimental.automation`, but the intention is to move everything to `pulumi.com.automation` once all the support has landed and is stable.

Next up will be `LocalWorkspace` and `WorkspaceStack`, and supporting code, which contains the most of the implementation, along with tests.


sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/exceptions/AutomationException.java
```diff
@@ -0,0 +1,17 @@
+// Copyright 2025, Pulumi Corporation
+
+package com.pulumi.experimental.automation.exceptions;
+
+public class AutomationException extends Exception {
```
justinvp, line 5: This is the primary checked exception that we expose in signatures. There are several subclasses in both this package as well as `automation.commands.exceptions`.

I've gone back and forth in my mind whether these should be in separate `.exceptions` packages. This is what the .NET Automation API does: It puts the exceptions in separate [`.Exceptions`](https://github.com/pulumi/pulumi-dotnet/tree/main/sdk/Pulumi.Automation/Exceptions) and [`.Commands.Exceptions`](https://github.com/pulumi/pulumi-dotnet/tree/main/sdk/Pulumi.Automation/Commands/Exceptions) namespaces. However, I'm not sure that was intentional. In the core .NET SDK, there are some exceptions that are placed inside an [`Exceptions`](https://github.com/pulumi/pulumi-dotnet/tree/main/sdk/Pulumi/Exceptions) subdirectory, but the actual namespace for the types remains `Pulumi` not `Pulumi.Exceptions`. So I'm wondering if the reason we have `Pulumi.Automation.Exceptions` and `Pulumi.Automation.Commands.Exceptions` is an accident of the IDE filling-in the namespace based on the directory structure, when the intention was to just have these exceptions go in `Pulumi.Automation` and `Pulumi.Automation.Commands`.

When the core .NET SDK was ported to Java, the exceptions were put inside an [exceptions](https://github.com/pulumi/pulumi-java/tree/main/sdk/java/pulumi/src/main/java/com/pulumi/exceptions) subdirectory, which happened to also place them in a `.exceptions` subpackage. Not sure if it was intentional.

All that being said, Claude tells me that some Java libraries do place exceptions in a `.exceptions` subpackage, if there are several exception types and a more complex inheritance hierarchy. So I've kept it this way.

But I've considered alternatives, like placing them all in `.automation` and `.automation.commands`, or putting the base `AutomationException` in `.automation`.

Curious if others have a strong opinion about it.

justinvp, line 5: Also note: the other SDKs don't have an `AutomationException`. The base exception type in the other SDKs is [`CommandException`](https://github.com/pulumi/pulumi-java/pull/1618/files#diff-5c4630475530817b84cabf8d7bcee9df81b2a5ae695d3a052c6421c0bff3eff4). However, since we actually have to put the exception in the signature (e.g. `throws AutomationException`) for most methods, it felt better to use a more generic "AutomationException" rather than using `throws CommandException`, for two reasons:
1. There could be an error that isn't `CommandException` (that will be wrapped by `AutomationException`)
2. If we move to another communication mechanism under the covers (instead of CLI invocations), having the error be `AutomationException` gives us flexibility.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/commands/exceptions/CommandException.java
```diff
@@ -0,0 +1,12 @@
+// Copyright 2025, Pulumi Corporation
+
+package com.pulumi.experimental.automation.commands.exceptions;
+
+import com.pulumi.experimental.automation.commands.CommandResult;
+import com.pulumi.experimental.automation.exceptions.AutomationException;
+
+public class CommandException extends AutomationException {
```
justinvp, line 8: I made this a subclass of `AutomationException` for convenience, but thinking about it more now, I'm wondering if it should just subclass Java's `Exception` and if we throw `CommandException` wrap that in an `AutomationException`, e.g. `catch (CommandException e) { throw new AutomationException(e); }`. 

Also now thinking the other exceptions that subclass `CommandException` probably should subclass `AutomationException` directly, and probably not live in `commands.exceptions` but next to wherever `AutomationException` lives.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/Workspace.java
```diff
@@ -0,0 +1,569 @@
...
+     * Sets the specified key-value pair in the provided stack's config.
+     *
+     * @param stackName the name of the stack to operate on
+     * @param key       the config key to set
+     * @param value     the config value to set
+     * @param path      the key contains a path to a property in a map or list to
+     *                  set
+     * @throws AutomationException if there was an issue setting the config
+     */
+    public abstract void setConfig(String stackName, String key, ConfigValue value, boolean path)
```
justinvp, line 287: Should we proactively expose `--type` (https://github.com/pulumi/pulumi/pull/18287) or at least have a plan for what it would look like if we're not adding it now?

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/exceptions/AutomationException.java
lunaris: Personally I'd put them in `com.pulumi.automation` -- packages for me are about semantic separation, and this is an exception that gets thrown by functionality in that package. I'm not even sure the `.commands` package is super useful, but I guess that depends how much we end up with `.automation`.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/OperationType.java
```diff
@@ -0,0 +1,32 @@
...
+    READ(7),
+    READ_REPLACEMENT(8),
+    REFRESH(9),
+    READ_DISCARD(10),
+    DISCARD_REPLACED(11),
+    REMOVE_PENDING_REPLACE(12),
+    IMPORT(13),
+    IMPORT_REPLACEMENT(14);
+
+    private final int value;
```
lunaris, line 23: Do you need the `value`? I can't see anywhere it's used (maybe it's coming in a future PR, or I've missed it like an idiot), and I don't think you need it if all you care about is e.g. comparing the enum values.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/PluginInfo.java
```diff
@@ -0,0 +1,91 @@
...
+public class PluginInfo {
+    private final String name;
+    private final PluginKind kind;
+    @Nullable
+    private final String version;
+    private final long size;
+    private final Instant installTime;
+    private final Instant lastUsedTime;
+
+    PluginInfo(
```
lunaris, line 21: Any reason this is `package` visibility and not `public`?

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/ProjectPlugins.java
```diff
@@ -0,0 +1,142 @@
...
+
+import javax.annotation.Nullable;
+
+/**
+ * A set of plugins configured for a Pulumi project.
+ */
+public class ProjectPlugins {
+    private final List<ProjectPluginOptions> providers;
+    private final List<ProjectPluginOptions> languages;
+    private final List<ProjectPluginOptions> analyzers;
```
lunaris, line 17: Guessing we don't "do" converters in automation API?

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/ProjectRuntimeOptions.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Objects;
+
+import javax.annotation.Nullable;
+
+/**
+ * Various configuration options that apply to different language runtimes.
+ */
+public class ProjectRuntimeOptions {
+    @Nullable
+    private final Boolean typescript;
```
lunaris, line 12: I guess this is boxed because we want to observe `null` vs `false`?

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/commands/PulumiCommand.java
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * @param args    the arguments to pass to the command
+     * @param options the options for running the command
+     * @return the command result
+     * @throws AutomationException if the command fails
+     */
+    public abstract CommandResult run(
+            List<String> args,
+            CommandRunOptions options) throws AutomationException;
+}
```
lunaris, line 42: Better as an interface with default implementations?

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/events/CancelEvent.java
```diff
@@ -0,0 +1,10 @@
+// Copyright 2025, Pulumi Corporation
+
+package com.pulumi.experimental.automation.events;
+
+/**
+ * {@link CancelEvent} is emitted when the user initiates a cancellation of the
+ * update in progress, or the update successfully completes.
+ */
+public class CancelEvent {
```
lunaris, line 9: Do events need any common supertype other than `Object`?

lunaris (approved):
Awesome stuff! I have a bunch of questions and comments but none feel major/this is all going in `.experimental` for now so happy to keep moving!

PR #1624, opened 2025-02-06
auto: Implementation and initial tests

justinvp:
This change adds `LocalWorkspace`, `WorkspaceStack`, and `LocalPulumiCommand`, and supporting types, which provides the bulk of the implementation. Apologies for the large PR. An initial set of tests is included, but more need to be added.

Beyond adding more tests, there are some known issues that I'm currently working on to be done as separate PRs:
- ~Bug: `NoSummaryEventException` is thrown when running previews on local programs~ Fixed!
- https://github.com/pulumi/pulumi-java/issues/1641
- https://github.com/pulumi/pulumi-java/issues/1628
- https://github.com/pulumi/pulumi-java/issues/1630
- https://github.com/pulumi/pulumi-java/issues/1632
- https://github.com/pulumi/pulumi-java/issues/1633

There's also some follow-up fixes / enhancements that we want to consider, subsequently:

- https://github.com/pulumi/pulumi-java/issues/1634
- https://github.com/pulumi/pulumi-java/issues/1629
- https://github.com/pulumi/pulumi-java/issues/1631
- https://github.com/pulumi/pulumi-java/issues/1636
- https://github.com/pulumi/pulumi-java/issues/1635
- https://github.com/pulumi/pulumi-java/issues/1637
- https://github.com/pulumi/pulumi-java/issues/1638
- https://github.com/pulumi/pulumi-java/issues/1639
- https://github.com/pulumi/pulumi-java/issues/1640

Part of #1607

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/experimental/automation/LocalWorkspaceTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/experimental/automation/LocalWorkspaceTest.java
@@ -6,6 +6,7 @@ import com.pulumi.Context;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.condition.DisabledIfEnvironmentVariable;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import java.nio.file.Path;
@@ -33,6 +34,8 @@ public class LocalWorkspaceTest {
         return result.toString();
     }
 
+    // Temporarily disable when running in CI.
+    @DisabledIfEnvironmentVariable(named = "GITHUB_ACTIONS", matches = "true")
     @Test
     @Timeout(value = 10, unit = TimeUnit.MINUTES)
     void testStackLifecycleInlineProgram(@EnvVars Map<String, String> envVars) {
@@ -105,6 +108,8 @@ public class LocalWorkspaceTest {
         });
     }
 
+    // Temporarily disable when running in CI.
+    @DisabledIfEnvironmentVariable(named = "GITHUB_ACTIONS", matches = "true")
     @Test
     @Timeout(value = 10, unit = TimeUnit.MINUTES)
     void testStackLifecycleLocalProgram(@EnvVars Map<String, String> envVars) {


sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/UpdateOptions.java
```diff
@@ -0,0 +1,432 @@
...
+     * @return the JSON output setting
+     */
+    public boolean isJson() {
+        return json;
+    }
+
+    /**
+     * Base builder for all update options.
+     */
+    public static abstract class Builder<B extends Builder<B>> {
```
lunaris, line 216: We don't have to do this pre-go-live but I think we should discuss not doing this when this out the door. I don't think the amount of code saved by inheriting e.g. `DestroyOptions` from `UpdateOptions` and the clever `Builder` inheritance tricks is _super_ worth it. I appreciate that we already do this in e.g. `ResourceOptions`, though it's a bit messy there and has made certain kinds of changes fiddly in the past.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/LocalWorkspace.java
```diff
@@ -0,0 +1,1115 @@
...
+        if (!existingSettings.isPresent()) {
+            saveProjectSettings(settings);
+        } else if (!existingSettings.get().isDefault() && !existingSettings.get().equals(settings)) {
+            var path = findSettingsFile();
+            throw new ProjectSettingsConflictException(path.toString());
+        }
+    }
+
+    /**
+     * Creates a workspace using the specified options. Used for maximal control and
```
lunaris, line 128: I think this comment is wrong; there are no options passed in this case.

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/LocalWorkspace.java
```diff
@@ -0,0 +1,1115 @@
...
+        var hasSkipEnvVar = environmentVariables != null &&
+                environmentVariables.containsKey(LocalPulumiCommand.SKIP_VERSION_CHECK_VAR);
+        var optOut = hasSkipEnvVar || System.getenv(LocalPulumiCommand.SKIP_VERSION_CHECK_VAR) != null;
+        return optOut;
+    }
+
+    /**
+     * Creates a stack with a {@link LocalWorkspace} utilizing the specified inline
+     * (in process) {@code program}. This program is fully debuggable and runs in
+     * process. If no {@link LocalWorkspaceOptions#getProjectSettings()} option is
```
lunaris, line 174: Again this comment is wrong I think -- no options here. There may be others too; I've probably missed some overloads. These kinds of things I don't mind leaving to post though.

lunaris (approved):

PR #1642, opened 2025-02-06
auto: Investigate CI timeouts

justinvp:
Re-enable a pared down test to uncover what's causing the timeout.


justinvp:
Found the issue. Opening a separate PR with the fix.

PR #1645, opened 2025-02-07
Configure and Construct implementation in the provider service

mikhailshilkov:
This PR implements Configure and Construct in the provider protocol - enough to enable very basic Component Packages. Serializing and deserializing to/from PropertyValues has to be done by component authors. (utilities for that are coming in later PRs)

Both Maven and Gradle provider tests are changed from a basic provider returning a static schema to a functional Component provider with a basic component. Tests deploy those with ProgramTest and verify the output shape.

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/resources/CustomTimeoutsTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/resources/CustomTimeoutsTest.java
@@ -44,6 +44,8 @@ class CustomTimeoutsTest {
                     assertThat(CustomTimeouts.parseTimeoutString("0")).isEqualTo(Duration.ZERO)),
 
             // Single unit cases
+            dynamicTest("zero seconds", () ->
+                    assertThat(CustomTimeouts.parseTimeoutString("0s")).isEqualTo(Duration.ZERO)),
             dynamicTest("nanoseconds", () ->
                     assertThat(CustomTimeouts.parseTimeoutString("100ns")).isEqualTo(Duration.ofNanos(100))),
             dynamicTest("microseconds", () ->


sdk/java/pulumi/src/main/java/com/pulumi/resources/CustomTimeouts.java
```diff
@@ -92,6 +92,78 @@ public static String golangString(Optional<Duration> duration) {
...
+                duration = duration.plusNanos((long)v);
+                s = s.substring(2);
+            } else if (s.startsWith("µs") || s.startsWith("us")) {
+                duration = duration.plusNanos((long)(v * 1000));
+                s = s.substring(2);
+            } else if (s.startsWith("ms")) {
+                duration = duration.plusNanos((long)(v * 1_000_000));
+                s = s.substring(2);
+            } else if (s.startsWith("s")) {
+                duration = duration.plusNanos((long)(v * 1_000_000_000));
```
tgummerer, line 149:
```suggestion
                duration = duration.plusSeconds((long)v);
```

Would be easier to read I think.

sdk/java/pulumi/src/test/java/com/pulumi/resources/CustomTimeoutsTest.java
```diff
@@ -31,4 +32,67 @@ Stream<DynamicNode> testGolangString() {
...
+    @TestFactory
+    Stream<DynamicNode> testParseTimeoutString() {
+        return Stream.of(
+            // Basic cases
+            dynamicTest("null input", () ->
+                    assertThat(CustomTimeouts.parseTimeoutString(null)).isNull()),
+            dynamicTest("empty input", () ->
+                    assertThat(CustomTimeouts.parseTimeoutString("")).isNull()),
+            dynamicTest("zero", () ->
+                    assertThat(CustomTimeouts.parseTimeoutString("0")).isEqualTo(Duration.ZERO)),
```
tgummerer, line 44:
```suggestion
                    assertThat(CustomTimeouts.parseTimeoutString("0")).isEqualTo(Duration.ZERO)),
                    assertThat(CustomTimeouts.parseTimeoutString("0s")).isEqualTo(Duration.ZERO)),
```

I think this should just work, but might be worth an additional test case.

tgummerer (approved):

sdk/java/pulumi/src/main/java/com/pulumi/resources/CustomTimeouts.java
mikhailshilkov: Yeah, but we lose precision if we round v to a long here (and plusSeconds accepts a long).

tgummerer: Ah because the value can be with commas, I missed that.  Yeah `plusNanos` makes sense them.

PR #1646, opened 2025-02-07
auto: Fix inline programs timing out in CI

### LINKED ISSUE #1641: auto: Inline program lifecycle tests are timing out in CI
The inline ~and local~ program lifecycle tests are timing out in CI, so they were temporarily disabled. Need to investigate why they are timing out (they're relatively quick <20 seconds on my machine) and re-enable the tests.

justinvp:
Fixes #1641

Three commits:

## 1. Don't block thread pool threads

Don't tie up shared thread pool threads blocked on reading stdout/stderr when running commands, which can lead to deadlocks running inline programs when no thread pool threads are available to make forward progress.

Similarly, when reading the event log, we can use a single thread as to not block a shared thread pool thread.

We also don't need to write stdin using a thread.

## 2. Don't block in the gRPC language server's `run` method

Instead of blocking, we can complete the request by calling `onNext` and `onCompleted` asynchronously from a continuation after running the inline program.

## 3. Unskip `testStackLifecycleInlineProgram` test

Also reduce the timeout to 1 minute in both lifecycle tests.

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/experimental/automation/LocalWorkspaceTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/experimental/automation/LocalWorkspaceTest.java
@@ -6,7 +6,6 @@ import com.pulumi.Context;
 
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
-import org.junit.jupiter.api.condition.DisabledIfEnvironmentVariable;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


tgummerer (approved):

sdk/java/pulumi/src/main/java/com/pulumi/experimental/automation/LanguageRuntimeImpl.java
```diff
@@ -63,23 +64,33 @@ public void run(RunRequest request,
...
+                        }
+                        responseObserver.onNext(responseBuilder.build());
+                        responseObserver.onCompleted();
+                        return null;
+                    } finally {
+                        DeploymentImpl.internalUnsafeDestroyInstance();
+                        semaphore.release();
+                    }
+                });
+            } catch (Exception e) {
```
davidemassarenti-optio3, line 86: Is there a reason why this is not just a finally, removing the need for the nested try/finally?

justinvp: If it was a finally here, I think it'll be called too soon, because `runAsync(program).handle(...)` is async, and will return immediately. I want to release the lock after the inline program is finished running.

PR #1653, opened 2025-02-10
Deserialization of PropertyValue into resource args

mikhailshilkov:
This PR implements a utility class to convert PropertyValue's into Java classes, particularly resource arg classes. This is handy for many provider SDK methods, but for now is especially relevant for more hands-off implementations of component packages via Construct (see two changed examples).

Largely ported from .NET https://github.com/pulumi/pulumi-dotnet/blob/main/sdk/Pulumi/Provider/PropertyValueSerializer.cs but Java reflection is different enough from .NET for the code to become fairly different. I tried to cover it with a decent set of unit tests.

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/PropertyValueSerializerTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/PropertyValueSerializerTest.java
@@ -27,7 +27,7 @@ import java.util.Set;
 
 public class PropertyValueSerializerTest {
 
-   static class BasicArgs extends ResourceArgs {
+    static class BasicArgs extends ResourceArgs {
         @Import(name="intValue", required=true)
         private Output<Integer> intValue;
 
@@ -42,6 +42,9 @@ public class PropertyValueSerializerTest {
 
         @Import(name="numberValue", required=true)
         private Output<Number> numberValue;
+
+        @Import(name="valueWithOverridenName")
+        private Output<String> otherStringValue;
 
         public Output<Integer> intValue() {
             return this.intValue;
@@ -62,6 +65,10 @@ public class PropertyValueSerializerTest {
         public Output<Number> numberValue() {
             return this.numberValue;
         }
+
+        public Output<String> otherStringValue() {
+            return this.otherStringValue;
+        }
     
         private BasicArgs() {}
     }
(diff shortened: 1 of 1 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyDeserializationException.java
```diff
@@ -0,0 +1,56 @@
+package com.pulumi.provider.internal.properties;
```
tgummerer, line 1: Not necessarily something we need to fix in this PR, but I'm curios, do we not do copyright/license headers for `java` files in this repository?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValueSerializer.java
```diff
@@ -0,0 +1,359 @@
...
+
+    private static Object deserializeValue(PropertyValue value, java.lang.reflect.Type targetType) {
+        String[] rootPath = new String[]{"$"};
+        return deserializeValue(value, targetType, rootPath);
+    }
+
+    private static Object deserializeValue(PropertyValue value, java.lang.reflect.Type targetType, String[] path) {
+        Class<?> rawType = resolveGenericType(targetType);
+        System.out.printf("deserializeValue: value.type=%s, targetType=%s, rawType=%s%n", 
+            value.getType(), targetType.getTypeName(), rawType.getName());
```
tgummerer, line 72: debugging leftovers?  Should this be logged instead?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyDeserializationException.java
justinvp: It's lacking throughout. I added a copyright header to all the new auto API files. But there's no tooling yet to enforce it being there and in a consistent format. Opened https://github.com/pulumi/pulumi-java/issues/1658

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValueSerializer.java
mikhailshilkov: yeah, thank you for catching!

Issue #30, opened 2022-02-01
Consider removing global state from the API


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Global state is inherited from the C# translation. It primarily is a Deployment instance object and, indirectly, Stack. If this can be avoided by explicit param passing or other means, we get the ability for several Pulumi deployments coexisting peacefully in the same JVM process. While we do not know use cases for this yet, it may become desirable someday - just a general design sense.

Notes from Pawel (from #1):

```

- global state of the deployment instance is essentially a god object,
  but there might be ways to avoid that, e.g. by using a state machine
  to separate uninitialized deployment and initialized deployment as
  separate types, here is the code:
  https://github.com/pulumi/pulumi-java/blob/d302f9c049ac7a9653d56e9073ca3f091be0ace9/sdk/jvm/pulumi/src/main/java/io/pulumi/deployment/internal/DeploymentInstanceHolder.java#L16

- related to the above, Stack has this in the constructor:
https://github.com/pulumi/pulumi-java/blob/d302f9c049ac7a9653d56e9073ca3f091be0ace9/sdk/jvm/pulumi/src/main/java/io/pulumi/Stack.java#L80

- C# has AsyncLocal, as an alternative, we could look into one of those:
    - https://github.com/eclipse/microprofile-context-propagation
    - https://github.com/tunnelvisionlabs/java-threading/blob/c4fe47cfb25f724f897aa2cf6ea63d9c95aecee4/src/com/tunnelvisionlabs/util/concurrent/AsyncLocal.java

```

cowwoc:
FYI, the aforementioned PR was closed on May 2022. There doesn't seem to be any progress on this issue.

t0yv0:
You can help us prioritize this by upvoting and explaining your use case! Thank you!

mikhailshilkov:
This issue is relevant again, so I will try to provide some context.

Pulumi SDKs have a notion of Deployment, that helps coordinate communications with the engine and tracks deployment progression. As two examples: every resource constructor invokes a `registerResource` RPC call, and every Output<T> needs to be tracked until it resolves. Both of those flows will use a Deployment to "register" itself into a single coherent program execution flow.

[...]

PR #1664, opened 2025-02-12
Serialization of a Component Resource outputs into PropertyValues

mikhailshilkov:
The counterpart to https://github.com/pulumi/pulumi-java/pull/1653

This PR implements a utility class to convert Java resource classes' outputs into PropertyValue's. This is handy for many provider SDK methods, but for now is especially relevant for more hands-off implementations of component packages via Construct (see two changed examples).

Largely ported from .NET https://github.com/pulumi/pulumi-dotnet/blob/main/sdk/Pulumi/Provider/PropertyValueSerializer.cs.

This brings Java MLCs on par with .NET ones, modulo https://github.com/pulumi/pulumi-java/issues/30

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValue.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValue.java
@@ -131,10 +131,6 @@ public class PropertyValue {
         return new PropertyValue(ValueType.NUMBER, value);
     }
 
-    public static PropertyValue of(int value) {
-        return new PropertyValue(ValueType.NUMBER, Integer.valueOf(value));
-    }
-
     public static PropertyValue of(String value) {
         return new PropertyValue(ValueType.STRING, value);
     }


sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValueSerializer.java
```diff
@@ -366,4 +371,211 @@ private static String propertyName(Field field, Class<?> declaringType) {
...
+            return PropertyValue.NULL;
+        }
+
+        // Handle primitive types and their wrappers
+        if (value instanceof String) {
+            return PropertyValue.of((String) value);
+        }
+
+        if (value instanceof Integer) {
+            return PropertyValue.of(((Integer) value).doubleValue());
```
tgummerer, line 498: Shouldn't this be 

```suggestion
            return PropertyValue.of(((Integer) value).intValue());
```

?  The property value type will still be "Number", but the value object will be of type Integer if we use `.intValue()`, which I think makes a difference when we deserialize the value.  My java knowledge is pretty outdated though, so I might be missing something.

sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/PropertyValueSerializerTest.java
```diff
@@ -413,6 +419,201 @@ void testDeserializingAssetAndArchiveWorks() {
         assertThat(archive).isEqualTo(fileArchive);
     }
 
+    class BasicResource  {
```
tgummerer, line 422: Might be nice to have a test that serializes and then deserializes a resource, checking that we get the same thing back where possible?

tgummerer (approved):

sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/PropertyValueSerializerTest.java
mikhailshilkov: This isn't directly possible because we deserialize input Args classes but serialize resource classes. They aren't compatible - e.g. using Import vs Export decorators.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValueSerializer.java
tgummerer: I think the difference would be [here](https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValue.java#L134-L136), where if we pass in an int value we store an `Integer` object here, which you can get back out [here](https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValue.java#L171).  Though for [getNumberValue](https://github.com/pulumi/pulumi-java/blob/main/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/properties/PropertyValue.java#L203) we always return doubles, so maybe this is right as it is.

tgummerer: Though maybe then there's a bug in `PropertyValue of(int value)`, and it should be `return new PropertyValue(ValueType.NUMBER, Double.valueOf(value));` instead :thinking:

mikhailshilkov: Good point. PropertyValue of(int value) is actually currently never used, so I'll just remove it.

tgummerer: Sounds good to me!

PR #1665, opened 2025-02-12
Add ability to disable toString on Output<T>

justinvp:
This change provides an ability that is already present in the Python and Node.js SDKs, and was recently added to the .NET SDK, to to disable converting `Output<T>` to a string.

Currently in the Java SDK this will return a helpful message in the string contents to advise not to do this but does not fail the Pulumi program.

This functionality is opted-in by setting the environment variable `PULUMI_ERROR_OUTPUT_STRING` to `true` or `1` to throw an exception.

Port of https://github.com/pulumi/pulumi-dotnet/pull/461

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,5 @@
 ### Improvements
 
+- Add ability to disable toString on Output<T>
+
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


justinvp:
https://github.com/pulumi/pulumi-java/pull/1666 fixes the Go lint issue (needs upgrade for go1.24).

Frassle (approved):

PR #1668, opened 2025-02-14
Infer schema from Component resource classes

mikhailshilkov:
Implements `ComponentAnalyzer.generateSchema` which takes a list of component resources and produces a serializable package schema strcture. Infers component resources and their dependent types. Here are the features that this PR implements and tests:

- [x] Inputs
- [x] Outputs
- [x] Plain types
- [x] Boolean, integer, number, string
- [x] Complex types
- [x] Nested complex types
- [x] Self-recursive complex types
- [x] Mutually recursive complex types
- [x] Required properties & inputs
- [ ] Component description
- [ ] Complex type description
- [ ] Property description
- [x] Optional
- [x] Array
- [x] Map
- [x] Asset/Archive
- [ ] Resource reference
- [ ] Discriminated Union
- [ ] Enum
- [ ] Methods
- [ ] References to component in same provider
- [ ] pulumi.Any

An existing component provider example is updated to use a generated schema instead of a hard-coded JSON file.

Resolve https://github.com/pulumi/pulumi-java/issues/1674 (we'll track missing features in separate issues)

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
@@ -315,12 +315,12 @@ public final class ComponentAnalyzer {
         if (type instanceof Class<?>) {
             Class<?> clazz = (Class<?>) type;
             if (isBuiltinType(clazz)) {
-                return TypeSpec.ofBuiltin(getBuiltinTypeName(clazz), isOutput ? null : true);
+                return TypeSpec.ofBuiltin(getBuiltinTypeName(clazz), !isOutput);
             } 
             
             var specialTypeRef = getSpecialTypeRef(clazz);
             if (specialTypeRef.isPresent()) {
-                return TypeSpec.ofRef(specialTypeRef.get(), isOutput ? null : true);
+                return TypeSpec.ofRef(specialTypeRef.get(), !isOutput);
             }
 
             if (isResourceType(clazz)) {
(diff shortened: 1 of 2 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -0,0 +1,355 @@
...
+        
+        if (type instanceof Class<?>) {
+            Class<?> clazz = (Class<?>) type;
+            if (isBuiltinType(clazz)) {
+                return TypeSpec.ofBuiltin(getBuiltinTypeName(clazz), isOutput ? null : true);
+            } 
+            
+            var specialTypeRef = getSpecialTypeRef(clazz);
+            if (specialTypeRef.isPresent()) {
+                return TypeSpec.ofRef(specialTypeRef.get(), isOutput ? null : true);
```
tgummerer, line 323: Is this a common pattern for Java to pass `null` for `Boolean`s here rather than false? Iow, couldn't we just pass `!isOutput` here rather than having this ternary operator everywhere?

tgummerer (approved):
Nice!

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
mikhailshilkov: Great question. Honestly, I don’t know what is idiomatic in Java. But my goal here is not to add plain:false into the schema but leave it out instead (for cleanness). I’ll double-check if there is an option to omit default values in serialization.

tgummerer: Ah that makes sense.  Thanks for the explanation!

PR #1671, opened 2025-02-18
Update pulumi submodule to master and use the submodule for pulumi-java

brandonpollack23:
This way for testing on pull-recursive you always have the latest
submitted pulumi/pulumi for tests and you can run against it instead of
pulling a released version.

The logic here is we develop against head and have to manually do this,
so why not leave the repository in that state and change the release
process to check out a specific SDK if necessary?


brandonpollack23:
we discussed this in ideation and decided to go with rennovate bot to update pulumi/pulumi submodule's pointer and the go.mod version to always point to the current master externally, as well as open a PR for it so all the tests (and consequently any new conformance tests) run and fail, letting us know to implement or bypass the test.

PR #1676, opened 2025-02-18
Cleanup fixed l2-map-keys conformance test.

brandonpollack23:
bug is resolved and it passed for me locally, letting run in presubmit and if it passes merging


brandonpollack23:
I guess this was a mistake, still failing

PR #1678, opened 2025-02-18
Dev/davide/context aware fixup

davidemassarenti-optio3:
The current code was relying on a static field to track the Deployment context. 
This meant that it was not possible to run parallel deployments.

These changes swap the static field with thread-local storage. Because the system heavily depends on asynchronous operations, and the runtime controls which thread executes what callback, we need to inject the correct state into the thread at each reentrancy point.

ContextAwareCompletableFuture<T> is a subclass of CompletableFuture<T> which exposes the same helper methods, but captures and reapplies the deployment state on restart.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/DeploymentInstance.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/DeploymentInstance.java
@@ -10,5 +10,9 @@ public interface DeploymentInstance extends Deployment {
 
     boolean isInvalid();
 
+    /**
+     * We store the context in thread local storage, and we also capture it in completable futures.
+     * Calling this method makes sure async callbacks that complete after the core program loop completes do not access the stale context.
+     */
     void markInvalid();
 }


sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/ResourceProviderService.java
```diff
@@ -140,78 +150,83 @@ public void configure(pulumirpc.Provider.ConfigureRequest request,
...
+
+                    domResponse.getStateDependencies().forEach((propertyName, dependencies) -> {
+                        var propertyDeps = pulumirpc.Provider.ConstructResponse.PropertyDependencies.newBuilder()
+                            .addAllUrns(dependencies)
+                            .build();
+                        responseBuilder.putStateDependencies(propertyName, propertyDeps);
+                    });
+
+                    var grpcResponse = responseBuilder.build();
+                    responseObserver.onNext(grpcResponse);
```
mikhailshilkov, line 221: Don't we need to clean up DeploymentInstanceHolder after DeploymentImpl.fromInline completes? It could be that construct would be called twice sequentially on the same thread?

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
```diff
@@ -0,0 +1,650 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+@InternalUse
+public class ContextAwareCompletableFuture<T> extends CompletableFuture<T> {
```
mikhailshilkov, line 21: Can user code mess it up if they use a non-context-aware CompletableFuture in their code? What are the failure modes, if any?

mikhailshilkov (commented):
Looks promising! A couple of high-level questions to make sure I understand the implications.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceHolder.java
```diff
@@ -6,43 +6,69 @@
...
-import java.util.concurrent.atomic.AtomicReference;
+
+import static com.pulumi.core.internal.Exceptions.getStackTrace;
 
 @InternalUse
 public abstract class DeploymentInstanceHolder {
 
-    private static final AtomicReference<DeploymentInstance> instance = new AtomicReference<>();
+    private static final ThreadLocal<DeploymentInstance> instance = new ThreadLocal<>();
+    private static final ThreadLocal<String> instance2 = new ThreadLocal<>();
```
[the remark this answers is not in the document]
davidemassarenti-optio3: I removed the logic completely, because we can’t control the full lifecycles of threads in a pool (thus we don’t get a chance to cleanup its state) and junit seems to use a thread pool and one test was randomly failing.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
```diff
@@ -0,0 +1,650 @@
...
+    @Override
+    public T get(long timeout,
+                 TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+        DeploymentInstanceHolder.setInstanceNoThrow(context);
+        return super.get(timeout, unit);
+    }
+
+    @Override
+    public T join() {
+//        DeploymentInstanceHolder.setInstanceNoThrow(context);
```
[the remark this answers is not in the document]
davidemassarenti-optio3: Testing leftover, removed

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/ResourceProviderService.java
davidemassarenti-optio3: We can't really control all the path through threads, we don't know when the last consumer is done

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
davidemassarenti-optio3: The only way to mess things up is to manually queue a callback with a thread pool. I've aded comments here to explain the pattern.

sdk/java/pulumi/src/main/java/com/pulumi/automation/LanguageRuntimeImpl.java
```diff
@@ -67,6 +67,7 @@ public void run(RunRequest request,
             // TODO Remove lock once https://github.com/pulumi/pulumi-java/issues/30 is
             // resolved.
```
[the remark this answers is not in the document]
davidemassarenti-optio3: If the semaphore was only used to block access to deployment holder, yes. I removed the logic, unit tests and pumuli preview still happy

sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/DeploymentInstanceHolder.java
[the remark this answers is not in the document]
davidemassarenti-optio3: Added invalidation logic, it should cover some of these checks

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/ResourceProviderService.java
davidemassarenti-optio3: Added invalidation logic, it should cover this

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
davidemassarenti-optio3: To completely solve the issue, we'd have to move to Javassist to patch the code for CompletableFuture directly. I'd rather not go down that path, quite brittle.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/OutputFactory.java
```diff
@@ -25,7 +25,7 @@ public <T> Output<T> of(T value) {
     }
 
     public <T> Output<T> ofNullable(T value) {
-        var dataFuture = CompletableFuture.supplyAsync(
+        var dataFuture = ContextAwareCompletableFuture.supplyAsync(
```
t0yv0, line 28: Around this change `OutputInternal<T>` is very important to fixup as well as that's highly user-facing. I think once it changes from 

```
    private final CompletableFuture<OutputData<T>> dataFuture;

```

to the new enhanced future we should be good!

t0yv0:
Hi! One of original maintainers here, thanks for picking this up, this is indeed the way to go, we were trying to implement this but run out of time. One key aspect for users is that if users call `x.applyValue(callback)` on outputs, the callback needs to get the right deployment context. I think this approach can accomplish this.

Places to call out as dangerous in documentation is where we expose raw CompletableFuture (un-annotated), probably in results of function invokes.

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/OutputFactory.java
davidemassarenti-optio3: Yes, wrapped in the new class now.

davidemassarenti-optio3:
> Hi! One of original maintainers here, thanks for picking this up, this is indeed the way to go, we were trying to implement this but run out of time. One key aspect for users is that if users call `x.applyValue(callback)` on outputs, the callback needs to get the right deployment context. I think this approach can accomplish this.

applyValue uses the dataFuture field. Now that is a ContextAwareCompletableFuture, which creates more ContextAwareCompletableFuture when chained. 

> Places to call out as dangerous in documentation is where we expose raw CompletableFuture (un-annotated), probably in results of function invokes.

Is there a specific part of the code that I need to look at for that?

t0yv0:
https://github.com/pulumi/pulumi-aws/blob/a8cf81f16ce43402856bfffc3f051e34de7267ec/sdk/java/src/main/java/com/pulumi/aws/iam/IamFunctions.java#L148 for example

davidemassarenti-optio3:
> https://github.com/pulumi/pulumi-aws/blob/a8cf81f16ce43402856bfffc3f051e34de7267ec/sdk/java/src/main/java/com/pulumi/aws/iam/IamFunctions.java#L148 for example

That code pattern ends up in DeploymentImpl, which already wraps the futures in the context-aware class.
Just to be extra safe, I did change ComputableFutures.Builder to start with a context aware object, such that the chaining keeps the context

mikhailshilkov:
@davidemassarenti-optio3  Thank you for adding a description to the PR! I still think we'd love to see it expanded to the next level of details about the implementation, why and how it works, and any caveats. I feel like this is one of those PRs where we need to over-communicate the ideas and intention and then add some more on top...

In terms of testing, I tried a few async things today and they all worked fine for me ([code here](https://gist.github.com/mikhailshilkov/cb203b1e076cf7e1147f2ca0e89b724c)):

- Invoke
- Plain invoke
- CompletableFuture.supplyAsync
- An HTTP request
- Create a resource inside apply

I assume we want to include a cleaned-up version of this into the PR as a test.

I suspect we also want to add a suite of unit- or rather lower-level-integration tests too to exercise scenarios and primitives of the ContextAwareCompletableFuture and their interactions with outputs, resources, deployments, etc. Basically, get our code coverage numbers and confidence up. Also, to help us make changes in this area in the future and feel at least somewhat confident.

Thank you again!

davidemassarenti-optio3:
> @davidemassarenti-optio3 Thank you for adding a description to the PR! I still think we'd love to see it expanded to the next level of details about the implementation, why and how it works, and any caveats. I feel like this is one of those PRs where we need to over-communicate the ideas and intention and then add some more on top...
> 
> In terms of testing, I tried a few async things today and they all worked fine for me ([code here](https://gist.github.com/mikhailshilkov/cb203b1e076cf7e1147f2ca0e89b724c)):
> 
> * Invoke
> * Plain invoke
> * CompletableFuture.supplyAsync
> * An HTTP request
> * Create a resource inside apply
> 
> I assume we want to include a cleaned-up version of this into the PR as a test.
> 
> I suspect we also want to add a suite of unit- or rather lower-level-integration tests too to exercise scenarios and primitives of the ContextAwareCompletableFuture and their interactions with outputs, resources, deployments, etc. Basically, get our code coverage numbers and confidence up. Also, to help us make changes in this area in the future and feel at least somewhat confident.
> 
> Thank you again!

@mikhailshilkov , you mean adding extra information in the commit / code, not in the PR, right?

mikhailshilkov:
> you mean adding extra information in the commit / code, not in the PR, right?

Both a PR description and code/tests. Sorry for mixing it up.

mikhailshilkov:
@t0yv0 @justinvp I’d appreciate another detailed look at this PR, thank you

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
```diff
@@ -0,0 +1,704 @@
...
+        return () ->
+        {
+            DeploymentInstanceHolder.setInstance(context);
+            try
+            {
+                return callback.get();
+            }
+            finally
+            {
+                DeploymentInstanceHolder.setInstance(null);
```
t0yv0, line 99: Should this revert to null or should it revert to previously located context (another DeploymentInstanceHolder.getInstanceNoThrow). Might not matter in practice.

davidemassarenti-optio3: I chose to reset to null everywhere possible (unfortunately, we have some code paths that don't return to us) to ensure we catch a scenario where we leak context to the wrong thread.

t0yv0:
LGTM 🚢

t0yv0 (approved):

sdk/java/pulumi/src/main/java/com/pulumi/core/internal/ContextAwareCompletableFuture.java
t0yv0: Yeah this makes sense.

sdk/java/pulumi/src/main/java/com/pulumi/deployment/DeploymentInstance.java
```diff
@@ -7,4 +7,8 @@
 public interface DeploymentInstance extends Deployment {
     @InternalUse
     DeploymentImpl.Config getConfig();
+
+    boolean isInvalid();
+
+    void markInvalid();
```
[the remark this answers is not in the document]
davidemassarenti-optio3: Not just testing, it's a protective move, to reduce chances that bad code patterns we can't control on the customer's code would lead to a misuse of a deployment, especially in the context of concurrent execution of independent programs.

I've added a comment to describe the purpose

Issue #1675, opened 2025-02-18
High-level SDK to author Component providers in Java


mikhailshilkov:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The existing Java provider SDK enables users to author component packages but it requires some boilerplate to wire everything with GetSchema and Construct implementations.

We want to provide a higher-level opinionated library that would allow creating a component host with one or two lines with sensible defaults. It should:

- Accept a package object (the calling package by default)
- Specify the project name and version (default to package name by default)
- Implement GetSchema to serve the schema automatically inferred from the discovered component resources
- Implement Construct to instantiate the right component resource based on its type, (de-)serialize args and outputs properly

### Affected area/feature

Component packages

PR #1686, opened 2025-02-19
Implement a high-level component provider host

mikhailshilkov:
Similar to other language SDKs, this PR implements a high-level component provider host that discovers schemas and constructors of all components in a given package at runtime.

Instantiating a component provider is now as simple as authoring a component and then calling

```java
public class App {
    public static void main(String[] args) throws IOException, InterruptedException {
        new ComponentProviderHost(new Metadata("javap"), App.class.getPackage()).start(args);
    }
}
```

(see the modified test provider-maven, provider-gradle still tests a manual provider implementation)

I moved the `Metadata` type from the `infer` module to the parent `provider` module.

Resolves https://github.com/pulumi/pulumi-java/issues/1675


tgummerer (approved):
Neat!

PR #1704, opened 2025-02-28
Split out fast vs. slow tests

### LINKED ISSUE #1640: auto: Split out fast vs. slow tests
Since the Automation API tests can take longer than typical unit tests (to run the full lifecycle of deployments), the other languages have split up unit tests into _fast_ and _slow_ tests and have made it possible to easily just run the fast tests. We should consider the same for Java, as right now, `gradle build` will run the tests, and they're starting to take a long time with the Automation API tests included.

justinvp:
Automation API tests can take longer than typical unit tests (because often they are running deployments, or running CLI commands). Because of this, our other language SDKs split out the unit tests into separate fast and slow automation tests. This change does the same for Java.

- To build the project and **run only the fast unit tests** (as usual), run `gradle build` or `make build`.
- To build the project and **run only the slower automation tests**, run `gradle build autoTest` or `make test_auto`.
- To build the project and **run all tests**, both fast and slow, run `gradle build testAll` or `make test_all`.

Fixes #1640

### DIFF
--- a/sdk/java/Makefile
+++ b/sdk/java/Makefile
@@ -1,3 +1,5 @@
+.PHONY: build ensure test_all test_auto install
+
 # Builds the project and runs fast unit tests.
 build::		ensure
 	gradle --console=plain build


sdk/java/Makefile
```diff
@@ -1,11 +1,20 @@
...
 build::		ensure
 	gradle --console=plain build
 
 ensure::
 	java --version
 	gradle --version
 
+# Builds the project and runs all unit tests, both fast and auto tests.
+# A coverage report is generated at pulumi/build/reports/jacoco/allTests
+test_all::		ensure
```
tgummerer, line 11: Should all these targets be marked `.PHONY`?  It probably doesn't matter now, but if someone accidentally has a weird file in their repo for whatever reason they might get confused. (This applies to the existing targets as well, but I just noticed it now)

tgummerer (approved):

sdk/java/Makefile
justinvp: Yep. Done.

PR #1714, opened 2025-03-03
respect base package for java program gen

tgummerer:
Java program gen currently doesn't respect the PackageInfo fields in program gen (but does in SDK gen).  Fix that, so we respect these fields in program gen as well.

### DIFF
--- a/pkg/codegen/java/gen_program.go
+++ b/pkg/codegen/java/gen_program.go
@@ -515,9 +515,14 @@ func sanitizeImport(name string) string {
 	return replacedSlash
 }
 
+// Checks if this is a builtin stack reference or stack reference args
+func isBuiltin(pkg, module, member string) bool {
+	return pkg == "pulumi" && module == "pulumi" && (member == "StackReference" || member == "StackReferenceArgs")
+}
+
 func pulumiImport(pkg string, module string, member string) string {
 	module = cleanModule(module)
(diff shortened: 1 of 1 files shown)


pkg/codegen/java/gen_program.go
```diff
@@ -528,6 +528,29 @@ func pulumiImport(pkg string, module string, member string) string {
 	return "com.pulumi." + sanitizeImport(pkg) + "." + sanitizeImport(module) + "." + member
 }
 
+func pulumiResourceImport(r *pcl.Resource, pkg string, module string, member string) string {
+	module = cleanModule(module)
+	if pkg == "pulumi" && module == "pulumi" && (member == "StackReference" || member == "StackReferenceArgs") {
+		return "com.pulumi.resources." + member
+	}
```
lunaris, line 535: I wonder if this is worth hoisting out to something like `ifIsBuiltin` or the like, since we use it above as well.

pkg/codegen/java/gen_program.go
```diff
@@ -528,6 +528,29 @@ func pulumiImport(pkg string, module string, member string) string {
 	return "com.pulumi." + sanitizeImport(pkg) + "." + sanitizeImport(module) + "." + member
 }
 
+func pulumiResourceImport(r *pcl.Resource, pkg string, module string, member string) string {
```
lunaris, line 531: Worth a comment to mention how this is different/what its intended use is?

tgummerer: Done, thanks!

PR #1717, opened 2025-03-03
Add Mise to `pulumi-java`

lunaris:
This change adds a `.mise.toml` to the repository so that [Mise](https://mise.jdx.dev) can be used to manage developer tooling for `pulumi-java`. This is inline with other PRs across our estate, such as pulumi/pulumi#18626. Developers with Mise installed can now run `mise trust` in `pulumi-java` and have the relevant tooling installed for running `make`, etc.

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,20 @@
+[env]
+
+[tools]
+github-cli = "latest"
+"go:github.com/go-delve/delve/cmd/dlv" = "latest"
+"go:golang.org/x/tools/gopls" = "latest"
+go = "1.24"
+gofumpt = "latest"
+golangci-lint = "1.64.2"
+jq = "latest"
+node = "20"
+"npm:pnpm" = "latest"
+"npm:typescript" = "latest"
+"npm:typescript-language-server" = "latest"
+"npm:yarn" = "latest"
+python = "3.11"
+java = "adoptopenjdk-11.0.26+4"
+gradle = "8.10.2"
+
+[plugins]


.mise.toml
```diff
@@ -0,0 +1,20 @@
...
+go = "1.24"
+gofumpt = "latest"
+golangci-lint = "1.64.2"
+jq = "latest"
+node = "20"
+"npm:pnpm" = "latest"
+"npm:typescript" = "latest"
+"npm:typescript-language-server" = "latest"
+"npm:yarn" = "latest"
+python = "3.11"
```
tgummerer, line 16: Is this required to build Java?

tgummerer (approved):

.mise.toml
lunaris: I've been a bit cautious since the `pulumi` submodule means that in theory you may be dipping in and out of the "full monorepo" at points and it's (IMO) handy to have a superset of that tooling available. But I'm happy if we think this is a bad call.

i-am-tom:
I got the following error on this branch:

```
$ mise install --verbose
DEBUG ARGS: /opt/homebrew/bin/mise install --verbose
DEBUG config: ~/Git/pulumi/java/.mise.toml
DEBUG config: ~/.config/mise/config.toml
DEBUG installing 1 leaf tools first
DEBUG install_some_versions: java@openjdk-11
INFO  java@openjdk-11                     install
Error:
   0: failed to install core:java@openjdk-11
   1: no metadata found for version openjdk-11

Location:
   src/plugins/core/java.rs:254

Version:
   2025.2.6 macos-arm64 (2025-02-16)

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

Do I need to do some extra setup?

.mise.toml
tgummerer: Shoudl we add `dotnet` too in that case?  I'm fine either way to be clear, happy to get this in eithre way.

blampe (approved):
We should turn on Renovate's mise support if we haven't already. It can handle bumping these tools for us.

iwahbe:
We have added mise to this repo.

PR #1719, opened 2025-03-03
java: import packages from com.<namespace>...

tgummerer:
Currently all packages are namespaced in the `com.pulumi` namespace, unless explicitly overridden in the package info.  Use the new namespace field to import packages from `com.<namespace>`.

Depends on https://github.com/pulumi/pulumi-java/pull/1714


tests/integration/java_integration_test.go
```diff
@@ -225,3 +226,19 @@ func stackTransformationValidator() func(t *testing.T, stack integration.Runtime
...
+
+func TestPackageAddWithNamespaceSetJava(t *testing.T) {
+	t.Parallel()
+
+	e := ptesting.NewEnvironment(t)
+	defer e.DeleteIfNotFailed()
+
+	e.ImportDirectory("packageadd-namespace")
+	e.CWD = filepath.Join(e.RootPath, "java")
+	e.RunCommand("pulumi", "package", "add", "../provider/schema.json")
```
brandonpollack23, line 238: I may be just blind but I can't find this file and dont understand how this test is getting set up.  In the other repos its usually in the testdata dir but I'm a bit lost.  I know it is unrelated but could you add a small readme to make this easier in the future ?

tgummerer: The testdata is in `tests/integration/packageadd-namespace`.  We should probably move all that to `tests/integration/testdata`, but just doing it for that one test would just add more inconsistencies, so I avoided it for now.

tests/integration/java_integration_test.go
```diff
@@ -225,3 +226,19 @@ func stackTransformationValidator() func(t *testing.T, stack integration.Runtime
...
+
+	// Make sure the SDK was generated in the expected directory
+	_, err := os.Stat(filepath.Join(e.CWD, "sdks", "example-mypkg", "src", "main", "java",
+		"com", "example", "mypkg", "Utilities.java"))
+	require.NoError(t, err)
```
brandonpollack23, line 243: we ought to check the file itself declares `package com.example` right?

brandonpollack23 (changes requested):

tests/integration/java_integration_test.go
tgummerer: This is covered in a conformance test now, which captures it better than an integration test can.  The integration test is mainly useful to see that `package add` behaves correctly.

PR #1726, opened 2025-03-10
Add a test with a component with async context switching

mikhailshilkov:
Adds a follow-up test to https://github.com/pulumi/pulumi-java/pull/1678 that has a component with multiple styles of async operations

### DIFF
--- a/tests/integration/provider-maven/src/main/java/com/pulumi/example/provider/SampleAsync.java
+++ b/tests/integration/provider-maven/src/main/java/com/pulumi/example/provider/SampleAsync.java
@@ -1,0 +1,104 @@
+package com.pulumi.example.provider;
+
+import com.pulumi.core.annotations.Import;
+import com.pulumi.core.annotations.Export;
+import com.pulumi.core.Output;
+import com.pulumi.random.RandomString;
+import com.pulumi.random.RandomStringArgs;
+import com.pulumi.resources.ComponentResource;
+import com.pulumi.resources.ComponentResourceOptions;
+import com.pulumi.resources.CustomResourceOptions;
+import com.pulumi.resources.ResourceArgs;
+import com.pulumi.std.inputs.AbsArgs;
+import com.pulumi.std.StdFunctions;
+import com.pulumi.std.inputs.AbsPlainArgs;
+import java.util.concurrent.CompletableFuture;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+import java.net.URI;
+
+class SampleAsyncArgs extends ResourceArgs {
+    @Import(name="length", required=true)
+    private Output<Integer> length;
+
+    public Output<Integer> length() {
+        return this.length;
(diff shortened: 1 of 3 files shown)


tests/integration/provider-maven/src/main/java/com/pulumi/example/provider/SampleAsync.java
```diff
@@ -0,0 +1,104 @@
...
+                    
+                    HttpResponse<String> response = client.send(request, 
+                        HttpResponse.BodyHandlers.ofString());
+                    return randomStr + " - Todo #" + todoId + ": " + response.body();
+                } catch (Exception e) {
+                    throw new RuntimeException("HTTP request failed", e);
+                }
+            });
+    }
+}
```
tgummerer, line 104: Missing newline here.

tests/integration/provider-maven/src/main/java/com/pulumi/example/provider/SampleAsync.java
```diff
@@ -0,0 +1,104 @@
...
+                var randomString2 = new RandomString(name + "2", 
+                    RandomStringArgs.builder()
+                        .length(totalLength)
+                        .special(false)
+                        .build(), resOpts);
+                
+                try {
+                    HttpClient client = HttpClient.newHttpClient();
+                    HttpRequest request = HttpRequest.newBuilder()
+                        .uri(URI.create("https://jsonplaceholder.typicode.com/todos/" + todoId))
```
tgummerer, line 93: Does Java have something like Go's test HTTP server?  Or is there another way to simulate this asyncronicity? It's not ideal to make an actual HTTP request in our tests, especially to a site we don't control.

mikhailshilkov: I could do a Java http server but I don't think I have the right lifecycle to start/stop it...

```
HttpServer server = HttpServer.create(new InetSocketAddress(8000), 0);
server.createContext("/test", exchange -> {
    String response = "{\"status\":\"ok\"}";
    exchange.sendResponseHeaders(200, response.length());
    try (var os = exchange.getResponseBody()) {
        os.write(response.getBytes());
    }
});
server.start();
```

We could do something fake like

```
URI uri = URI.create("data:text/plain,Hello%20World");
HttpRequest request = HttpRequest.newBuilder().uri(uri).build();
```

but I'm not sure it catches all the right problems?

The TODO thing is specifically designed for this case: "Free fake and reliable API for testing and prototyping". I think we are relatively safe and get a real test with some networking and latency, which may affect the asynchronous behavior?

tgummerer: >  I think we are relatively safe and get a real test with some networking and latency

Every network call in the tests are a potential source for flakes/CI breakages, so I'd really rather not introduce the dependency. Even the most stable services tend to break sometimes, in which case CI is gonna be red.

PR #1730, opened 2025-03-12
Add default `ci-ok` checks for `CHANGELOG`-only PRs

lunaris:
When a pull request (PR) only touches `CHANGELOG*` files, we skip CI checks in order to make processes that only modify these files, such as releases, much faster. Recently, we tightened controls on this repository so that all PRs require a passing `ci-ok` step in order to merge. As a result of this, we need to tweak our skipping approach to produce a "no-op" `ci-ok` step so that `CHANGELOG*`-only PRs can merge. This change does that by copying the GitHub-sanctioned pattern used in `pulumi/pulumi`. This will be tested by the upcoming 1.6.0 release PR.

### DIFF
--- a/.github/workflows/pr-ci-ok-only.yml
+++ b/.github/workflows/pr-ci-ok-only.yml
@@ -1,0 +1,40 @@
+# This workflow is a version of `pr.yml` that runs in exactly the cases that `pr.yml` doesn't. That is, where `pr.yml`
+# skips execution due to `paths-ignore`, this workflow runs only in those cases. We do this so that we can always
+# produce a `ci-ok` check, which we configure as a required passing step on all PRs to be merged into the repository.
+#
+# This is an officially blessed pattern for handling skipped but required status checks. See
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+# for more information.
+
+on:
+  pull_request:
+    paths-ignore:
+      # Skip all files *except CHANGELOG* changes.
+      - '**'
+      - '!CHANGELOG.md'
+      - '!CHANGELOG_PENDING.md'
+  push:
+    branches:
+      - main
+    tags:
+      - sdk/v*.*.*
+    paths-ignore:
+      # Skip all files *except CHANGELOG* changes.
+      - '**'
+      - '!CHANGELOG.md'
+      - '!CHANGELOG_PENDING.md'
+
+jobs:
+  no-op:
+    name: Skip CI on CHANGELOG changes
+    runs-on: ubuntu-latest
+    steps:
+      - name: Skip CI on CHANGELOG changes
+        run: echo 'No need to run CI checks when only CHANGELOGs change'
+
+  ci-ok:
+    name: ci-ok
+    runs-on: ubuntu-latest
+    steps:
+      - name: CI succeeded
+        run: exit 0
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,6 +1,8 @@
 on:
   pull_request:
     paths-ignore:
+      # Skip CI checks if the only things that have changed are CHANGELOG entries (which is typically the case when a PR
+      # or Push is occurring as part of a release process).
       - 'CHANGELOG.md'
       - 'CHANGELOG_PENDING.md'
   push:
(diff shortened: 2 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -9,14 +11,14 @@ on:
...
+      # Skip CI checks if the only things that have changed are CHANGELOG entries (which is typically the case when a PR
+      # or Push is occurring as part of a release process).
       - 'CHANGELOG.md'
       - 'CHANGELOG_PENDING.md'
   repository_dispatch:
     types:
       - run-acceptance-tests-command
 
-# TODO[pulumi/java#10] protect jobs from crypto mining PR attacks
-# before making the repo public.
```
lunaris, line 19: This issue is long closed so this comment is not relevant; I took the opportunity to clean it up as part of this change -- hope this is OK!

.github/workflows/pr-ci-ok-only.yml
```diff
@@ -0,0 +1,43 @@
+# This workflow is a version of `on-pr.yml` that runs in exactly the cases that `on-pr.yml` doesn't. That is, where
+# `on-pr.yml` skips execution due to `paths-ignore`, this workflow runs only in those cases. We do this so that we can
+# always produce a `ci-ok` check, which we configure as a required passing step on all PRs to be merged into the
+# repository.
```
tgummerer, line 4:
```suggestion
# This workflow is a version of `pr.yml` that runs in exactly the cases that `pr.yml` doesn't. That is, where
# `pr.yml` skips execution due to `paths-ignore`, this workflow runs only in those cases. We do this so that we can
# always produce a `ci-ok` check, which we configure as a required passing step on all PRs to be merged into the
# repository.
```

.github/workflows/pr-ci-ok-only.yml
```diff
@@ -0,0 +1,43 @@
...
+      - main
+    tags:
+      - sdk/v*.*.*
+    paths-ignore:
+      # Skip all files *except CHANGELOG* changes.
+      - 'CHANGELOG.md'
+      - 'CHANGELOG_PENDING.md'
+  repository_dispatch:
+    types:
+      - run-acceptance-tests-command
```
tgummerer, line 28: I don't think we need this workflow in these cases.  It doesn't matter for pushes to `main`, but I think the `repository_dispatch` bit is wrong, this shouldn't run when `/run-acceptance-tests` is run I think (do we even have that in this repo?)

.github/workflows/pr.yml
tgummerer: Hah, I stumbled over thsi so many times, and always forget to do something about it.  Thanks for removing it.  We're definitely protected from this by existing CI infrastructure already.

.github/workflows/pr-ci-ok-only.yml
```diff
@@ -0,0 +1,43 @@
...
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+# for more information.
+
+on:
+  pull_request:
+    paths-ignore:
+      # Skip all files *except CHANGELOG* changes.
+      - '**'
+      - 'CHANGELOG.md'
+      - 'CHANGELOG_PENDING.md'
```
tgummerer, line 16:
```suggestion
      - '!CHANGELOG.md'
      - '!CHANGELOG_PENDING.md'
```

These need to be excluded from the pattern.

.github/workflows/pr-ci-ok-only.yml
lunaris: Happy to give it a shot -- removed.

.github/workflows/pr-ci-ok-only.yml
```diff
@@ -0,0 +1,39 @@
...
+      - '!CHANGELOG_PENDING.md'
+  push:
+    branches:
+      - main
+    tags:
+      - sdk/v*.*.*
+    paths-ignore:
+      # Skip all files *except CHANGELOG* changes.
+      - '!CHANGELOG.md'
+      - '!CHANGELOG_PENDING.md'
```
tgummerer, line 24:
```suggestion
      - '**'
      - '!CHANGELOG.md'
      - '!CHANGELOG_PENDING.md'
```

If we want this to run for pushes to `main` we first need to include `**` before excluding the other files.

.github/workflows/pr.yml
```diff
@@ -475,4 +477,4 @@ jobs:
         if: ${{ needs.go-tests.result != 'success' || needs.examples.result != 'success' || needs.go-lint.result != 'success' || needs.test_integrations.result != 'success' || needs.test_templates.result != 'success' }}
         run: exit 1
       - name: CI succeeded
-        run: exit 0
+        run: exit 0
```
tgummerer, line 480: unintended change?

tgummerer (approved):

.github/workflows/pr.yml
lunaris: Yep reverted

PR #1744, opened 2025-03-26
automatically infer namespaces

tgummerer:
Allow Java components to automatically infer the namespace based on the package name of the component.


sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/infer/ComponentAnalyzerTests.java
```diff
@@ -32,7 +32,8 @@ void testGenerateSchemaWithoutMetadata() {
         var expected = new PackageSpec()
             .setName("infer")
             .setDisplayName("infer")
-            .setVersion(null);
+            .setVersion(null)
+            .setNamespace("pulumi");
```
tgummerer, line 36: Before this PR there would not have been any namespace set, so this will test it.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -68,25 +68,33 @@ public static PackageSpec generateSchema(Metadata metadata, Class<?>... classes)
 
         ComponentAnalyzer analyzer = new ComponentAnalyzer(metadata);
         Map<String, ResourceSpec> components = new HashMap<>();
-        
+
+        if (classes.length == 0) {
+            throw new IllegalArgumentException("At least one component class must be provided");
+        }
+
+        String namespace = classes[0].getPackage().getName().split("\\.")[1];
```
tgummerer, line 76: Not sure if we should error out if not all classes have the same namespace?

mikhailshilkov: Probably a safer way to go

mikhailshilkov, line 76: Can there be no `.` at all?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/schema/PackageSpec.java
```diff
@@ -58,6 +66,10 @@ public String getName() {
         return name;
     }
 
+    public String getNamespace() {
+	return namespace;
```
mikhailshilkov, line 70: Thank you for cleaning up all whitespaces. We probably need a linter. But - nit: this line doesn't look right?

mikhailshilkov: In another view it's fine but here it's shown as
image

tgummerer: Whoops, that's tabs vs.spaces, sorry about that, should be fixed now.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
tgummerer: There can be. I think in that case we just leave the namespace empty.  I made it so now.

tgummerer: Done, added an exception.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -50,10 +50,19 @@ public static PackageSpec generateSchema(Class<?>... classes) {
...
+
         return generateSchema(new Metadata(name, null, null), classes);
     }
 
+    private static String getNamespace(Class<?> clazz) {
+        String[] split = clazz.getPackage().getName().split("\\.");
+        if (split.length == 0) {
+            return "";
+        }
+        return split[1];
```
mikhailshilkov, line 62: This will fail for length == 1. Unit tests, please.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -68,25 +77,39 @@ public static PackageSpec generateSchema(Metadata metadata, Class<?>... classes)
...
+            if (!getNamespace(clazz).equals(namespace)) {
+                throw new IllegalArgumentException("All classes must be in the same top level package. Expected: " + namespace + " Found: " + getNamespace(clazz));
+            }
             if (ComponentResource.class.isAssignableFrom(clazz) && !clazz.isInterface() && !java.lang.reflect.Modifier.isAbstract(clazz.getModifiers())) {
                 components.put(clazz.getSimpleName(), analyzer.analyzeComponent(clazz));
             }
         }
 
-        return analyzer.generateSchema(metadata, components, analyzer.typeDefinitions);
+        // Java package names are not allowed to contain underscores, but namespaces expect dashes instead.
```
mikhailshilkov, line 96:
```suggestion
        // Java package names are allowed to contain underscores, but namespaces expect dashes instead.
```
?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -68,25 +77,39 @@ public static PackageSpec generateSchema(Metadata metadata, Class<?>... classes)
...
+        namespace = namespace.replaceAll("_", "-");
```
mikhailshilkov, line 97: Move to getNamespace?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
```diff
@@ -68,25 +77,39 @@ public static PackageSpec generateSchema(Metadata metadata, Class<?>... classes)
...
+
+        if (classes.length == 0) {
+            throw new IllegalArgumentException("At least one component class must be provided");
+        }
+
+        String namespace = getNamespace(classes[0]);
+
         for (Class<?> clazz : classes) {
+            if (!getNamespace(clazz).equals(namespace)) {
+                throw new IllegalArgumentException("All classes must be in the same top level package. Expected: " + namespace + " Found: " + getNamespace(clazz));
```
mikhailshilkov, line 89: Test this condition?

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
tgummerer: Done.

sdk/java/pulumi/src/main/java/com/pulumi/provider/internal/infer/ComponentAnalyzer.java
tgummerer: Done.

sdk/java/pulumi/src/test/java/com/MyComponent.java
```diff
@@ -0,0 +1,13 @@
+package com;
```
tgummerer, line 1: Not sure if there's a better way to add these test classes other than checking these in?  At least it's only in `src/test` so won't end up in the real package, so maybe it's okay?

mikhailshilkov (approved):
LGTM

PR #1761, opened 2025-04-08
Get Everything building locally with Mise

i-am-tom:
After a lot of experimenting, this has turned out to be very underwhelming - mostly just picking very specific versions of things and making stack names fully qualified in tests.

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/automation/LocalWorkspaceTest.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/automation/LocalWorkspaceTest.java
@@ -945,8 +945,8 @@ public class LocalWorkspaceTest {
           if (System.getenv("PULUMI_TEST_ORG") != null) {
               return System.getenv("PULUMI_TEST_ORG");
           }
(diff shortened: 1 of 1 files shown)


sdk/java/pulumi/src/test/java/com/pulumi/automation/LocalWorkspaceTest.java
```diff
@@ -933,4 +935,18 @@ void testStateDeleteForce(@EnvVars Map<String, String> envVars) throws Exception
...
+    }
+
+    static String getTestOrg() {
+      if (System.getenv("PULUMI_ACCESS_TOKEN") != null) {
+          if (System.getenv("PULUMI_TEST_ORG") != null) {
+              return System.getenv("PULUMI_TEST_ORG");
+          }
+          return "organization";
+      }
+      return "pulumi-test";
```
tgummerer, line 950: I think these two are the wrong way around, no? If `PULUMI_ACCESS_TOKEN` is set we want to return `pulumi-test` by default, and return `organization` if it's not set at all.

PR #1763, opened 2025-04-10
Add import resource option codegen to java.

brandonpollack23:
Blocked on pulumi/pulumi#19163


brandonpollack23:
Once https://github.com/pulumi/pulumi/pull/19163 is submitted, run conformance tests and reupload

brandonpollack23:
not entirely sure when i run with accept nothing is generated but it does pass

PR #1766, opened 2025-04-16
update pulumi submodle to latest `master`

tgummerer:
To bring in `l2-namespaced-provider`. Unfortunately there's a lot of conformance tests failing, so those are being excluded here. `l2-namespaced-provider` also doesn't work because it relies on components that don't currently work in Java.  The codegene'd code still shows that imports are handled correctly.

Fixes https://github.com/pulumi/pulumi/issues/19150


julienp (approved):
LGTM. I think the test failure is unrelated.

PR #1767, opened 2025-04-16
run CI tests on a regular schedule

tgummerer:
Run the CI tests in this repo on a regular schedule, to guard against bitrot.  "Inspired" by https://github.com/pulumi/pulumi-java/actions/runs/14494159872/job/40657808628?pr=1766 failing, which I don't think has anything to do with my PR, but if we had a cron job running the tests on a regular schedule it would be easier to check.

Also rename the CI job from pr.yml -> ci.yml, since it's not only running on PRs already and even less so now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,9 @@ on:
     types:
       - run-acceptance-tests-command
 
+  schedule:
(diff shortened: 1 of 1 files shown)


julienp (approved):
Nice, good idea!

PR #1780, opened 2025-04-25
Add `--target` and `--target-dependents` to the Automation API

i-am-tom:
Filling in some gaps before I add the `--exclude` and `--exclude-dependents` flags.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/automation/DestroyOptions.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/automation/DestroyOptions.java
@@ -2,11 +2,6 @@
 
 package com.pulumi.automation;
 
-import java.util.Collections;
-import java.util.List;
-
-import javax.annotation.Nullable;
-
 /**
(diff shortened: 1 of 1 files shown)


sdk/java/pulumi/src/main/java/com/pulumi/automation/DestroyOptions.java
```diff
@@ -29,6 +38,15 @@ public static Builder builder() {
         return new Builder();
     }
 
+    /**
+     * The list of specified targets.
+     *
+     * @return The specified targets
+     */
+    public List<String> targets() {
```
justinvp, line 46: This is exposed on the base `UpdateOptions`: 

https://github.com/pulumi/pulumi-java/blob/e8009acfe682371b1d6704efef3d5ea1545e601e/sdk/java/pulumi/src/main/java/com/pulumi/automation/UpdateOptions.java#L87-L95

But I guess I missed passing the CLI flags if it's set?

PR #1781, opened 2025-04-28
Add `--exclude` and `--exclude-dependents` to the Automation API

i-am-tom:
Copying the same structure of `targets` and `targetDependents`.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/automation/UpOptions.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/automation/UpOptions.java
@@ -155,6 +155,7 @@ public final class UpOptions extends UpdateOptions {
         private boolean diff;
         @Nullable
         private List<String> replaces;
+        private boolean excludeDependents;
         private boolean targetDependents;
         @Nullable
         private Consumer<Context> program;
@@ -199,6 +200,18 @@ public final class UpOptions extends UpdateOptions {
          */
         public Builder replaces(List<String> replaces) {
             this.replaces = replaces;
+            return this;
+        }
+
+        /**
+         * Allows exclusion of dependent targets discovered but not specified
+         * {@link #excludes}.
+         *
+         * @param excludeDependents true if dependent targets should be excluded
+         * @return the builder
+         */
+        public Builder excludeDependents(boolean excludeDependents) {
+            this.excludeDependents = excludeDependents;
             return this;
         }
 


sdk/java/pulumi/src/main/java/com/pulumi/automation/UpdateOptions.java
```diff
@@ -269,6 +285,19 @@ public B message(String message) {
...
+         * (*, **) are also supported.
+         *
+         * @param excludes the list of excluded resource URNs
+         * @return the builder
+         */
+        @SuppressWarnings("unchecked")
+        public B excludes(List<String> excludes) {
+            this.excludes = excludes;
+            return (B) this;
+        }
```
lunaris, line 299: Part of me wonders why `--*-dependents` isn't on `UpdateOptions`. I'm wondering if it's because of the bug/miss that you fixed when you joined, which is that `destroy` (or `refresh`?) didn't support `--target-dependents`? And so that was mirrored into the automation API. Given that we've now fixed this, I wonder if we can move `--target-dependents` up to `UpdateOptions`. It might be breaking -- not sure. Either way, not for this PR.

But what _is_ for this PR, is `--exclude-dependents`. It's not symmetric (and thus potentially ugly), but I think it's safe to put it on `UpdateOptions`, no? Might cut down the code a bit too?

lunaris: Related: this PR misses `RefreshOptions`, which omits `--target-dependents` (which we should fix) and thus (I think) confirms my theory

lunaris (approved):
Some thoughts but happy for us to get this in!

sdk/java/pulumi/src/main/java/com/pulumi/automation/UpdateOptions.java
i-am-tom: Yeah, with this PR I was trying to follow the current structure - I agree that it seems safe to move the `--*-dependents` flag onto the `UpdateOptions`, but I'd equally be happy to move both `--*-dependents` flags to the `UpdateOptions` in a follow-up? .NET works in exactly the same way and I made a [near-identical PR](https://github.com/pulumi/pulumi-dotnet/pull/582), so I'm happy to do a follow-up for each one?

lunaris: I think moving both _might_ be a breaking change (not sure), so I was only thinking just "do this one right" but am happy either way.

PR #1852, opened 2025-07-11
Fix Java publishing workflows

lunaris:
OSSRH, the service we have been using to publish and release Java artifacts, reached end-of-life on June 30th 2025. As a result, we can no longer use it. This commit migrates us to the recommended new URL for publishing under Maven Central, which will hopefully work in exactly the same way.


i-am-tom (approved):
Love software

Issue #1359, opened 2024-05-10
Sonatype migrating from OSSHR to a new Maven Central


ringods:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

This issue is to create awareness around the effort Sonatype is doing to migrate the main Java package repository to a new backend.

* [OSSHR (Legacy)](https://central.sonatype.org/register/legacy/) - generated `pom.xml` files uses this way of publishing
* [Central Portal (New)](https://central.sonatype.org/register/central-portal/)

There is no timeline yet when the legacy repository will be phased out.

### Affected area/feature

Publishing of Java SDKs.

ringods:
Is it possible to add support for the new way as an addition to the generated `pom.xml` files? With Pulumiverse, we already opted to the new Central Portal. Without the support in the `pom.xml` files, we can't publish Java SDKs.

t0yv0:
Do you have a fully working example, this would be appreciated. I am currently looking into the internal publishing of Pulumi Java providers which stopped working, it utilizes Gradle however.

mjeffryes:
Sonatype has announced that they are shutting down OSSRH by June 30, 2025

t0yv0:
We're affected by this.

t0yv0:
Because of https://github.com/pulumi/pulumi-package-publisher/blob/c1672c7928591d563dccb12729e05e315c21f8c2/lang/java/action.yml#L26 the provider releases fail silently when they try to package Java SDKs.

t0yv0:
There are three ways to accommodate Java publishing of Pulumi packages to use Maven Central:

1. Use Maven, instead of Gradle, and use official Maven plugins. This is the recommended option from Maven Central
2. Use Gradle with community plugins, such as [JReleaser](https://jreleaser.org) being the most popular option recommended by Maven Central
3. Use Gradle with io.github.gradle-nexus.publish-plugin plugin and [OSSRH-style Staging API](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/); this is the option Pulumi is going for internally for all Pulumi-maintained providers as it provides the lowest lift and continuity with how the CI has been setup thus far

At this point I have the most context on option 3 and here is how it looks like.

[...]

PR #1854, opened 2025-07-11
Move Java publishing defaults from OSSRH to Maven Central

t0yv0:
Updates from defaulting to OSSRH to defaulting to Maven Central

Java Publish code is updated to default to pushing to Maven Central.

PUBLISH_REPO_URL env var now needs full URL such as:

    https://central.sonatype.com/repository/maven-snapshots/

This is a change from accepting shorter URL such as:

    https://central.sonatype.com/

A new PUBLISH_STAGING_URL env var is added for customizing the URL of the OSSRH-style staging API used by the
gradle-nexus plugin:

    https://ossrh-staging-api.central.sonatype.com/service/local/

Relates to https://github.com/pulumi/pulumi-java/issues/1359


t0yv0:
Before pushing this forward I will attempt to build and release an alpha version of a provider and see how that goes.

t0yv0:
I've managed an end-to-end release on this code:

https://central.sonatype.com/artifact/com.pulumi/databricks/v1.74.0-alpha.1

I think we can roll this out. PTAL.

PR #1912, opened 2025-11-03
Fix package add test

julienp:
This repo’s integration tests are setup in a somewhat opinionated manner. To run in CI, tests must be a subtest of `TestIntegrations` and listed in the test matrix `https://github.com/pulumi/pulumi-java/blob/6f7c39035e5a42ce799508c1186705b114f83e3b/.github/workflows/ci.yml#L329`

Addtionally, the package add test was not setup correcrtly. The schema specifies

```
  "name": "mypkg",
  "namespace": "example",
```

so the expected path is `com/mynamespace/mypkg`.


.github/workflows/ci.yml
```diff
@@ -337,6 +337,8 @@ jobs:
           - retype-parents
           - provider-maven
           - provider-gradle
+          - package-add
+          - convert
```
julienp: This test is skipped with `t.Skip`, but let's include it here so we don't forget about it.

I had a quick look at unskipping the convert test, but it's failing because it needs an azure subscription ID. Maybe we can change this to use `random` instead, to make this test easily runnable locally and in CI.

We probably want to revise this whole testing setup one day.

tgummerer (approved):

.github/workflows/ci.yml
julienp: Yeah we should. I didn't want to perturb too many things right now. I want to do a change touching package add, so just making sure things are tested right now.

Maybe something for the ops rotation.

PR #1913, opened 2025-11-04
Remove the integration matrix from CI

i-am-tom:
This PR removes the test matrix, and just runs all the tests with maximum concurrency.


julienp (approved):
Thanks!

PR #1928, opened 2025-11-20
Manage changelogs with `changie`

i-am-tom:
Java's deploy process is just a little bit more effort than dotnet and yaml. We have to merge the PENDING entries into CHANGELOG, which triggers a job that uses the content of PENDING, and then push some tags and finally remove PENDING. With `changie`, we can always see what would have been the contents of PENDING using the `latest` command, so we can remove the final step entirely: do a `changie` PR as we would in dotnet or java, and then push the tags and we're done.

In a follow-up, I'd like to take out the second step there, but this one was already looking scary. Really, this change should be +44/-7 but we have to migrate all previous entries and reformat the changelog, so it looks much scarier than it is.

### DIFF
--- a/.github/workflows/release-java-provider.yml
+++ b/.github/workflows/release-java-provider.yml
@@ -46,7 +46,18 @@ jobs:
               ${{ steps.go-cache-paths.outputs.go-build }}
               ${{ steps.go-cache-paths.outputs.go-mod }}
           key: go-cache-${{ hashFiles('**/go.sum') }}
+      - name: Check if release exists
+        # This lets us update CHANGELOG.md and with updated release notes without retriggering a release
+        # that's already happened.
+        id: check-release
+        run: |
+          if [ -z "$(git tag -l ${{ env.GORELEASER_CURRENT_TAG }})" ]; then
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+          else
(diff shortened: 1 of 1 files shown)


i-am-tom:
Check currently fails because of an aqua registry break, should be fixed in next release https://github.com/jdx/mise/pull/7018

.changie.yaml
```diff
@@ -0,0 +1,33 @@
...
+fragmentFileFormat: '{{.Kind}}-{{.Custom.PR}}'
+components:
+ - sdk
+ - runtime
+ - codegen
+ - language
+changeFormat: '- [{{.Component}}] {{.Body}} [#{{.Custom.PR}}](https://github.com/pulumi/pulumi-java/pull/{{.Custom.PR}})'
+custom:
+  - key: PR
+    label: GitHub Pull Request
```
Frassle:
```suggestion
    label: GitHub Pull Request or Issue Number
```

.github/workflows/release-java-provider.yml
```diff
@@ -39,4 +50,4 @@ jobs:
         uses: goreleaser/goreleaser-action@v6
```
Frassle: Shouldnt this copy the check-release step from https://github.com/pulumi/pulumi-dotnet/blob/main/.github/workflows/release-dotnet-provider.yml as well?

PR #1959, opened 2025-12-12
Update goldenfiles

i-am-tom:
I ran `PULUMI_ACCEPT=1 make test_go` and noticed we didn't have the goldenfiles for `l2-resource-names` committed.


iwahbe:
@i-am-tom is it possible that you were not at the latest version somehow? CI should enforce that the golden files are up-to-date.

i-am-tom:
@iwahbe https://github.com/pulumi/pulumi/pull/21216/changes?w=1 (this is what I did to make this fail on `main`)

PR #1973, opened 2026-01-08
Update pulumi submodule to 3.215.0

julienp:
By using a tag in .gitmodules we enable renovate to update this for us when new versions are released:
https://docs.renovatebot.com/modules/manager/git-submodules/


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "pulumi"]
 	path = pulumi
 	url = https://github.com/pulumi/pulumi
-    branch = master
\ No newline at end of file
+    branch = v3.215.0
```
tgummerer: Minor nit, but we should have a newline here at the end of the file.

julienp: fixed

PR #2032, opened 2026-02-25
go mod tidy

julienp:
There are some extra deps in go.mod, tidy them. I think this originates from some rebasing I did in https://github.com/pulumi/pulumi-java/pull/2010, or maybe renovate being weird.


julienp:
huh, these are actually needed by github.com/pulumi/pulumi/pkg/v3/testing/pulumi-test-language?

PR #2129, opened 2026-04-07
Fix inherited input properties missing from Java component schema

wlami:
## Summary

- When an args class inherits from a base class, only fields declared on the immediate class were included in the generated schema - inherited properties were missing
- Root cause: `type.getDeclaredFields()` in `analyzeType` only returns fields declared directly on the class, not inherited ones
- Added `getAllFields()` helper that walks the class hierarchy, stopping at Pulumi SDK base classes (`ResourceArgs`, `ComponentResource`, `CustomResource`)
- Related fixes: pulumi/pulumi-dotnet#930 (C#), pulumi/pulumi#22446 (TypeScript), pulumi/pulumi#22484 (Python)

**Note:** Outputs still use `getDeclaredFields` (direct only). Switching outputs to `getAllFields` would require similar base-class filtering - left as a separate fix.

## Test plan

- [x] Added `testInheritedInputProperties` test with `BaseComponentArgs`/`DerivedComponentArgs` inheritance
- [x] All 18 ComponentAnalyzer tests pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/infer/ComponentAnalyzerTests.java
+++ b/sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/infer/ComponentAnalyzerTests.java
@@ -838,13 +838,23 @@ public class ComponentAnalyzerTests {
     @Test
     void testInheritedInputProperties() {
         var schema = ComponentAnalyzer.generateSchema(metadata, InheritedComponent.class);
-        var resource = schema.getResources().get("my-component:index:InheritedComponent");
 
-        assertNotNull(resource);
-        // Both base and derived input properties should be present
-        assertTrue(resource.getInputProperties().containsKey("baseProp"),
-            "baseProp from base class should be present");
-        assertTrue(resource.getInputProperties().containsKey("childProp"),
-            "childProp from derived class should be present");
+        var expected = createBasePackageSpec()
+            .setResources(Map.of(
+                "my-component:index:InheritedComponent",
+                new ResourceSpec(
+                    Map.of(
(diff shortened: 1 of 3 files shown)


sdk/java/pulumi/src/test/java/com/pulumi/provider/internal/infer/ComponentAnalyzerTests.java
```diff
@@ -813,4 +813,38 @@ void testAnalyzeComponentMutuallyRecursiveComplexTypes() {
...
+    void testInheritedInputProperties() {
+        var schema = ComponentAnalyzer.generateSchema(metadata, InheritedComponent.class);
+        var resource = schema.getResources().get("my-component:index:InheritedComponent");
+
+        assertNotNull(resource);
+        // Both base and derived input properties should be present
+        assertTrue(resource.getInputProperties().containsKey("baseProp"),
+            "baseProp from base class should be present");
+        assertTrue(resource.getInputProperties().containsKey("childProp"),
+            "childProp from derived class should be present");
```
iwahbe, line 848: Instead of asserting only that we contain the props we want, let's assert on the whole list containing exactly the properties we want. 

Right now, this test doesn't verify that we exclude properties from `ResourceArgs`, `ComponentResource` or `CustomResource` for example. We need to assert that those fields are not inadvertently added here.

iwahbe (changes requested):
Thanks for the PR and sorry for the long review time. I'm not a Java expert, but I think the change looks good. The test needs a small change before merging.

iwahbe (approved):

PR #2156, opened 2026-05-02
[Fix] Allow child resources to override parent's protect setting

addu390:
## Summary

Allow a child resource to override its parent's `protect` setting in the Java SDK.

This PR:
- Changes the field to `@Nullable Boolean` so explicit `false` is distinguishable.
- Adds a `protect(@Nullable Boolean)` builder overload, existing `.protect(true)` / `.protect(false)` calls work as-is.
- Child wins when explicitly set, otherwise inherit from parent.

Fixes the Java side of pulumi/pulumi#18935. Matches the Go fix in pulumi/pulumi#18862 pulumi/pulumi#18924.

## Questions

1. Should this be coordinated with the pending Go fix in #18935?
2. Similar bug exists for `retainOnDelete`. Good to fold it into this one?


addu390:
@Frassle Your take on how to best go about releasing this in the Java SDK?

sdk/java/pulumi/src/main/java/com/pulumi/resources/ResourceOptions.java
```diff
@@ -370,7 +382,7 @@ public Output<List<Resource>> getDependsOn() {
      * @see Builder#protect(boolean)
      */
     public boolean isProtect() {
```
Frassle, line 392: This is what's actually used to set the RegisterResource request, so I think we need another method here "isProtectNullable"(?) to check in the register resource code to work out if we should actually send true, false, or null to the engine.

Also it would be good to get the l2-resource-parent-inheritance test passing with this change, even if it requires a manually written program file.

addu390: @Frassle Went with `Optional<Boolean> getProtect()` instead, similar to the other `Optional<>` getters.

And for the test: `retainOnDelete` needed the same tri-state fix to make it pass, and while in there, `deleteBeforeReplace` had the same bug. Saw your draft in #1956 was heading the same way, so applied the pattern across all three.

Tests passed locally.

addu390:
Hi @Frassle, no rush, but could you review when you get a chance. Was hoping to get the fix deployed for Java users.

Frassle (approved):
Looks reasonable to me, just get a changelog added and we can merge

addu390:
> Looks reasonable to me, just get a changelog added and we can merge

Thank you. Added the change-log.

PR #2190, opened 2026-05-18
Fix l1-config-types-object language test

### LINKED ISSUE #2005: Implement `l1-config-types-object`

iwahbe:
## Summary

Implements `l1-config-types-object` by teaching the Java program codegen to handle map, object, and dynamic configuration types.

- **`object(...)` configs**: synthesize a nested static class `<Name>Config` inside `App` with a private field and method accessor per property. The retrieval call becomes `config.requireObject(\"<key>\", <Name>Config.class)`, which Gson deserializes via reflection.
- **`map(...)` configs**: lower to `config.requireObject(key, TypeShape.map(String.class, <Element>.class))`, with attribute and index access generating `.get(key)` for the resulting `java.util.Map`.
- **Dynamic / untyped configs** (`config \"x\" {}`): retrieved as `Map<String, Object>` via the dynamic `TypeShape`. The intrinsic-convert lowering emits a `((Number) ...).doubleValue()` / `((Boolean) ...)` / `((String) ...)` cast so arithmetic and equality on dynamic values compiles. The traversal tracks when it has crossed a dynamic boundary and inserts a `(java.util.List)` cast before any subsequent list indexing.
- **SDK**: adds `Config.requireObject(key, TypeShape)` and `Config.requireSecretObject(key, TypeShape)`, mirroring the existing `Class<T>` overloads.

Fixes #2005

### DIFF
--- a/.changes/unreleased/Improvements-l1-config-types-object.yaml
+++ b/.changes/unreleased/Improvements-l1-config-types-object.yaml
@@ -1,0 +1,6 @@
+kind: Improvements
+body: Correctly generate object types for program config
+time: 2026-05-18T00:00:00Z
+custom:
+  PR: "2189"
+component: codegen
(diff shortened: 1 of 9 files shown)


tgummerer (approved):
I'm trusting the tests more than my ability to understand this code.

Issue #2192, opened 2026-05-19
v1.27.0 failed to publish to Maven Central


jkodroff:
v1.27.0 failed to publish to Maven Central: https://mvnrepository.com/artifact/com.pulumi/pulumi

The docs _did_ get published to the Pulumi site: https://www.pulumi.com/docs/reference/pkg/java/

Looks like @iwahbe attempted a fix: https://github.com/pulumi/pulumi-java/commit/6236f33f2412c2f3e990be8852a491a02b160813

But then this workflow run failed: https://github.com/pulumi/pulumi-java/actions/runs/25802671524/job/75796731002

jkodroff:
What's done is done - 1.28.0 is already out. Closing.

PR #2198, opened 2026-05-19
Consolidate release into a single workflow

### LINKED ISSUE #2195: Release workflow needs to be restructured
A couple of problems have been observed with the release workflow:

1. It's possible for the docs to be updated without the actual package being published to Maven Central (see #2192)
2. The docs publishing workflow may have been broken by a workflow rename (see https://github.com/pulumi/pulumi-java/actions/runs/26110531106)

The release workflow should be a single workflow with the following characteristics:

1. The call to pulumi/docs to generate the SDK docs should not fire until the package has been successfully pushed to Maven Central
2. The call to `pulumi/docs` should be correct and working

jkodroff:
## Summary
- Collapses `release-java-provider.yml`, `release-java-sdk-to-maven-central.yml`, and `update-docs.yml` into a single `release.yml` with three sequential jobs (`release-provider` → `publish-maven-central` → `trigger-docs-update`), so the docs trigger only fires after Maven Central publishing succeeds (fixes the v1.27.0 scenario in #2192).
- Updates the `pulumi/docs` workflow filename to the renamed `pulumi-sdk-java-docs.yml` (the prior `pulumi-java-sdk.yml` was deleted, which caused the v1.28.0 docs trigger to fail with HTTP 422 in [run 26110531106](https://github.com/pulumi/pulumi-java/actions/runs/26110531106)).
- Drops the `pulumictl` + `scripts/versions.sh` hop in the Maven Central job — the version is now passed directly from the `release-provider` job's output as `PULUMI_JAVA_SDK_VERSION` (`v` prefix stripped).
- Adds an `on:` comment documenting the release interface (`changie batch <version>` → Release PR → CHANGELOG.md merge fires the workflow).

Fixes #2195

## Test plan
- [x] `actionlint .github/workflows/release.yml` — only pre-existing info-level warnings remain
- [ ] After merge, manually invoke via `gh workflow run release.yml` to confirm the `Check if release exists` gate short-circuits goreleaser cleanly
- [ ] On the next real release: confirm jobs run sequentially and the `pulumi/docs` PR is opened only after the artifact appears on Maven Central

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+          version: latest
+          args: -p 3 -f .goreleaser.yml --clean --skip=validate --release-notes=${{ env.RELEASE_NOTES_PATH }}
+      - name: Push additional tags
+        if: steps.check-release.outputs.exists == 'false'
+        run: |
+          git tag sdk/${{ env.GORELEASER_CURRENT_TAG }}
+          git push origin sdk/${{ env.GORELEASER_CURRENT_TAG }}
+
+  publish-maven-central:
+    needs: release-provider
```
iwahbe, line 72: Otherwise `publish-maven-central` will skip if `release-provider` skips, which makes retrying impossible.
```suggestion
    needs: release-provider
    if: ${{ !failure() && !cancelled() }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+          java-version: '11'
+          distribution: 'temurin'
+          cache: 'gradle'
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@ac638b010cf58a27ee6c972d7336334ccaf61c96 # v4.4.1
+        with:
+          gradle-version: 8.14.1
+      - name: Publish Pulumi Java SDK
+        run: |
+          cd sdk/java && gradle pulumi:publishToSonatype closeAndReleaseSonatypeStagingRepository
```
iwahbe, line 130: Can we make this idempotent?

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+
+  trigger-docs-update:
+    needs: [release-provider, publish-maven-central]
```
iwahbe, line 136:
```suggestion
    needs: publish-maven-central
    if: ${{ !failure() && !cancelled() }}
```

iwahbe (approved):
Looks a lot better.

I'm worried about making partial retries possible, so it would be really nice if the steps were idempotent.

That said, it's no worse then the old jobs AFAIK.

PR #2209, opened 2026-05-30
Feat mill executor

ken1ma:
## Summary

Adds Mill 1.x (https://mill-build.org/) as a build-tool executor for `pulumi-language-java`,
alongside the existing maven, gradle, sbt, jbang, and jar executors.
A separate first commit also fixes a pre-existing bug in sbt project detection.

## Motivation

`executor_sbt.go` invokes sbt with `-batch`, which forces a full JVM cold-start
on every `pulumi preview`.
sbt's client-server protocol does not forward per-invocation client environment variables to forked tasks,
so even with a long-running sbt server in another terminal,
`pulumi preview` cannot use it — `PULUMI_MONITOR` and other env vars set by `pulumi-language-java`
do not reach the forked Pulumi program JVM.

Mill 1.x's `RunModule` defines:

```scala
def propagateEnv: T[Boolean] = Task { true }
```

In `runForkedTask`, this causes the client's live environment to be merged into the forked JVM's env before any user-defined
`forkEnv`:

```scala
env = (if (propEnv) ctx.env else Map()) ++ env,
```

Source: [`libs/javalib/src/mill/javalib/RunModule.scala`](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/RunModule.scala)

This makes Mill's daemon — enabled by default in 1.x — genuinely usable for Pulumi programs
without any build-side configuration.
Measured on a real Pulumi-Scala project (84 resources, Scala 3),
`pulumi preview --diff` wall clock drops from ~19.6s (sbt warm) to ~13.4s (Mill warm daemon).

## Commit 1 — Fix sbt project detection always matching

`executor_sbt.go`'s marker-file check used `if err == nil` against `fsys.FileExists`:

```go
for _, p := range sbtMarkers {
    _, err := fsys.FileExists(opts.WD, p)
    if err == nil {
        return true, nil
    }
}
```

`FileExists` returns `(false, nil)` when a file does not exist (it filters `IsNotExist` errors internally),
so `err == nil` matched whether or not the marker files were present.
Result: sbt detection returned true for any directory the OS could stat.
The fix follows the pattern already used by `executor_jbang.go` — inspect the returned `exists` bool,
not the absence of an error.

Split into its own commit so reviewers can mentally separate the bug fix from the feature;
happy to land as a standalone PR if maintainers prefer.

### `pkg/internal/executors/executor_mill.go`

Detects projects by the presence of `build.mill`, `build.mill.scala`, or `build.mill.yaml`.
Probes `./mill` then `mill` on `$PATH` (matches existing executors).

For programmable builds (`build.mill`, `build.mill.scala`), the module that holds the Pulumi
program is named explicitly via a required `mill-module` runtime option in `Pulumi.yaml`:

```yaml
runtime:
  name: java
  options:
    mill-module: app   # the `object app extends ScalaModule` in build.mill
```

Mill addresses tasks by module path and, unlike sbt's root project, has no implicit default
module for a programmable build. Inferring the name by scanning the build file is unreliable —
a regex cannot tell a real `object X extends ScalaModule` declaration from one inside a comment
or a string literal, nor know which of several modules is the Pulumi program — so when the
option is missing the executor fails with a clear error rather than guessing. The option is a
typed `MillModule` field on `JavaExecutorOptions`, parsed in `main.go` alongside `binary` and
`use-executor`; other executors ignore it.

For declarative `build.mill.yaml` builds, the file itself describes a single root module with no explicit name;
tasks are invoked without a module prefix and no option is needed.

Invocations:

- `BuildArgs:  <module>.compile`  (or `compile` for YAML root)
- `RunArgs:    <module>.run`      (or `run`)
- `PluginArgs: --ticker false <module>.runMain com.pulumi.bootstrap.internal.Main packages`

`--ticker false` is the canonical Mill 1.x flag for silencing the progress ticker
per the [CLI flags reference](https://mill-build.org/mill/cli/flags.html),
keeping STDOUT clean for the bootstrap JSON output.
`--disable-ticker` is documented as deprecated in favor of this form.

### Scope: Mill 1.x only

| Test | Covers |
|---|---|
| `TestMillExecutor` | Happy path; asserts `Cmd`, `RunArgs`, `BuildArgs`, `PluginArgs` |
| `TestMillBuildMillScalaVariant` | `build.mill.scala` IDE-friendly variant |
| `TestMillBuildMillYaml` | `build.mill.yaml` declarative form; bare task names |
| `TestMillBuildScIgnored` | Explicit rejection of Mill 0.x `build.sc` |
| `TestMillModuleRequired` | Error when a programmable build has no `mill-module` option |
| `TestMillUseExecutor` | `runtime.options.use-executor` override |
| `TestMillMillw` | `./mill` wrapper script takes precedence over `$PATH` |

### Integration test (`tests/examples/minimalmill/`)

Mirrors `tests/examples/minimalsbt/`: same `App.scala`, same `Pulumi.yaml` shape (plus a
`runtime.options.mill-module: minimalmill` entry naming the program's module),
equivalent build setup using a minimal Mill 1.x `build.mill`, with Mill's default source layout (flat).
A `minimalmill` subtest is added to `tests/examples/java_examples_test.go` immediately after `minimalsbt`,
with identical `Config` / `Secrets` and the same `ExtraRuntimeValidation`.
`.github/workflows/ci.yml`'s examples matrix gains `minimalmill`
between `minimal` and `minimalsbt` (alphabetical order).

The fixture ships Mill's official launcher script (mirroring how `minimalsbt` ships an `sbt` wrapper).
The launcher reads `//| mill-version: 1.1.6` from `build.mill`
and downloads the matching Mill JAR on first run, so CI does not need a separate Mill toolchain install.

## Limitations / future work

- **The module must be named explicitly** via `mill-module` for programmable builds.
- **Mill 0.x is intentionally unsupported.** Could be added on demand; would warrant a separate executor.
- **`package.mill.yaml`** in subdirectories (for multi-module YAML projects) is not detected.
  The current implementation only inspects workspace-root build files.

## Compatibility / breaking changes

None. Mill detection runs last in the executor chain;
existing sbt/maven/gradle/jbang/jar projects are unaffected.

## Testing

- `mise exec -- make lint` — 0 issues
- `cd pkg && go test -count=1 ./...` — all four packages pass (`pulumi-language-java`,
  `codegen/java`, `codegen/java/names`, `internal/executors`)
- `make test_example.minimalmill` — end-to-end pass in ~12s
- Verified end-to-end against an unrelated real Pulumi-Scala project — wall clock drops
  from ~19.6s to ~13.4s, confirming Mill 1.x's `propagateEnv` semantics deliver the expected warm-daemon benefit


pkg/internal/executors/executor_mill.go
```diff
@@ -0,0 +1,171 @@
...
+	"io/fs"
+	"regexp"
+	"strings"
+
+	"github.com/pulumi/pulumi-java/pkg/internal/fsys"
+)
+
+type mill struct{}
+
+var _ javaExecutorFactory = &mill{}
```
iwahbe:
```suggestion
var _ javaExecutorFactory = mill{}
```

pkg/internal/executors/executor_mill.go
```diff
@@ -0,0 +1,171 @@
...
+//
+// For declarative build.mill.yaml projects, the file itself describes a
+// single root module with no explicit name, and Mill tasks are invoked
+// without a module prefix (e.g. `mill compile` rather than `mill foo.compile`).
+// findModuleName returns the empty string in that case, which taskPath
+// translates into bare task selectors.
+//
+// Projects with multiple programmable modules where the Pulumi program is not
+// the first can override by setting runtime.options.use-executor to a path
+// that contains "mill" (TODO: support a dedicated option for module name).
```
iwahbe: Let's not include this. Instead, let's open an issue to track the additional feature request.

pkg/internal/executors/executor_mill.go
```diff
@@ -0,0 +1,171 @@
...
+func (mill) findModuleName(opts JavaExecutorOptions) (string, error) {
```
iwahbe: As I understand this, we are running a regex over the build file to try to find the correct module. That seems extremely error prone. The obvious example is if someone has the value in a string or commented out with a multi-line comment.

I'd feel better if we made this an explicit required user option.

iwahbe (commented):
Hi @ken1ma, thanks for the PR.

In general, the PR looks reasonable. The only place I have a real issue is `findModuleName`.

pkg/internal/executors/executor_mill.go
ken1ma: Thanks for the review.

I'm not familiar with Go, but I think I follow the issue. I've made the change in `executor_mill.go`.

This now diverges from the other executors (`executor_sbt.go`, `executor_gradle.go`, …), which still use the pointer form, e.g. `var _ javaExecutorFactory = &sbt{}`. I kept the change scoped to this PR's file rather than touching unrelated executors - happy to update the others for consistency if you'd prefer, or leave them as-is.

pkg/internal/executors/executor_mill.go
ken1ma: Removed.

pkg/internal/executors/executor_mill.go
ken1ma: @iwahbe 
Agreed.  I'll drop the regex and require the module explicitly via a mill-module runtime option (programmable `build.mill` requires it; declarative `build.mill.yaml` stays unprefixed).

Quick question on plumbing it through `JavaExecutorOptions`: prefer a generic `Options map[string]any` that each executor reads its own keys from, or a typed `MillModule` string field parsed in `main.go`? I lean generic to keep a Mill-only field off the shared struct, but happy either way.

The user-facing config is the same regardless of which we pick:

    runtime:
      name: java
      options:
        mill-module: app   # the `object app extends ScalaModule` in build.mill

iwahbe: I'd add the typed `MillModule` field. Other executors can ignore it.

ken1ma: @iwahbe 
Implemented the typed `MillModule`, and updated the PR description accordingly.
Could you take a look?

iwahbe (approved):
Thanks for working on this PR.

PR #2216, opened 2026-06-03
Improve error logs when serialization fails

julienp:
Ref https://github.com/pulumi/pulumi/issues/21901


iwahbe (approved):
I think this is fine

PR #2242, opened 2026-07-07
Add extension parameterization codegen

guineveresaenger:
This pull request adds codegen capability for provider parameterization extensions.

Part of https://github.com/pulumi/pulumi/issues/17059.


iwahbe (approved):
The implementation looks good, but I want us to rely on the language test from pu/pu, not on "extension-parameterized". 

We shouldn't merge until we get that test passing, and given that, I don't think "extension-parameterized" adds much.

.changes/unreleased/Dependencies-2242.yaml
iwahbe, line 1: This shouldn't be a changelog entry.

iwahbe (commented):
Other then dropping the spurious changelog entry, you are good to go.

PR #2246, opened 2026-07-10
Support cancellation of in-flight stack operations in the Automation API

addu390:


addu390:
@Frassle @iwahbe Whenever you get a chance, I'd really appreciate a review on this one. No hurry, and thanks for your time!

sdk/java/pulumi/src/main/java/com/pulumi/automation/CancellationToken.java
```diff
@@ -0,0 +1,95 @@
...
+ * {@link CancellationTokenSource}.
+ */
+public final class CancellationToken {
+    private static final CancellationToken NONE = new CancellationToken(false);
+
+    private final Object lock = new Object();
+    private final Map<CancellationTokenRegistration, Runnable> registrations = new LinkedHashMap<>();
+    private boolean cancellationRequested;
+
+    CancellationToken(boolean canceled) {
```
Frassle: This is only ever false? Probably just drop the parameter

sdk/java/pulumi/src/main/java/com/pulumi/automation/CancellationToken.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * {@link CancellationToken} propagates notification that an operation should
+ * be canceled. Tokens are created and canceled through a
+ * {@link CancellationTokenSource}.
+ */
+public final class CancellationToken {
+    private static final CancellationToken NONE = new CancellationToken(false);
```
Frassle: This will accumulate registrations forever, that might be a memory leak that matters.

addu390: Moved the cancellation state into `CancellationTokenSource` so the shared `none()` token no longer stores registrations.

addu390:
Hi @Frassle 
Not a hurry, addressed the comments, would appreciate a review when you get a chance.

PR #2267, opened 2026-08-07
Generate typed interfaces for discriminated unions

### LINKED ISSUE #2266: Discriminated unions collapse to Object; const tags are not pre-filled; output deserialization casts by shape
Mirrors pulumi/pulumi-dotnet#1098 for Java:

1. Union-typed slots render as `Object`; any value compiles.
2. Const-tagged discriminator properties are user-supplied builder arguments (`.__type("PermissionExpressionAnd")`) rather than pre-filled from the schema const.
3. `Converter.tryConvertOneOf` uses `canBeCast`, which fails on Map values, instead of dispatching on the discriminator tag.

Anchors: `l2-discriminated-union`, `l2-discriminated-union-many`, `l2-discriminated-union-internal` (pulumi/pulumi#24213). #2021 tracks implementing the conformance harness; this issue tracks the codegen and runtime changes needed to pass it with typed output.

Producer context: pulumi/pulumi-pulumiservice#1025.

lukebward:
## Summary

- Generate opt-in typed Java interfaces for fully mapped discriminated unions.
- Set discriminator tags automatically and deserialize union values by tag.
- Preserve existing output unless `language.java.fullyTypedUnions` is enabled. Fixes #2266.

## Example

Variants are statically checked, and callers no longer set the discriminator:

```java
var resource = new Example("example", ExampleArgs.builder()
    .unionOf(Variant1Args.builder()
        .payload("value")
        .build())
    .build());

resource.unionOf().applyValue(union ->
    union instanceof Variant1 variant ? variant.payload() : null);
```

## Test plan

- [x] Codegen tests covering typed and fallback generation
- [x] Java serializer and converter tests
- [x] Java SDK and integration checks


iwahbe:
How common are discriminated unions with < 3 members? This implementation makes adding a 3rd member to such a union breaking, which I think is surprising. If the registry doesn't have any discriminated unions now with 1 or 2 members, I think we should consider a break here. 

If the feature is used, then we might have to deal with the incongruity between unions of different sizes.

Frassle:
> If the feature is used, then we might have to deal with the incongruity between unions of different sizes.

It's a break change anyway going from object -> typed thing probably. I wonder if we should add a flag to let providers opt-in to this break change. `fullyTypedUnions` option for each language in the Language section of the schema as we add support for each language. And then make it apply to 1-2 element unions as well.

lukebward:
At least for a provider like azure-native, there's a few hundred < 3 member discriminated unions, and also a few hundred that are 3+.  So, both are large.  

> fullyTypedUnions option for each language in the Language section of the schema as we add support for each language. And then make it apply to 1-2 element unions as well.

Would the idea be with this flag enabled, we break at 1-2 elements?  And just use Interfaces from the start?

iwahbe:
> At least for a provider like azure-native, there's a few hundred < 3 member discriminated unions, and also a few hundred that are 3+. So, both are large.

Unfortunate.

> > fullyTypedUnions option for each language in the Language section of the schema as we add support for each language. And then make it apply to 1-2 element unions as well.
> 
> Would the idea be with this flag enabled, we break at 1-2 elements? And just use Interfaces from the start?

The idea is with this flag enabled, we do the interface for all unions. Without it enabled, nothing changes. Going from untyped to typed is also a breaking change.

pgavlin:
@lukebward has all the feedback for this PR been addressed? @iwahbe / @Frassle do we want to apply this approach to https://github.com/pulumi/pulumi-dotnet/pull/1108 and https://github.com/pulumi/pulumi/pull/24238 as well?

lukebward:
> @lukebward has all the feedback for this PR been addressed? @iwahbe / @Frassle do we want to apply this approach to [pulumi/pulumi-dotnet#1108](https://github.com/pulumi/pulumi-dotnet/pull/1108) and [pulumi/pulumi#24238](https://github.com/pulumi/pulumi/pull/24238) as well?

All previous feedback has been applied

Frassle:
> @iwahbe / @Frassle do we want to apply this approach to https://github.com/pulumi/pulumi-dotnet/pull/1108 and https://github.com/pulumi/pulumi/pull/24238 as well?

Yeh, we need the opt-in flag for all languages so we don't break providers.

lukebward:
> > @iwahbe / @Frassle do we want to apply this approach to [pulumi/pulumi-dotnet#1108](https://github.com/pulumi/pulumi-dotnet/pull/1108) and [pulumi/pulumi#24238](https://github.com/pulumi/pulumi/pull/24238) as well?
> 
> Yeh, we need the opt-in flag for all languages so we don't break providers.

Flags added to https://github.com/pulumi/pulumi-dotnet/pull/1108 and https://github.com/pulumi/pulumi/pull/24238.

.changes/unreleased/Improvements-discriminated-union-runtime.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Improvements
+body: Deserialize discriminated union values by dispatching on the schema discriminator property via the new DiscriminatedUnion annotation, instead of matching on the value shape
```
iwahbe: This is an implementation detail of how we handle discriminated unions. It doesn't need a changelog entry.

.changes/unreleased/Improvements-discriminated-union-codegen.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Improvements
+body: 'Add the `fullyTypedUnions` Java package option, which generates a typed interface for each discriminated union that its members implement, instead of collapsing the union to Object or to Either. Member count does not affect the decision, so adding a member to a union does not change the generated shape. Const-valued properties are now filled in by generated builders regardless of the option, so a discriminator tag never has to be written by hand'
```
iwahbe:
```suggestion
body: 'Add the `fullyTypedUnions` Java package option, which generates a typed interface for each discriminated union that its members implement, instead of collapsing the union to Object or to Either
```

pkg/codegen/java/gen_defaults.go
iwahbe: This should be a separate PR (moved out to https://github.com/pulumi/pulumi-java/pull/2298). If you need this functionality, you can stack your PR.

pkg/codegen/java/discriminated_union.go
```diff
@@ -0,0 +1,550 @@
...
+
+	"github.com/pulumi/pulumi-java/pkg/codegen/java/names"
+)
+
+const typeRefPrefix = "#/types/"
+
+// discriminatedUnion is a schema union that carries a discriminator with a non-empty mapping. Such
+// a union is generated as a Java interface that each member class implements, rather than degrading
+// to java.lang.Object or to Either<L, R>. Member count never takes part in the decision, so adding
+// a member to a union does not change the shape of the code generated for it.
```
iwahbe:
```suggestion
// a union is generated as a Java interface that each member class implements.
```

PR #2282, opened 2026-08-18
[sdk] Parse gRPC messages with a raised protobuf recursion limit

### LINKED ISSUE #2277: AgentcoreGatewayTarget with an inline tool schema breaks pulumi preview for the entire program (protobuf recursion limit)
### What happened?

Declaring an `aws.bedrock.AgentcoreGatewayTarget` with an **inline** tool schema makes `pulumi preview` abort the entire program with a protobuf recursion error. The failure is not scoped to that resource — no other resource in the stack can be previewed while it is declared, so the program becomes undeployable-by-inspection.

```
io.grpc.StatusRuntimeException: CANCELLED: Failed to read message.
Caused by: io.grpc.StatusRuntimeException: INTERNAL: Invalid protobuf byte sequence
Caused by: com.google.protobuf.InvalidProtocolBufferException:
    Protocol message had too many levels of nesting. May be malicious.
    Use setRecursionLimit() to increase the recursion depth limit.
      at InvalidProtocolBufferException.recursionLimitExceeded(InvalidProtocolBufferException.java:133)
```

The process exits 32. `Resources: 1 errored`, with no resource named.

Observed sequence, which may matter: the **first** `pulumi up` created the target successfully. Every subsequent `preview`/`up` fails, while the resource is in state. Removing the resource from the program restores preview immediately, including planning its deletion — so reading its existing state is fine; the failure appears when the desired state is also present.

### Likely cause

`AgentcoreGatewayTarget` embeds AWS's `SchemaDefinition`, which is genuinely recursive (JSON Schema: an object property contains another `SchemaDefinition`). The provider cannot express recursion, so it flattens it into a chain of distinct generated types:

```
AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaInlinePayloadInputSchemaPropertyItemsPropertyArgs
AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaInlinePayloadOutputSchemaPropertyItemsItemsArgs
```

In `com.pulumi:aws:7.41.0` this resource generates **93 types**, **42** of them for the inline tool schema alone. Counted from the sources jar:

```
unzip -l aws-7.41.0-sources.jar | grep -cE 'AgentcoreGatewayTarget[A-Za-z]*\.java'                       # 93
unzip -l aws-7.41.0-sources.jar | grep -cE 'AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchema.*\.java'  # 42
```

Protobuf caps deserialization at 100 nesting levels, adjustable through `CodedInputStream.setRecursionLimit()`. gRPC Java does not expose that setting (grpc/grpc-java#8256, grpc/grpc-java#9643), so a Java client cannot raise it.

If that is the mechanism, the same message would be accepted by SDKs whose gRPC bindings allow the limit to be raised, making this Java-specific rather than a provider schema problem. **I have not verified that** — I only have the Java reproduction, so it is a hypothesis for whoever picks this up rather than a finding.

### Steps to reproduce

Declare the resource with a minimal inline schema — the payload itself is a few hundred bytes, so this is not a size limit:

```kotlin
AgentcoreGatewayTarget(
    "echo-target",
    AgentcoreGatewayTargetArgs.builder()
        .name("echo")
        .gatewayIdentifier(gatewayId)
        .credentialProviderConfiguration(
            AgentcoreGatewayTargetCredentialProviderConfigurationArgs.builder()
                .gatewayIamRole(
                    AgentcoreGatewayTargetCredentialProviderConfigurationGatewayIamRoleArgs.builder().build(),
                ).build(),
        ).targetConfiguration(
            AgentcoreGatewayTargetTargetConfigurationArgs.builder()
                .mcp(
                    AgentcoreGatewayTargetTargetConfigurationMcpArgs.builder()
                        .lambda(
                            AgentcoreGatewayTargetTargetConfigurationMcpLambdaArgs.builder()
                                .lambdaArn(functionArn)
                                .toolSchema(
                                    AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaArgs.builder()
                                        .inlinePayloads(
                                            AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaInlinePayloadArgs
                                                .builder()
                                                .name("echo")
                                                .description("Returns the message it was given")
                                                .inputSchema(
                                                    AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaInlinePayloadInputSchemaArgs
                                                        .builder()
                                                        .type("object")
                                                        .build(),
                                                ).build(),
                                        ).build(),
                                ).build(),
                        ).build(),
                ).build(),
        ).build(),
    CustomResourceOptions.builder().parent(this).build(),
)
```

Then run `pulumi preview`. (Extracted from a working component rather than run standalone, so identifiers such as `gatewayId` and `functionArn` are placeholders for `Output<String>` values.)

### Expected behavior

`pulumi preview` completes, showing a plan for this resource alongside the rest of the stack.

### Actual behavior

The program aborts before producing a plan for any resource.

### Workaround

Use the S3 form of the tool schema instead, which AWS documents as an equal alternative (1MB inline vs 10MB in S3):

```kotlin
.toolSchema(
    AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaArgs.builder()
        .s3(
            AgentcoreGatewayTargetTargetConfigurationMcpLambdaToolSchemaS3Args.builder()
                .uri(Output.format("s3://%s/%s", bucket.bucket(), obj.key()))
                .bucketOwnerAccountId(accountId)
                .build(),
        ).build(),
)
```

Two fields rather than the 42-type chain. Preview then works, and the existing target updates in place rather than being replaced.

### Why this is worth fixing rather than documenting

The blast radius is the whole program, not the resource. A user who declares one of these loses `pulumi preview` for every resource in the stack — on a large stack that removes the primary safety check before deployment, while `pulumi up` continues to work.

### Output of `pulumi about`

```
CLI
Version      3.256.0
Go Version   go1.26.5
Go Compiler  gc

Plugins
KIND      NAME  VERSION
language  java  1.34.0

Host
OS       darwin
Version  26.6
Arch     arm64

This project is written in java:
  version='openjdk 21.0.2 2024-01-16' javac='21.0.2' gradle='9.7.0'

Dependencies
  com.pulumi:pulumi  1.36.0
  com.pulumi:aws     7.41.0
```

(Backend section omitted — self-managed S3.)

morhaf-ko:
Fixes #2277.

### What happens today

Resource inputs and outputs travel as protobuf `Struct`s, where one level of a user-visible object costs three levels of protobuf nesting: `Struct` -> map entry -> `Value`. Protobuf's Java runtime refuses to parse past 100 levels, so the Java SDK tops out at roughly 33 levels of property nesting.

gRPC's default marshaller keeps that default and, until recently, offered no way to change it (grpc/grpc-java#8256). When a provider returns something deeper - the recursive JSON Schema shapes AWS flattens into long generated type chains, for example - the parse fails on the channel rather than on the resource:

```
io.grpc.StatusRuntimeException: CANCELLED: Failed to read message.
Caused by: io.grpc.StatusRuntimeException: INTERNAL: Invalid protobuf byte sequence
Caused by: com.google.protobuf.InvalidProtocolBufferException:
    Protocol message had too many levels of nesting. May be malicious.
```

Because the failure is on the channel, the whole program aborts: no resource in the stack gets a plan, and no resource is named in the error.

### The fix

grpc-java has exposed `ProtoUtils.marshallerWithRecursionLimit` since 1.56, and this repo is on 1.57.2, so no dependency bump is needed. `GrpcRecursionLimit` rebuilds the generated marshallers with the limit raised to 10,000 - matching what the Go engine parses with, so the Java SDK is no longer the narrower end of the connection - and installs them:

- `GrpcMonitor` and `GrpcEngine` add a channel interceptor that swaps in the raised **response** marshaller. This is the path in the issue.
- `ResourceProviderService` rebinds its service definition so a Java component provider parses deeply nested **requests** too. A `ServerInterceptor` cannot do this: the method descriptor has already deserialized the request by the time one is consulted.

Serialization is untouched - protobuf imposes no depth limit when writing.

### Tests

`GrpcRecursionLimitTest` covers both directions with 64 levels of `Struct` nesting, which is comfortably past the default:

- `monitorParsesDeeplyNestedResponses` runs a real `ResourceMonitor` server over loopback and calls it through `GrpcMonitor`, so it exercises the wiring and a real serialize/parse round trip. (An in-process channel would not reproduce the bug - gRPC hands the message across without serializing it.)
- `reboundServiceParsesDeeplyNestedRequests` and `reboundServiceKeepsItsMethodsAndHandlers` cover the server-side rebind.

I checked that `monitorParsesDeeplyNestedResponses` fails with the original `INTERNAL: Invalid protobuf byte sequence` when the interceptor is removed from `GrpcMonitor`.

### Notes

- 10,000 is a policy choice. Deep enough that the engine's own limit binds first; bounded, because protobuf parses recursively and an unbounded limit trades a clear error for a `StackOverflowError`. Happy to move it if you would rather it were lower.
- This does not add the conformance test @Frassle mentioned in #2277 - that lives in `pulumi/pulumi` - so it is up for grabs alongside this.

### DIFF
--- a/.changes/unreleased/bug-fixes-2282.yaml
+++ b/.changes/unreleased/bug-fixes-2282.yaml
@@ -1,0 +1,6 @@
+component: sdk
+kind: bug-fixes
+body: Parse gRPC messages with a raised protobuf recursion limit, so resources with deeply nested properties no longer abort the whole program with "Protocol message had too many levels of nesting"
+time: 2026-08-18T12:00:00.000000000Z
+custom:
+    PR: "2282"
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/GrpcEngine.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/GrpcEngine.java
@@ -2,6 +2,7 @@ package com.pulumi.deployment.internal;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.pulumi.core.internal.ContextAwareCompletableFuture;
+import com.pulumi.internal.GrpcRecursionLimit;
 import io.grpc.ManagedChannelBuilder;
 import pulumirpc.EngineGrpc;
 import pulumirpc.EngineOuterClass.GetRootResourceRequest;
@@ -25,7 +26,8 @@ public class GrpcEngine implements Engine {
         var channelBuilder = ManagedChannelBuilder
                 .forTarget(engine)
                 .usePlaintext() // disable TLS
-                .maxInboundMessageSize(maxRpcMessageSizeInBytes);
+                .maxInboundMessageSize(maxRpcMessageSizeInBytes)
+                .intercept(GrpcRecursionLimit.clientInterceptor());
         var interceptor = Instrumentation.getClientInterceptor();
         if (interceptor != null) {
             channelBuilder.intercept(interceptor);
--- a/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/GrpcMonitor.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/deployment/internal/GrpcMonitor.java
@@ -2,6 +2,7 @@ package com.pulumi.deployment.internal;
 
 import com.google.common.util.concurrent.ListenableFuture;
 import com.pulumi.core.internal.ContextAwareCompletableFuture;
+import com.pulumi.internal.GrpcRecursionLimit;
 import com.pulumi.resources.Resource;
 import io.grpc.ManagedChannelBuilder;
 import pulumirpc.Provider.CallResponse;
@@ -33,7 +34,8 @@ public class GrpcMonitor implements Monitor {
         var channelBuilder = ManagedChannelBuilder
                 .forTarget(monitor)
                 .usePlaintext() // disable TLS
-                .maxInboundMessageSize(maxRpcMessageSizeInBytes);
+                .maxInboundMessageSize(maxRpcMessageSizeInBytes)
+                .intercept(GrpcRecursionLimit.clientInterceptor());
         var interceptor = Instrumentation.getClientInterceptor();
         if (interceptor != null) {
             channelBuilder.intercept(interceptor);
--- a/sdk/java/pulumi/src/main/java/com/pulumi/internal/GrpcRecursionLimit.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/internal/GrpcRecursionLimit.java
@@ -1,0 +1,113 @@
+package com.pulumi.internal;
+
+import com.google.protobuf.Message;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.MethodDescriptor;
+import io.grpc.MethodDescriptor.Marshaller;
+import io.grpc.MethodDescriptor.PrototypeMarshaller;
+import io.grpc.ServerMethodDefinition;
+import io.grpc.ServerServiceDefinition;
+import io.grpc.ServiceDescriptor;
+import io.grpc.protobuf.ProtoUtils;
+
+import java.util.ArrayList;
+
+/**
+ * Raises the nesting limit protobuf enforces when Pulumi's gRPC messages are parsed.
+ *
+ * <p>Resource inputs and outputs travel as protobuf {@code Struct}s, and each level of a
+ * user-visible object costs three levels of protobuf nesting: {@code Struct}, map entry,
+ * {@code Value}. Protobuf's Java runtime refuses to parse past 100 levels, and gRPC's default
+ * marshaller keeps that default, so a resource whose schema nests deeply - AWS' recursive
+ * JSON Schema shapes, for instance - fails to deserialize. The failure lands on the channel
+ * rather than the resource, which aborts the whole program instead of one registration.
+ *
+ * <p>The marshallers installed here parse at {@link #RECURSION_LIMIT} instead. Serialization is
+ * left alone: protobuf imposes no depth limit when writing.
+ */
+public final class GrpcRecursionLimit {
+
+    /**
+     * Matches the limit the Go engine parses with, so that the Java SDK is not the narrower end
+     * of the connection.
+     */
+    public static final int RECURSION_LIMIT = 10_000;
+
+    private static final ClientInterceptor CLIENT_INTERCEPTOR = new ClientInterceptor() {
+        @Override
+        public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+                MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+            return next.newCall(
+                    method.toBuilder()
+                            .setResponseMarshaller(raise(method.getResponseMarshaller()))
+                            .build(),
+                    callOptions);
+        }
+    };
+
+    private GrpcRecursionLimit() {}
+
+    /**
+     * Returns a channel interceptor that parses responses at {@link #RECURSION_LIMIT}.
+     *
+     * @return the interceptor, safe to share between channels
+     */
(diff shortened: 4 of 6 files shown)


morhaf-ko:
The three red checks here are the fork-OIDC limitation, not the patch.

`prerequisites`, `Lint Go`, and `ci-ok` each fail after 3-10s on their first step, `Fetch secrets from ESC`, with `Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`. `ci.yml` asks for `permissions: write-all`, but GitHub withholds the OIDC id-token from `pull_request` runs triggered by a fork, so `pulumi/esc-action` cannot mint a staging token. `java-sdk-tests` is the only job in the workflow that does not start with that step, which is why it is the only one that ran - and it passed. `go-tests`, `test_integrations`, `test_templates`, and `examples` show as skipped because they gate on `prerequisites`.

To cover what CI could not run, I ran the suite locally against this branch (mise toolchain, JDK 11, Gradle 8.14.1):

| Suite | Result |
| --- | --- |
| `make lint` | pass, 0 issues |
| `make build` | pass |
| `cd sdk/java && gradle build testAll` | pass - 542 tests, 0 failures (`test` 469, `autoTest` 72, `isolatedTest` 1) |
| `cd pkg && go test -timeout 40m ./...` | pass - `pulumi-language-java` 528s, `pkg/codegen/java` 126s |

`make test_integrations` does not run on my machine, but it fails identically on an unmodified `main` checkout - the same seven subtests (`stale-parameterized-packageref`, `package-add`, `about`, `provider-maven`, `stack-transformation`, `provider-gradle`, `stack-reference`), so it is a pre-existing environment gap rather than anything this branch introduced. The causes are local: the test projects pin Gradle 7.4, which cannot run under the JDK 21 that is my system default (`Unsupported class file major version 65` while compiling `settings.gradle`, before any Java runs), and the `about` test needs the versioning env vars and `maven` that CI provisions from `versions.txt`. `examples` and `test_templates` need real cloud credentials and the staging backend, so they are out of reach entirely from outside the org.

For a trusted run, I think hosting it from a `pulumi`-owned branch - the way #2220 did for #2209 - is the only option that actually covers this branch. `/run-acceptance-tests` would dispatch, but `ci.yml` reads `env.PR_COMMIT_SHA` for the `go-lint` checkout and that variable is not defined anywhere in the repo, and no job reads `github.event.client_payload`, so a `repository_dispatch` run checks out the default branch rather than the PR head. Happy to be told I have misread that.

One reviewable judgement call worth your eyes: I set the limit to 10,000 to match what protobuf-go parses with, so the Java SDK stops being the narrower end of the connection. It is deliberately bounded rather than `Integer.MAX_VALUE`, because protobuf parses recursively and an unbounded limit trades a clear error for a `StackOverflowError`. Easy to change if you would rather it were lower.

morhaf-ko:
Thanks - I checked this branch against the 300 levels `l2-large-map` uses, and pushed the test depth up from 64 to 300 to match it, so the SDK unit test now pins the same shape the conformance test will exercise.

Where Java sits, measured on this branch:

| Nesting depth | Result |
| --- | --- |
| 300 (the conformance test) | pass |
| 350 | pass |
| 1000 | `StackOverflowError` |

So the raised limit is not what binds any more - protobuf parses recursively, and the thread's stack gives out somewhere between 350 and 1000 levels, well before the 10,000 limit is reached. That answers the "why 10,000" question better than my earlier note did: the constant exists to stop protobuf's artificial cap of 100 from being the limit, and past that the runtime's own limit applies. I have said as much in a comment on `RECURSION_LIMIT` rather than tuning the number, since lowering it would just reintroduce an arbitrary ceiling under the real one.

Your 350 observation matches ours, for whatever that is worth as a cross-check.

`gradle build testAll` stays green at the new depth: 542 tests, 0 failures.

Frassle (approved):
I think this looks ok

morhaf-ko:
> I think this looks ok

Thanks @Frassle could you please let me know how we can enable the checks to work?

PR #2298, opened 2026-09-01
Fill in constant property values from the generated args type

iwahbe:
## Summary

Split out of #2267, which needs this to set discriminator tags automatically.

This PR makes 3 changes:

1. Literal properties have that property correctly set.
2. Literal builder args are now deprecated, since they were previously not consistently applied. 
3. Programgen now omits generating literal properties, since that isn't necessary (1) and would warn (2).

### DIFF
--- a/pkg/codegen/java/gen.go
+++ b/pkg/codegen/java/gen.go
@@ -411,7 +411,13 @@ type propJavadocOptions struct {
 func genPropJavadoc(ctx *classFileContext, prop *schema.Property, options propJavadocOptions) {
 	w := ctx.writer
 
-	if prop.Comment == "" && prop.DeprecationMessage == "" {
+	deprecationMessage := prop.DeprecationMessage
+	if options.isBuilder && prop.ConstValue != nil && deprecationMessage == "" {
+		// The generated type supplies the constant itself, so the setter is redundant.
+		deprecationMessage = "This property has a constant value, which is set automatically."
+	}
+
+	if prop.Comment == "" && deprecationMessage == "" {
 		return
 	}
 	fprintf(w, "%s/**\n", options.indent)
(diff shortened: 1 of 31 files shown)


pkg/codegen/java/gen.go
```diff
@@ -411,7 +411,13 @@ type propJavadocOptions struct {
 func genPropJavadoc(ctx *classFileContext, prop *schema.Property, options propJavadocOptions) {
 	w := ctx.writer
 
-	if prop.Comment == "" && prop.DeprecationMessage == "" {
+	deprecationMessage := prop.DeprecationMessage
+	if options.isBuilder && prop.ConstValue != nil && deprecationMessage == "" {
+		// The generated type supplies the constant itself, so the setter is redundant.
+		deprecationMessage = "This property has a constant value, which is set automatically."
+	}
```
Frassle: Should we not add this to the existing deprecation message if there is one?

pkg/codegen/testing/test/testdata/kubernetes20/java/src/main/java/com/pulumi/kubernetes/core/v1/ConfigMapList.java
Frassle: So now makeArgs sets these but when we call build we also set them? Can't we clean up makeArgs to not bother setting the constants?

pkg/codegen/java/gen.go
iwahbe: Yes we should.

pkg/codegen/testing/test/testdata/kubernetes20/java/src/main/java/com/pulumi/kubernetes/core/v1/ConfigMapList.java
iwahbe: We can. It changes behavior slightly, in that it lets users override const values universally, instead of only in nested objects.

I've made the change, since the resulting SDK is more consistent & understandable. It would be better if we just didn't let users set constant values at all, but c'est la vie.

iwahbe:

- #2298 ◀
    - #2302
        - #2303

Frassle (approved):

PR #2302, opened 2026-09-02
Select union members by wire shape in the runtime

iwahbe:
This is the runtime half of strongly typed unions. It is the Java equivalent to https://github.com/pulumi/pulumi/pull/24484. This lets Java deserialize types into unions based on the shape of the type.

Instead of Python's reflection, we use Java's annotations:

- Add `@UnionType` and `@UnionType.Case` for interfaces that stand for a schema union, `@ConstValue` for properties a schema pins to a constant, and `UnionCase` for case classes that hold a non-object member by value.
- When the runtime deserializes a value into a `@UnionType` interface, it selects the one member whose wire shape admits the value. An object member admits a value under the closed reading: every key is a declared property, every required property is present, and a constant property holds its constant. The rules mirror `codegen.IsWireDiscriminatableUnionType` in pulumi/pulumi and `_wire_matches` in the Python SDK.
- A value that fits no member, or more than one, fails to deserialize with an error that names the members.
- Both the resource `Converter` and the provider `PropertyValueSerializer` apply the rule. Both serializers unwrap a case class to its value, so providers see no change on the wire.

### DIFF
--- a/sdk/java/pulumi/src/main/java/com/pulumi/core/TypeShape.java
+++ b/sdk/java/pulumi/src/main/java/com/pulumi/core/TypeShape.java
@@ -10,6 +10,7 @@ import com.pulumi.core.internal.annotations.InternalUse;
 import javax.annotation.CheckReturnValue;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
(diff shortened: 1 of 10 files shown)


iwahbe:

- #2298
    - #2302 ◀
        - #2303
            - #2304

PR #2303, opened 2026-09-02
Generate typed interfaces for unions under fullyTypedUnions

iwahbe:
This is the Java version of https://github.com/pulumi/pulumi/pull/24485, but heavily modified for an OO language. It changes codegen (under the `fullyTypedUnions` flag) so:

- Wire discriminatable unions are typed in Java.
- Each union is generated as a `interface`.
- Objects that are members of the union implement the interface.
- Primitive types that are members of the union can be lifted into the interface with a type-overloaded static function on the interface.
- Interfaces which are subsets of other interfaces extend them.

Language tests have had `fullyTypedUnions` flipped on. Since this only effects unions, I thought it was unnecessary to add a degree to our conformance test matrix for this.

### Dependencies

`codegen.IsWireDiscriminatableUnionType` landed in pulumi/pulumi after v3.260.0, so `go.mod` points at the master commit that adds it. Replace it with v3.261.0 once that ships.

`FullyTypedUnionsSdkVersion` pins the minimum `com.pulumi:pulumi` a generated SDK depends on to 1.37.0, the release that will carry #2302. Adjust if the release number differs.

### Conformance changes to note

- `l2-discriminated-union-many` moves to expected failures. Its program passes an output union interface where the input Args interface is expected. It passed before only because both sides were `java.lang.Object`, which is the same output-to-input gap Java has for object types.
- Input forms get an interface for every union whose members have distinct Java types. Output forms keep the discriminability gate; an output union that fails it collapses to the primitive its members share, or to `Object`, and is never `Either`. `l2-union` passes once #2304 lifts outputs into the interface.
- The pulumi/pulumi master commit also brings the new `l2-secret-unknown` test, which fails because wrapping an unknown output with `secret()` drops the secret marker, and refreshed SDK snapshots for the constant and output providers.
- A conformance-level legacy variant was tried and dropped: two variants publish the same provider SDK coordinates with different content, and the program build resolves whichever is in `~/.m2`. The `unions-legacy` codegen golden test locks in the legacy shape instead.

## Testing

- `unions` golden test covers scalar, enum, list, map, asset, archive, resource, and object members, a subset union, equal unions, a union nested in an object, and a function input and output.
- `unions-legacy` golden test generates the same schema without the option.
- `unions_test.go` covers location naming, name stability when a member is added, subset and equal relations, and case class names.
- `make lint`, `go test ./codegen/java/`, and the union conformance tests pass locally.


iwahbe:

- #2298
    - #2302
        - #2303 ◀
            - #2304

PR #2304, opened 2026-09-02
Lift outputs into union interfaces in generated programs

iwahbe:
## Summary

Stacked on #2303. Program generation now applies the union interface factory to an output whose element type is a wrapped member of the property's union, so `Output<String>` reaches a setter that expects `Output<ExampleTypedEnumPropertyArgs>`:

```java
.typedEnumProperty(enumOutputExample.type().applyValue(ExampleTypedEnumPropertyArgs::of))
```

The lift looks at the expression under the binder's `__convert` call, since every resource input is converted to an output type, and it leaves object values alone because those already implement the interface.

`l2-union` leaves the expected failures and passes.

## Testing

- `l2-union` and `l2-discriminated-union` pass locally with the option on.
- `make lint` and `go test ./codegen/java/` pass.


iwahbe:

- #2298
    - #2302
        - #2303
            - #2304 ◀

